PR #9, opened 2022-01-14
Use tektoncd/pipeline main branch

guillaumerose:
With the merge of https://github.com/tektoncd/pipeline/pull/4459, Task, ClusterTask, and Pipeline CRDs are optional. 
The controller will start fine even if the cluster doesn’t know about them.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -19,8 +19,8 @@ then
 fi
 if [[ ! -d ./pipeline ]]
 then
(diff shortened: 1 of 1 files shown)


vdemeester (approved):
Note: I think [v0.32.0](https://github.com/tektoncd/pipeline/releases/tag/v0.32.0) has this change too, and thus should "just" work too (checking out the tag)

guillaumerose:
I changed the code to use v0.32.0. Thanks! It's better than my fork.

PR #15, opened 2022-02-07
Controllers no longer had to add kcp.dev/cluster labels

guillaumerose:
By using the new namespace scheduler.

When a ns is created, it is assigned to a cluster. Objects are automatically synced.


bnallapeta (approved):
Works locally with these changes. Although, the issue with continuous syncing of 'Resource=leases' exist.

run.sh
```diff
@@ -135,12 +135,10 @@ kubectl get pods,taskruns,pipelineruns
 # Add a secret in the physical cluster so that the event listener and interceptors can query KCP API
 cp ./.kcp/admin.kubeconfig ./.kcp/remote.kubeconfig
 gsed -i "s/\[::1\]/host.docker.internal/" ./.kcp/remote.kubeconfig
```
bnallapeta, line 137: gsed does not work on fedora. Had to change it to sed.

vdemeester: Ahhh. OSX users 😝

guillaumerose: yes sorry. If someone knows how to make it transparent, I am happy to configure my machine. :D

vdemeester: @chmouel ^^

chmouel: i think if you prepend the path in your .zshrc with the gnubin paths from brew and it should works: 

at least it does for me, ie i have this :

```
path=(/opt/homebrew/opt/*/libexec/gnubin
      /opt/homebrew/opt/file-formula/bin
      /opt/homebrew/sbin
      /opt/homebrew/bin $path) #prepend
```

guillaumerose: Fixed, thanks :)

PR #16, opened 2022-02-09
Run on Fedora with podman

Michkov:
* Update documentation with required configuration
* Avoid using host.docker.internal when podman is used


guillaumerose (approved):

guillaumerose:
@bnallapeta I think you will be interested!

PR #20, opened 2022-02-10
Bump kcp when running in a container + cleanups

guillaumerose:


bnallapeta (approved):
LGTM. Works locally.
@guillaumerose  I think with the changes you have made to pipelines-deployment.yaml, we can address @vdemeester  comments easily (by letting users provide env variables or provide default values)

https://github.com/openshift-pipelines/pipelines-service/pull/14#discussion_r797542997

PR #21, opened 2022-02-14
Bump kcp

guillaumerose:
Certs. are now bound to localhost. 
All controllers are managed by a single flag.

### DIFF
--- a/ckcp/config/kcp-deployment.yaml
+++ b/ckcp/config/kcp-deployment.yaml
@@ -17,7 +17,7 @@ spec:
     spec:
       serviceAccountName: anyuid
       containers:
-        - image: quay.io/bnr/ckcp:dfc490d656822da51234c9c18678c3a0c7952c0d
+        - image: quay.io/bnr/ckcp:986710c754ed0dac9ae1525661de931e5dd7c0cc
           name: ckcp
           readinessProbe:
             initialDelaySeconds: 20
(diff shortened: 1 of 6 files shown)


bnallapeta (approved):

vdemeester (approved):

ckcp/config/kcp-deployment.yaml
```diff
@@ -17,7 +17,7 @@ spec:
     spec:
       serviceAccountName: anyuid
       containers:
-        - image: quay.io/bnr/ckcp:dfc490d656822da51234c9c18678c3a0c7952c0d
+        - image: quay.io/bnr/ckcp:986710c754ed0dac9ae1525661de931e5dd7c0cc
```
matzew, line 20: where do these images come from ?

bnallapeta: [build.sh](https://github.com/openshift-pipelines/pipelines-service/blob/main/ckcp/build.sh)
The above build.sh builds a docker image called ckcp and uploads it to quay.io/<username> (whatever KO_DOCKER_REPO points to)

Also, we will change this from quay.io/bnr/ckcp to $KO_DOCKER_REPO/ckcp soon.

guillaumerose:
I tried locally and with a remote cluster this morning and it is working. 
Maybe it is a little bit flaky?

PR #28, opened 2022-03-02
WIP Bump kcp

guillaumerose:
Use the latest image from their CI. 
Still trying to understand why only secrets are synced.


bnallapeta:
Closing this PR as we have now moved onto the kcp shared instance.

PR #42, opened 2022-04-07
Register the pod CRD through ArgoCD

Roming22:
This is done to mirror the KCP staging instance, which does not
sync the pod CRD by default.

### DIFF
--- a/ckcp/docker/ckcp/entry.sh
+++ b/ckcp/docker/ckcp/entry.sh
@@ -6,4 +6,4 @@ rm -rf .kcp/
   --pull-mode=false \
   --run-controllers \
   --auto-publish-apis \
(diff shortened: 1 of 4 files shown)


bnallapeta (commented):
[ckcp/docker/ckcp/entry.sh](https://github.com/openshift-pipelines/pipelines-service/pull/42/files#diff-3af8d6cce9c8e161d65d90a8bd0492b3348d7b8adc532d6d4e91d003a7100e5e)

[gitops/ckcp/base/kustomization.yaml](https://github.com/openshift-pipelines/pipelines-service/pull/42/files#diff-248527507d007bac5bd9b77465bb9250daf8e015dd32ca5e32a3cf9c3914532c)

I think we have no use for our own ckcp instance anymore. It is running with a very old version of kcp which does not have the concepts of workspaces, workloadclusters etc. 
We can deprecate this and maintain only kcp shared instance one as discussed in the calls with @sbose78.

PR #44, opened 2022-04-11
Use the public KCP image for local development

Roming22:
Now that KCP provides a public image, it has become unnecessary
for the users to build and upload their own images.

@fgiloux, @bnallapeta This PR is a first step to bring a development environment that's easy to setup for users, and cleaning up the repo from deprecated files.
I plan on working on the following items once the deployment on our KCP staging is completed:

- Remove most of the code in ckcp/deploy.sh and keep only the ckcp and argocd deployment, then use kcp/deploy.sh to setup the cluster as if it was a staging cluster.
- Rewrite/move ckcp/test.sh to kcp/.
- Deprecate the local/ folder.


bnallapeta (approved):
Looks good. Just one question. I take it that you have not actually run ckcp container pod locally with the image 'ghcr.io/kcp-dev/kcp:976a437' because obviously we are targeting the shared instance first.
Do you think we should hold off on merging this PR onto main until we know its working? Or is it fine to merge with a note that says ckcp pod does not work for now?

gitops/ckcp/base/deployment.yaml
```diff
@@ -16,8 +16,19 @@ spec:
...
+          command: [
+            "/bin/sh",
+            "-c",
+            "rm -rf .kcp;
+            ./kcp/bin/kcp start \
+              --push-mode=true \
+              --pull-mode=false \
+              --run-controllers \
+              --auto-publish-apis \
+              --resources-to-sync=deployments.apps,statefulsets.apps,pods,services,secrets,persistentvolumeclaims"
```
fgiloux, line 30: With phase 1 approach we want to sync more, starting with:
ingresses.networking.k8s.io,deployments.apps,services,conditions.tekton.dev,pipelines.tekton.dev,pipelineruns.tekton.dev,pipelineresources.tekton.dev,runs.tekton.dev,tasks.tekton.dev,taskruns.tekton.dev

we may letter add resource types for other controllers

gitops/ckcp/base/deployment.yaml
```diff
@@ -16,8 +16,19 @@ spec:
     spec:
       serviceAccountName: anyuid
```
fgiloux, line 17: Why anyuid? Is there a requirement for that?

fgiloux:
Re @bnallapeta comment. I think it makes sense to validate that it is working. One thing is that I quickly checked and did not see any release tag added to the kcp images. It would be trivial to amend their [GitHub action](https://github.com/kcp-dev/kcp/blob/main/.github/workflows/kcp-image.yaml) for that. I may open a pull request for that when I have some time.
We could then follow a release version, which would provide some level of stability and avoid breaking. We can then bump the version when a new kcp release is provided and we have generally moved to it and when we want to pick specific bug fixes

gitops/ckcp/base/deployment.yaml
fgiloux: [Here](https://github.com/kcp-dev/kcp/blob/main/manifest/kcp.yaml#L87) no specific Service Account is configured. The syncer needs however to be able to create namespaces.
Generally I see two approaches and we could have both:
- run everything on a development machine and for that kind & kcp binaries (with etcd) should be enough. kcp could possible be run in a container with podman or docker but it is not a must
- run everything on a kubernetes cluster. I think that it is what this [directory](https://github.com/kcp-dev/kcp/tree/main/manifest) provides. We could leverage it, pin a release version and see what else needs to be done

bnallapeta: @fgiloux If I remember correctly, we had to add the anyuid service account for the deployment because it needed permission to create files
    - 'go build' create kcp executable.
    - .kcp/admin.kubeconfig was created after running 'go run'.

And all this was in our own pod without the kcp image. But now, as you have pointed out, when we move to using the kcp public image, we probably would not need anyuid. We should of course test this out.

gitops/ckcp/base/deployment.yaml
```diff
@@ -16,8 +16,19 @@ spec:
...
       containers:
-        - image: ckcp
+        - image: ghcr.io/kcp-dev/kcp:976a437
```
fgiloux, line 19: This is a commit of the main branch.
I would suggest that we stick to a release branch, currently 0.3 and aim to follow minor releases (every 4 weeks I believe). Obviously this does not prevent people to experiment with newer versions but it should provide confidence than things just work after git clone.
The latest git commit on the 0.3 branch is d2bec95

fgiloux: Created this [PR](https://github.com/kcp-dev/kcp/pull/887) to get the images tagged with branch names... let see whether it gets accepted

gitops/ckcp/base/deployment.yaml
Roming22: Ideally we would notset those resources via the command line, but register them later through KCP, so that we use the same behavior in local dev and on staging. So I'm in favor of reducing this list instead of increasing it.

gitops/ckcp/base/deployment.yaml
Roming22: I personally prefer when dependencies are frozen and do not change unless an action happens on the repo. It avoids things breaking for unknown reasons, and having 2 developers starting from the same clone to have different behaviors.
Not a big deal though, as any decision we take on this matter can easily be changed in the future.

gitops/ckcp/base/deployment.yaml
fgiloux: We need first to have something that works with the current state of the world.
When things have progressed in kcp and the syncer pull mode is available we will need to rework it by any way

Roming22:
I agree that those changes should be validated before being merged. I submitted the PR so you had an idea of the direction I was taking.

gitops/ckcp/base/deployment.yaml
fgiloux: I am not saying the opposite. I am saying two things:
- we should use release versions, i.e: d2bec95
- having the images with an additional label in ghcr helps with picking a suitable image without the need to check the git commits

Roming22:
Closing this PR as @psturc will be having a look at it.

PR #51, opened 2022-04-28
Refactor custom images build

Roming22:
This patch cleans up the root of the repo and add scripts to help the user build the custom images.


Roming22:
@bnallapeta While the scripts build the image just fine, the image built does not seem to work. I stumble across an error we've already seen:
```
{"level":"error","ts":"2022-04-28T15:44:51.414Z","logger":"eventlistener","caller":"sink/sink.go:324","msg":"Post \"http://tekton-triggers-core-interceptors.kcp0f67c169c0584384a21b270e6f72de49b6b193151d1784254b1258f3.svc:80/github\": dial tcp: lookup tekton-triggers-core-interceptors.kcp0f67c169c0584384a21b270e6f72de49b6b193151d1784254b1258f3.svc on 172.30.0.10:53: no such host","eventlistener":"github-listener","namespace":"default","eventlistenerUID":"43b25874-56f0-434b-81d2-ea82c965b81e","/triggers-eventid":"e6a2de83-340d-443e-aab5-447ea35e4ed8","/trigger":"github-listener","stacktrace":"github.com/tektoncd/triggers/pkg/sink.Sink.processTrigger\n\tgithub.com/tektoncd/triggers/pkg/sink/sink.go:324\ngithub.com/tektoncd/triggers/pkg/sink.Sink.HandleEvent.func1\n\tgithub.com/tektoncd/triggers/pkg/sink/sink.go:166"}
```
IIRC, you knew exactly what to do last time. What was it?

images/eventlistenersink/make.sh
```diff
@@ -0,0 +1,59 @@
...
+
+SCRIPT_DIR="$(
+  cd "$(dirname "$0")" >/dev/null
+  pwd
+)"
+
+WORK_DIR="$SCRIPT_DIR/work"
+
+init() {
+  rm -rf "$WORK_DIR"
```
fgiloux, line 29: I have seen that many times but I always feel uneasy. If you are unlucky enough to source this bash in a script where there is a "work" directory it's gone... not just theoretical it happened to me (not with this script).
I prefer the use of mktemp -d

images/eventlistenersink/make.sh
```diff
@@ -0,0 +1,59 @@
...
+  mkdir -p "$WORK_DIR"
+}
+
+checkout_upstream_repo() {
+  git clone "https://github.com/tektoncd/triggers.git" "$WORK_DIR/upstream"
```
fgiloux, line 34: it may be nice to avoid doing a clone each time. In another script I offer the option to set the location in an environment variable and only clone if it is not set

images/eventlistenersink/make.sh
```diff
@@ -0,0 +1,59 @@
...
+}
+
+apply_patches() {
+  for patch in "$SCRIPT_DIR/patches"/*; do
+    git apply "$patch"
+  done
+}
+
+publish() {
+  cd cmd/eventlistenersink
```
fgiloux, line 46: pushd and popd may help with avoiding side effects

images/eventlistenersink/make.sh
fgiloux: Addendum: I see at the end of the script that you check BASH_SOURCE, why not using it here instead of $0?

fgiloux (commented):
This is a shallow review. I have not tried running the build

images/eventlistenersink/make.sh
Roming22: Done

images/eventlistenersink/make.sh
Roming22: Done

fgiloux:
Thanks Romain for the changes it looks good to moe

images/eventlistenersink/make.sh
```diff
@@ -0,0 +1,117 @@
...
+  mkdir -p "$WORK_DIR"
+  UPSTREAM_DIR="$WORK_DIR/upstream"
+}
+
+checkout_upstream_repo() {
+  if [ ! -e "$UPSTREAM_DIR" ]; then
+    git clone "https://github.com/tektoncd/triggers.git" "$UPSTREAM_DIR"
+  fi
+  pushd "$UPSTREAM_DIR"
+  git checkout -f "7fbff3b122fcb77d44e1b39bb45c8a935e61f5ed"
```
bnallapeta, line 93: We were using v0.18.0 for Triggers earlier. But this checkout seems to point to v0.19.1. Does everything work well with the newer version too? Did we test it?

images/eventlistenersink/patches/sink.patch
```diff
@@ -15,7 +15,7 @@ index c6d63bfb..633b75a2 100644
  
  		// TODO: Plumb through context from EL
 -		interceptorResponse, err := interceptors.Execute(context.Background(), r.HTTPClient, &request, url.String())
-+		interceptorResponse, err := interceptors.Execute(context.Background(), r.HTTPClient, &request, strings.ReplaceAll(url.String(), ".tekton-pipelines.", ".kcpa9f18e6516b976c21e45eb38fd4291927a3c9dd86fda1b7b7c03ead1."))
++		interceptorResponse, err := interceptors.Execute(context.Background(), r.HTTPClient, &request, strings.ReplaceAll(url.String(), ".tekton-pipelines.", ".kcp0f67c169c0584384a21b270e6f72de49b6b193151d1784254b1258f3."))
```
bnallapeta, line 18: As per our slack convo, this didn't seem to work right?
We need to give the namespace where the interceptor pod is running.

remove-conversion.patch
```diff
@@ -1,64 +0,0 @@
-diff --git a/config/300-pipeline.yaml b/config/300-pipeline.yaml
```
bnallapeta, line 1: Why is this file removed?

images/eventlistenersink/make.sh
Roming22: I use the changeset ID which is currently used to tag the image, so I would not expect any change.

bnallapeta: Since we have used v0.18.0 for Triggers and v0.32 for Pipelines, we could stick with for now. Can checkout this:
https://github.com/tektoncd/triggers/tree/v0.18.0

We can come up with a strategy for incremental upgrades to the newer versions for both pipelines and triggers for the long run.

Roming22: Fixed.

images/eventlistenersink/patches/sink.patch
Roming22: Fixed

remove-conversion.patch
Roming22: It's now a patch in kustomize

bnallapeta (approved):
lgtm

images/eventlistenersink/patches/sink.patch
bnallapeta: It should be this - kcpdf4f8df5a88c8e8d9a7557f1ff8340801b6ad56c1540301151487523

bnallapeta (changes requested):

bnallapeta (approved):

PR #38, opened 2022-04-04
Deploy CRDs on KCP staging instances through ArgoCD

Roming22:
Change summary:

- Modify gitops manifests to always use cluster names. This helps handling the deployment differences between local dev and staging.
- Add `kcp/login.sh` to automate the action of logging in the staging environment.
- Add `kcp/deploy.sh` to automate the deployment to the staging environment

The required resources are now deployed by ArgoCD, except for `triggers-interceptors`. We believe the issue is linked to https://github.com/kcp-dev/kcp/issues/730.

PR #52, opened 2022-04-28
development: Use Rootless Podman

adambkaplan:
- Use ALLOW_ROOTLESS environment variable to enable rootless podman.
- Update detect_container_engine to run rootless podman if desired.
- Use detect_container_engine function in building ckcp image.
- Use docker.io image references in ckcp Dockerfile.

Signed-off-by: Adam Kaplan <adam.kaplan@redhat.com>

### DIFF
--- a/ckcp/docker/ckcp/Dockerfile
+++ b/ckcp/docker/ckcp/Dockerfile
@@ -1,5 +1,5 @@
 #use golang as base image and compile kcp
-FROM golang:1.17 as src
+FROM docker.io/golang:1.17 as src
 
 WORKDIR /go/src/app
 RUN git clone https://github.com/kcp-dev/kcp.git && cd ./kcp && git checkout 986710c754ed0dac9ae1525661de931e5dd7c0cc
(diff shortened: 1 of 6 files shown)


ckcp/docker/ckcp/Dockerfile
```diff
@@ -1,13 +1,13 @@
 #use golang as base image and compile kcp
```
bnallapeta, line 1: This file (and the ckcp folder) is an older version of self hosted kcp we were using earlier (before the shared instance came into picture). After the merge of #38, we plan to do a clean up of the repo.

bnallapeta (commented):
[This](https://github.com/openshift-pipelines/pipelines-service/blob/2ffa8779d54e464e62e1e8a8a6ebafc8d8128cc8/ckcp/docker/ckcp/Dockerfile) file (and the ckcp folder) is an older version of self hosted kcp we were using earlier (before the shared instance came into picture). After the merge of https://github.com/openshift-pipelines/pipelines-service/pull/38, we plan to do a clean up of this folder and others in the repo.

PR #55, opened 2022-05-02
Improve kcp scripts behavior when sourced

Roming22:


kcp/deploy.sh
```diff
@@ -45,7 +45,6 @@ configure_pipeline_cluster() {
 
   echo "  - Service account for connecting KCP to the plnsvc cluster:"
   plnsvc_config kubectl apply -f "$SCRIPT_DIR/manifests/plnsvc/kcp-manager.yaml"
-  get_context plnsvc_config kcp plnsvc kcp-manager "$KUBECONFIG_PLNSVC_KCP"
```
bnallapeta, line 48: So removing this line essentially means that the $KUBECONFIG_PLNSVC_KCP is no more created on the fly but kcp-manager.yaml should exist before we run the script right?
Could you please help me understand why we're removing this now?

Roming22: This kubeconfig is not used in the script. I had to generate it at some point for troubleshooting, but it's now useless.

adambkaplan (approved):
/lgtm

bnallapeta (approved):

PR #56, opened 2022-05-02
Installation of Argo CD as part of the development setup

fgiloux:
This documents and automates the installation of Argo CD in the development environment.
This is a first step towards getting Pipelines Service entirely deployed and managed through GitOps.

### DIFF
--- a/local/argocd/setup.sh
+++ b/local/argocd/setup.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 The pipelines-service Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# Usage: KUBECONFIG="path-to-kubeconfig" ./setup.sh
+
+parent_path=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
+pushd "$parent_path"
+
+printf "Installing Argo CD\n"
+
+kubectl --kubeconfig ${KUBECONFIG} create namespace argocd
+kubectl --kubeconfig ${KUBECONFIG} apply -n argocd -f https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/install.yaml
+i=0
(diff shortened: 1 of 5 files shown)


local/kind/setup.sh
```diff
@@ -69,6 +69,13 @@ for cluster in "${CLUSTERS[@]}"; do
         KIND_EXPERIMENTAL_PROVIDER=${KIND_EXPERIMENTAL_PROVIDER:-} ${KIND_CMD} create cluster \
             --config "${TMP_DIR}/${cluster}.config" \
             --kubeconfig "${TMP_DIR}/${cluster}.kubeconfig"
+
+	if [[  ${KIND_CMD} == "sudo kind" ]]; then
+		sudo chmod +r "${TMP_DIR}/${cluster}.kubeconfig"
+	fi
+
+	printf "Provisioning ingress router in ${cluster}\n"
```
Roming22, line 77: Why use of `printf` over `echo` for simple strings?

local/argocd/setup.sh
```diff
@@ -0,0 +1,43 @@
...
+	i=$((i+1))
+	if [ $i -gt 60 ]; then
+		printf "Ingress deployment not ready in time\n"
+		exit 1
+	fi
+done
+kubectl --kubeconfig ${KUBECONFIG} apply -f ingress.yaml
+
+printf "Argo CD installed\n"
+
```
Roming22, line 42: Could you also log the user in with `argocd --config "${TMP_DIR}/argocd.yaml"`?
That will create a necessary file for running `deploy.sh`.

local/kind/setup.sh
```diff
@@ -79,4 +86,9 @@ for cluster in "${CLUSTERS[@]}"; do
 
 done
 
+NO_ARGOCD="${NO_ARGOCD:-}"
+if [[ "${NO_ARGOCD,,}" != "true" ]]; then
+	KUBECONFIG="${TMP_DIR}/${CLUSTERS[0]}.kubeconfig" ../argocd/setup.sh
+fi
+
```
Roming22, line 93: Running `deploy.sh` requires the presence of `plnsvc.yaml` which is the kubeconfig on which to install the pipeline service.
Would you mind adding `ln -sf "${TMP_DIR}/${CLUSTERS[0]}.kubeconfig" "${TMP_DIR}/plnsvc.yaml"`?

local/kind/setup.sh
fgiloux: [more standardized/consistent](. https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo). I try to favor printf over echo.

local/argocd/setup.sh
fgiloux: I am currently looking at this part. My current thinking is that it does not belong here. This should be limited to argoCD installation.

local/kind/setup.sh
fgiloux: With the phase 1 approach the Pipelines service operators need to be installed on the workload clusters. There may be more than 1. I am looking at this as well. The integration with next steps can come in a separate PR.

Roming22:
Integration with deploy.sh will be handled as a separate PR.

PR #59, opened 2022-05-04
Upgraded kcp to release-0.4 for the local dev env

bnallapeta:
With the 0.4 release, removed the flags that are no more in use.

### DIFF
--- a/local/kcp/start.sh
+++ b/local/kcp/start.sh
@@ -40,7 +40,7 @@ kcp-binaries () {
     git clone https://github.com/kcp-dev/kcp.git
     KCP_DIR="${KCP_PARENT_DIR}/kcp"
     pushd kcp
-    KCP_BRANCH="${KCP_BRANCH:-release-0.3}"
+    KCP_BRANCH="${KCP_BRANCH:-release-0.4}"
     git checkout "${KCP_BRANCH}"
     make build
     popd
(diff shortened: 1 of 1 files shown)


adambkaplan (approved):
/lgtm

I may follow up on this PR to use `go install` to download & install the kcp binary. So far it looks like we only need `kcp start`.

Roming22 (approved):

bnallapeta:
> /lgtm
> 
> I may follow up on this PR to use `go install` to download & install the kcp binary. So far it looks like we only need `kcp start`.

@adambkaplan Yes, we only need kcp start (same goes for kcp running inside a pod too).
The kcp binary is generated from [this](https://github.com/openshift-pipelines/pipelines-service/blob/main/local/kcp/start.sh#L45) line. (The `make build` command triggers a `go build $params` command which generates the kcp binary)

fgiloux:
> > /lgtm
> > I may follow up on this PR to use `go install` to download & install the kcp binary. So far it looks like we only need `kcp start`.
> 

yes that's a good point. However:
- kcp uses "replaces" in go.mod, which seems to get into the way of go install github.com/kcp-dev/kcp@version
- the existing approach allowed to follow a release branch (the prototype versions: 0.3 and now 0.4). Patches to release branches are not consistently tagged in kcp and with go install it is not possible to install binaries for a specific branch. Although, we may be fine without that.

PR #60, opened 2022-05-04
Remove references to the login script from the kcp README

Roming22:

### DIFF
--- a/kcp/README.MD
+++ b/kcp/README.MD
@@ -1,22 +1,18 @@
 # Pipeline Services on managed KCP
 
 ## Clusters access
-The `login.sh -e staging` script will log you in into the various systems:
-1. KCP
-2. KCP workspace
-3. pipeline cluster
-4. ArgoCD
+Create a directory with the following files:
+* `argocd.yaml`: kubeconfig with the active context connecting to the argocd instance;
+* `kcp.yaml`: kubeconfig with the active context connecting to your KCP workspace;
+* `plnsvc.yaml`: the kubeconfig with with the active context connecting to the Pipeline
+  Service cluster.
 
-The script will create various config files in `kcp/work`.
+This directory will be referenced as `KUBECONFIG_DIR`.
 
-Some tokens have a limited lifetime. If you haven't login for a while, you might start
-seeing errors related to authentication/permissions. Running the script again will
-refresh the required credentials.
-
-See `login.sh --help` for available options.
-
-### ArgoCD
-No setup is required.
+The first time one of the deploy/test/reset script is run, use `-k $KUBECONFIG_DIR`.
+This will create a local copy of those files in `kcp/work/kubeconfig`. Those files will
+then be used by default for every command, until you overwrite their value by specifying
+a new `-k $KUBECONFIG_DIR` parameter.
 
 ## Clusters configuration
 The `deploy.sh -e staging` script will configure the various systems in the staging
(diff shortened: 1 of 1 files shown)


kcp/README.MD
```diff
@@ -1,22 +1,17 @@
 # Pipeline Services on managed KCP
 
 ## Clusters access
-The `login.sh -e staging` script will log you in into the various systems:
-1. KCP
-2. KCP workspace
-3. pipeline cluster
-4. ArgoCD
+Create a directory with the following files:
```
bnallapeta, line 4: Created a directory called `pipelines-service/kcp/work/kubeconfig` with the following files:

kcp/README.MD
```diff
@@ -1,22 +1,17 @@
...
+* `argocd.yaml`: kubeconfig with the active context connecting to the argocd instance;
+* `kcp.yaml`: kubeconfig with the active context connecting to your KCP workspace;
+* `plnsvc.yaml`: the kubeconfig with with the active context connecting to the plnsvc cluster.
 
-The script will create various config files in `kcp/work`.
+This directory will referenced as `KUBECONFIG_DIR`.
```
bnallapeta, line 9: will be referenced*

kcp/README.MD
```diff
@@ -1,22 +1,17 @@
...
 
-Some tokens have a limited lifetime. If you haven't login for a while, you might start
-seeing errors related to authentication/permissions. Running the script again will
-refresh the required credentials.
-
-See `login.sh --help` for available options.
-
-### ArgoCD
-No setup is required.
+The first time one of the deploy/test/reset command is run, use `-k $KUBECONFIG_DIR`.
```
bnallapeta, line 11: deploy/test/reset **script** is run,

kcp/README.MD
```diff
@@ -1,22 +1,17 @@
...
+This will create a local copy of those files in `kcp/work/kubeconfig`. Those files will
+then be used by default for every command, until you overwrite their value by specifying
+a new `-k $KUBECONFIG_DIR` parameter.
```
bnallapeta, line 14: If the above suggestion is taken, then we can remove this.

bnallapeta (commented):
Minor changes.

kcp/README.MD
```diff
@@ -1,22 +1,17 @@
...
 ## Clusters access
-The `login.sh -e staging` script will log you in into the various systems:
-1. KCP
-2. KCP workspace
-3. pipeline cluster
-4. ArgoCD
+Create a directory with the following files:
+* `argocd.yaml`: kubeconfig with the active context connecting to the argocd instance;
+* `kcp.yaml`: kubeconfig with the active context connecting to your KCP workspace;
+* `plnsvc.yaml`: the kubeconfig with with the active context connecting to the plnsvc cluster.
```
fgiloux, line 7: This is assuming that there is a single workload cluster. We need to have something generic that can work with any number of workload clusters. Also, this is assuming that the reader knows what plnsvc cluster is but it is not explained.

fgiloux:
I am not sure of the general approach. One thing I would like is that we have something modular. Today everything is defined in deploy.sh. I am not against having a script doing the full deployment but it should source other scripts providing units of work. It is then easier for instance to add a workload cluster to an existing infrastructure.

kcp/README.MD
Roming22: I'm documenting how it currently works, the user could put the files under the work dir. However our login script create the appropriate dir, and I'm often nuking my work dirs, so it would be a pain to have to relogin every time.

kcp/README.MD
Roming22: Fixed

kcp/README.MD
Roming22: Fixed

kcp/README.MD
Roming22: See my response to your comment.

kcp/README.MD
Roming22: I'm documenting how it currently works. It certainly has short comings at the moment. The work we're currently doing should take care of handling multiple clusters.

I've replaced `plnsvc` by the full name.

Roming22:
> I am not sure of the general approach. One thing I would like is that we have something modular. Today everything is defined in deploy.sh. I am not against having a script doing the full deployment but it should source other scripts providing units of work. It is then easier for instance to add a workload cluster to an existing infrastructure.

As mentioned elsewhere, this PR is to update the README so it reflect what we currently have in the repo. I agree that changes will need to happen.

bnallapeta (approved):
Merging the PR as these intermediate changes looks good to me. Of course, as already discussed, the contents of this README and the kcp folder is changing in the already open PRs or soon to be opened PRs. Until they are merged, let's have the README reflect the latest information (as latest as it could be :) ).

PR #62, opened 2022-05-06
Document inotify issue deploying ingress

adambkaplan:
Ingress deployments on the local kind clusters can fail due to the
kernel limiting the number of file watch instances. This can be fixed by increasing
the number of file watches and instances for the machine.

Signed-off-by: Adam Kaplan <adam.kaplan@redhat.com>

### DIFF
--- a/local/kind/setup.sh
+++ b/local/kind/setup.sh
@@ -19,26 +19,25 @@ set -o nounset
 set -o pipefail
 
 prechecks () {
-  if ! command -v kind &> /dev/null
-  then
-    printf "Kind could not be found\n"
-    exit 1
-  fi
+    if ! command -v kind &> /dev/null; then
+        printf "Kind could not be found\n"
+        exit 1
+    fi
 
-  if [[ ${ALLOW_ROOTLESS} == "true" ]]; then
-    echo "Rootless mode enabled"
-  fi
+    if [[ "${ALLOW_ROOTLESS}" == "true" ]]; then
+        echo "Rootless mode enabled"
+    fi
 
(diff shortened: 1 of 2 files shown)


local/kind/setup.sh
```diff
@@ -75,12 +74,12 @@ for cluster in "${CLUSTERS[@]}"; do
             --config "${TMP_DIR}/${cluster}.config" \
             --kubeconfig "${TMP_DIR}/${cluster}.kubeconfig"
 
-	if [[  ${KIND_CMD} == "sudo kind" ]]; then
-		sudo chmod +r "${TMP_DIR}/${cluster}.kubeconfig"
-	fi
+        if [[  ${KIND_CMD} == "sudo kind" ]]; then
```
Roming22, line 77: You have a double space after `[[`

local/kind/setup.sh
```diff
@@ -19,26 +19,25 @@ set -o nounset
...
-    KIND_CMD="sudo kind"
-  else
-    KIND_CMD="kind"
-  fi
+    if ! command -v kind &> /dev/null; then
+        printf "Kind could not be found\n"
+        exit 1
+    fi
+
+    if [[ ${ALLOW_ROOTLESS} == "true" ]]; then
```
Roming22, line 27: `${ALLOW_ROOTLESS}` should probably be in quotes.

Roming22:
LGTM, but since kind is not working for me, I can't test it. @fgiloux would you mind validating the PR and merging it if it does not break anything?

fgiloux:
@Roming22 I don't think documentation will break anything and bash is not indentation-sensitive :-) 
@adambkaplan thanks for helping. inotify is not supported on RHEL 8 but if the information you provide helps on Fedora that's great. Could you please rebase? I can then merge the PR. Thanks again.

PR #63, opened 2022-05-07
Pipelines as Code: installation and repository registration

fgiloux:
Add manifests, a script and instructions for the installation of Pipelines as Code and the registration of repositories.

Goal:
Pipelines as Code can be leveraged for automating the registration of clusters to Argo CD and kcp when their kubeconfig file gets added to a git repository, following a GitOps approach.

### DIFF
--- a/gitops/pac/setup.sh
+++ b/gitops/pac/setup.sh
@@ -1,0 +1,166 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 The pipelines-service Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+usage() {
+
+    printf "Usage: KUBECONFIG="path-to-kubeconfig" GITOPS_REPO="https://gitops.com/group/project" GIT_TOKEN="XXXXXXXXX" WEBHOOK_SECRET="YYYYYYYYYY" ./setup.sh\n\n"
+	
+    # Parameters
+    printf "The following parameters need to be passed to the script:\n"
+    printf "KUBECONFIG: the path to the kubeconfig file used to connect to the cluster where Pipelines as Code will be installed\n"
+    printf "GITOPS_REPO: the repository for which Pipelines as Code needs to be set up\n"
+    printf "GIT_TOKEN: personal access token to the git repository\n"
+    printf "WEBHOOK_SECRET: secret configured on the webhook used to validate the payload\n"
+    printf "CONTROLLER_INSTALL (optional): the controller will be installed only if it is set to true. Pipelines as Code are installed together with the OpenShift Pipelines operator\n"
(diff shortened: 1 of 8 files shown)


bnallapeta:
@fgiloux Is this part of [PLNSRVCE-30](https://issues.redhat.com/browse/PLNSRVCE-30)?

Slack correspondence - https://coreos.slack.com/archives/C032EJ007C0/p1651643427510489

fgiloux:
@bnallapeta no this is not part of PLNSRVCE-30
There is a chicken and egg thing here:
We want to automate the registration of workload clusters into kcp. Therefore we need these clusters registered in a CD tool. Argo CD was chosen as first candidate. For Argo CD to be able to manage the clusters they need first to be added to its managed clusters. Personally I think that it would have been reasonable to keep this point as a pre-requisite for now.  This was what we discussed during our last call. It was decided to have now the registration triggered from git. There is a need of a tool to support this mechanism and it is what Pipelines as Code provides and the reason for this PR

bnallapeta:
Got it. Thanks for the explanation.

Roming22 (approved):
I'm not very comfortable with PaC yet, but that LGTM. @bnallapeta please merge if you're OK with the PR as well.

gitops/pac/setup.sh
```diff
@@ -0,0 +1,166 @@
...
+
+    printf "Usage: KUBECONFIG="path-to-kubeconfig" GITOPS_REPO="https://gitops.com/group/project" GIT_TOKEN="XXXXXXXXX" WEBHOOK_SECRET="YYYYYYYYYY" ./setup.sh\n\n"
+	
+    # Parameters
+    printf "The following parameters need to be passed to the script:\n"
+    printf "KUBECONFIG: the path to the kubeconfig file used to connect to the cluster where Pipelines as Code will be installed\n"
+    printf "GITOPS_REPO: the repository for which Pipelines as Code needs to be set up\n"
+    printf "GIT_TOKEN: personal access token to the git repository\n"
+    printf "WEBHOOK_SECRET: secret configured on the webhook used to validate the payload\n"
+    printf "OPERATOR_INSTALL (optional): the operator will be installed only if it is set to true. Pipelines as Code are installed together with the OpenShift Pipelines operator\n"
```
bnallapeta, line 31: It says that operator (I assume you're talking about [Red Hat Openshift Pipelines Operator](https://docs.openshift.com/container-platform/4.10/cicd/pipelines/installing-pipelines.html)) will be installed.

But in the operator_install function below at line 62, you're not installing the operator. You are manually installing just Pipelines as Code.

gitops/pac/setup.sh
```diff
@@ -0,0 +1,166 @@
...
+    cat <<EOF > ${TMP_DIR}/patch-secret.yaml
+- op: replace
+  path: /data/provider.token
+  value: ${GIT_TOKEN}
+- op: replace
+  path: /data/webhook.secret
+  value: ${WEBHOOK_SECRET}
+EOF
+
+    if [[ "${OPERATOR_INSTALL,,}" == "true" ]]; then
```
bnallapeta, line 97: why the commas after OPERATOR_INSTALL?

fgiloux: This is for making the comparison case insensitive. Apparently this is not working well with MacOS I will change it to:
 $(tr '[:upper:]' '[:lower:]' <<< "$OPERATOR_INSTALL")

gitops/pac/setup.sh
fgiloux: Good point I will rephrase it

bnallapeta: Would you be rephrasing it in the other places too? I still see 4 occurrences of 'operator_install' in the PR.

fgiloux: Sorry I should have been more diligent. It should be fixed now

bnallapeta: No worries. Thank you for the changes.

bnallapeta (approved):
LGTM. Merging.

PR #66, opened 2022-05-12
Implement new gitops model

Roming22:
This PR creates the necessary manifests to move to the gitops model.


gitops/plnsvc.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  namespace: openshift-gitops
+  name: plnsvc
+
```
fgiloux, line 6: We need a description for what this application does

gitops/plnsvc/base/kcp-manager.yaml
```diff
@@ -0,0 +1,54 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: plnsvc
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: kcp-manager
```
fgiloux, line 10: I am getting confused. The ServiceAccount name is kcp-manager but the name of the ClusterRole and its binding kcp-syncer and kcp-syncer-binding. What is that for?

gitops/plnsvc/base/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - https://raw.githubusercontent.com/roming22/pipelines-service/my/gitops/gitops/plnsvc/base/namespace.yaml
+  - https://raw.githubusercontent.com/roming22/pipelines-service/my/gitops/gitops/plnsvc/base/kcp-manager.yaml
+
```
fgiloux, line 6: same as previously:
- Why is it pointing to your own repo?
- Why are the files not directly in the directory?

gitops/plnsvc/base/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - https://raw.githubusercontent.com/roming22/pipelines-service/my/gitops/gitops/plnsvc/base/namespace.yaml
+  - https://raw.githubusercontent.com/roming22/pipelines-service/my/gitops/gitops/plnsvc/base/kcp-manager.yaml
```
fgiloux, line 5: I am not following the intent. Some description would be welcome

fgiloux (commented):
I am having difficulties to follow

gitops/kcp/base/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - https://raw.githubusercontent.com/openshift-pipelines/pipelines-service/main/gitops/triggers/triggers-crds/base/namespace.yaml
```
bnallapeta, line 4: I see just the triggers CRDs. What about pipelines CRDs?

gitops/kcp/base/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
...
+  - https://raw.githubusercontent.com/roming22/pipelines-service/my/gitops/gitops/kcp/base/plnsvc-manager.yaml
```
bnallapeta, line 5: Is providing your git repo (roming22) intentional for this PR? Would you be updating this to upstream repo later on?

gitops/plnsvc/base/kustomization.yaml
bnallapeta, line 5: Same comment as above :)

gitops/plnsvc/base/kcp-manager.yaml
Roming22: kcp-manager would be the SA that manages everything related to KCP. This currently only means the syncer, for which the appropriate role and binding are created.

If that's not a good model, let me know and I'll rename the SA to kcp-syncer.

gitops/plnsvc/base/kustomization.yaml
Roming22: Fixed

gitops/kcp/base/kustomization.yaml
Roming22: That's a mistake. I should have moved the resource locally.

This does instanciate the CRDs. The latest revision should make more sense.

gitops/kcp/base/kustomization.yaml
Roming22: Fixed

gitops/kcp/base/plnsvc-manager.yaml
```diff
@@ -0,0 +1,131 @@
+---
```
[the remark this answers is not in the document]
Roming22: Because the life cycle of these resources are currently tied to one another and having everything in one file makes it easy to show that relationship.

If I had the following files [namespace.default.yaml, kcp-manager.yaml, kcp-syncer.yaml, kcp-syncer-binding.yaml], one could remove kcp-manager.yaml and not know that the other files should be cleaned up as well.

If that's an anti-pattern and I'll split it.

gitops/plnsvc/base/kustomization.yaml
Roming22: The description has been added to the Application definition.
The reference to an external repository was an oversight on my part.

gitops/plnsvc/base/kcp-manager.yaml
fgiloux: There should not be much more than the syncer agent. I think that it would be good that the names align in one way or the other. I would vote for having the SA named kcp-syncer.

gitops/kcp/base/plnsvc-manager.yaml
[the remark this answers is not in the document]
fgiloux: Isn't it what kustomization.yaml or Argo CD Application bring?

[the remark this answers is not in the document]
Roming22: My understanding is that ArgoCD/Kustomize ties all the resources for a single application together, but that it does not allow to group a subset of those resources together.

Roming22:
Marked as a draft as there's a lot of overlap with PR67, and the latter has precedence.

Roming22:
This PR is no longer relevant. A new PR will be created to implement the new GitOps model.

PR #71, opened 2022-05-17
local dev: kubeconfig with IP instead of localhost

fgiloux:
For local development
- Added logic to have a kubeconfig file, which points to an IP, which is accessible from the KinD network instead of localhost.
This is useful when the cluster for instance needs to be registered to an Argo CD server running on a KinD cluster.
- Removed the generation of manifests for workloadclusters as in pull mode they are now generated by the syncer command.


fgiloux:
I have tested with podman on Linux. It would be great if someone can confirm that it works with MacOS and with docker.

psturc:
👀

Roming22 (approved):

PR #74, opened 2022-05-19
Remove port number from cluster name in register.sh

Roming22:
The cluster name has the port number when specified in the kubeconfig. Since this impact the naming of the folders, it's probably better to ignore the port when present.


bnallapeta (commented):
@Roming22 Seems like #75 supersedes this one. Let me know if this can be closed.

fgiloux:
Closing as we don't need the same change in two PRs

PR #75, opened 2022-05-19
Fix/argocd cluster name

Roming22:
The clusters are currently all added as `pac-manager`. This is likely to cause issues when registering multiple clusters, and it does not provide useful information to the admin to identify the cluster from its name.


bnallapeta (commented):
@Roming22 
1. How is it taking the name pac-manager by default?
2. There seems to be some conflict with the latest register.sh. Could you please resolve them?

fgiloux:
> 1.  How is it taking the name pac-manager by default?

The context name was used ( I thought I had fixed that). It is correct to use the cluster name instead
 
> 2.  There seems to be some conflict with the latest register.sh. Could you please resolve them?

This was missing the last commit where the context is specified for applying changes to the registered clusters

fgiloux:
>     1. How is it taking the name pac-manager by default?

It is correct that it should take the cluster name for the registration and not the context name. I thought it was the case.

>     2. There seems to be some conflict with the latest register.sh. Could you please resolve them?

PR #76, opened 2022-05-23
Refactor the gitops folder

Roming22:
The new structure should make it easier to navigate the various
configuration files and what they control.


bnallapeta (commented):
lgtm in general. Although, I'm not sure what to test here (I guess the whole of pipelines-service :p).

I do have one suggestion. Create a new folder called argocd-apps and put all the Application.yaml files in there.

pipelines-service/tree/my/refactor_gitops/gitops/argocd
  - argocd-apps
    - ckcp.yaml
    - pipelines-controller.yaml
    - pipelines-crds.yaml
    - triggers-controller.yaml
    - triggers-crds.yaml
    - triggers-interceptors.yaml

And of course, with this change, if needed, the respective paths need to be updated whereever these files are called.

Roming22:
> I do have one suggestion. Create a new folder called argocd-apps and put all the Application.yaml files in there.
> And of course, with this change, if needed, the respective paths need to be updated whereever these files are called.
Done

bnallapeta (approved):
LGTM. 

@fgiloux Please take a look at the PR and take this to closure if you are okay with the changes.

fgiloux:
@bnallapeta @Roming22 I have been through the PR. I will share my understanding of the repo structure. Please let me know if I am missing something.
- I see what is below sre/environment as the "gitops" repository in the sense that it contains the manifests that become the "base" of the "gitops" repository that an SRE team would create and manage
- I see that argocd-apps needs to be outside of this tree as it contains what SRE would use to create Argo CD apps that refer to the SRE gitops repository
- My understanding is that gitops/ckcp contains what is required to create a development environment with kcp running as a pod on a cluster. IMHO I don't think it is the best place for it as I don't see why it is under a directory called gitops and it does not reflect that it is not, at least currently, our deployment model for production. As it is not something modified by this PR we can let that aside.
- My feeling is that the rest under gitops should at some point either be removed or migrate into sre/environment

Now specifically on the changes introduced by this PR:
- gitops/argocd: Maybe that's fine for now but if/when (see my understanding above) we have moved things I feel like we would be left with a single branch gitops/argocd/argocd-apps in that case we would not need at least one level.
- sre: On a clean sheet I would rather call it gitops... but there is already a gitops directory. I think it is fine for now.

Please confirm whether we have a shared understanding. If that's the case I am fine with merging this PR with the assumption that we will go on with moving things bit by bit afterwards.

Roming22:
> * I see what is below sre/environment as the "gitops" repository in the sense that it contains the manifests that become the "base" of the "gitops" repository that an SRE team would create and manage
Correct. `gitops/sre` is the scaffold that has to be copied by an SRE team to create a new environment for them to manage.

> * I see that argocd-apps needs to be outside of this tree as it contains what SRE would use to create Argo CD apps that refer to the SRE gitops repository
Correct. `argocd-apps` contains the common configuration to all environment, managed by the upstream project.

> * My understanding is that gitops/ckcp contains what is required to create a development environment with kcp running as a pod on a cluster. IMHO I don't think it is the best place for it as I don't see why it is under a directory called gitops and it does not reflect that it is not, at least currently, our deployment model for production. As it is not something modified by this PR we can let that aside.
Agreed. The bit regarding ckcp will need to be refactored, probably when we clean up how we're dealing with local development.

> * My feeling is that the rest under gitops should at some point either be removed or migrate into sre/environment
I'm not sure what you're talking about here. Either you're talking about the last directory under gitops after refactoring, `pac`, in which case I do not understand why this would be moved under `sre/environment`, or you're talking about the remaining gitops apps from before the refactoring, in which case they've all been moved to the argocd folder, and should not be part of the `sre/environment` by default (SRE is then free to customize the configuration anyway they want).

> * gitops/argocd: Maybe that's fine for now but if/when (see my understanding above) we have moved things I feel like we would be left with a single branch gitops/argocd/argocd-apps in that case we would not need at least one level.
I would leave the manifests that are related to argocd under `gitops/argocd`, in order to keep the footprint of the SRE repository as small as possible. As previously mentioned, the SRE repository structure will be flexible enough to allow SRE to customize anything they have to.

> * sre: On a clean sheet I would rather call it gitops... but there is already a gitops directory. I think it is fine for now.
Agreed. Another option, could be to create a dedicated template upstream repository. so the process for SRE would be as simple as cloning and pushing to their new repository.

Please merge if you think we're all on the same page now.

fgiloux (approved):
That's fine. I will merge the PR. Hopefully not too many rebases for other in flight PRs

PR #83, opened 2022-05-30
Overwrite UBI labels with information specific to the image, move Argo CD registration logic to "exploration"

fgiloux:
Similar to what has been done with the "kcp register" image this PR overwrite labels set by the UBI image with information specific to this image.
It also renames the directory, so that it has the same pattern as the "kcp register" image.
 

bnallapeta (commented):
The label changes in the PR are good. But, do we need this argocd-registrar image at all now? The plan is to ask the customer to install ArgoCD and for us to install pipelines on each cluster right?

fgiloux:
> The label changes in the PR are good. But, do we need this argocd-registrar image at all now? The plan is to ask the customer to install ArgoCD and for us to install pipelines on each cluster right?

Yes we may start with Argo CD running on each cluster and "in-cluster" management only. In that case we don't need this image.

Later on when we leverage GitOps as a service, this image may be useful again. 

I feel that we should
- document the in-cluster approach
- get this out of the main directories to avoid confusion. Maybe in a directory called exploration. What do you think?

This PR was just about cleaning a bit things beforehands.

fgiloux:
@bnallapeta I have added a second commit to the PR that moves the Argo CD registration logic to an "exploration" directory. Please have a look.

bnallapeta (approved):
LGTM. However, I would like @Roming22 to take a look at the PR and merge if there are no other issues.

Roming22 (approved):

PR #84, opened 2022-05-30
Change event triggering Argo CD registration from pull request to push

fgiloux:
- Similar to the registration to kcp "push" is the right event and not "pull request". The workflow only need to get triggered when the change has been merged.
- Align the permissions on binaries with what is done for the kcp image

### DIFF
--- a/gitops/pac/.tekton/argo-registration.yaml
+++ b/gitops/pac/.tekton/argo-registration.yaml
@@ -4,7 +4,7 @@ metadata:
   name: registration-pipeline
   annotations:
     pipelinesascode.tekton.dev/on-target-branch: "[main]"
-    pipelinesascode.tekton.dev/on-event: "[pull_request]"
+    pipelinesascode.tekton.dev/on-event: "[push]"
 spec:
   params:
     - name: repo_url
--- a/images/argocd/Dockerfile
+++ b/images/argocd/Dockerfile
@@ -5,13 +5,13 @@ RUN mkdir /workspace && chmod 777 /workspace && chown 65532:65532 /workspace
 COPY ./register.sh /usr/local/bin/register.sh
 RUN JQ_VERSION=1.6 &&\
     curl -sSL -o /usr/local/bin/jq https://github.com/stedolan/jq/releases/download/jq-$JQ_VERSION/jq-linux64 && \
-    chmod +x /usr/local/bin/jq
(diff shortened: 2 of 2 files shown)


gitops/pac/.tekton/argo-registration.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: registration-pipeline
   annotations:
     pipelinesascode.tekton.dev/on-target-branch: "[main]"
-    pipelinesascode.tekton.dev/on-event: "[pull_request]"
+    pipelinesascode.tekton.dev/on-event: "[push]"
```
bnallapeta, line 7: I was under the impression that Pull is the right one. Because if it is push, this means that the customer will have to merge their kubeconfig into Github which includes private info like token, cluster names which is not recommended right?

fgiloux: not into GitHub... into their private repository, which may be GitHub but also Bitbucket or GitLab. This is until we have integrated with something like Vault.
The difference between pull_request and push is really the git event in the process. When you create a pull_request, the "pull_request" event is triggered. When the pull request is merged the "push" event is triggered (because that's what it is)

fgiloux: [Here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) it is described what triggers a pull request event.
It seems that it would also be possible to use [type "close"](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-when-a-pull-request-merges) with GitHub but I don't know whether it would work with other git repositories.

bnallapeta: Yes, it could be any of the Git providers.
I think if we are going to use Vault (future) and maintain kubeconfigs on the repo, then better to use a normal Push than pull request on close.
But do you think we could do push for specific files as described [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-affects-specific-files)?

fgiloux: I haven't seen a way to do the filtering in Pipelines as Code but it can surely be done in the PipelineRun within a git task

bnallapeta (approved):
LGTM

PR #85, opened 2022-05-31
PLNSRVCE-188 : Conform to the mid term architecture

bnallapeta:
This PR contains changes to accomplish this:
- Conforming to the Phase1 architecture.
- Pipelines and Triggers installed out of the box onto the Openshift clusters the customer would provide.
- Uses ArgoCD to do this.
- Currently, a script in place to do this. Would create the Dockerfile and PipelineRun to do the same in a subsequent PR.
- Once that is done, would raise a PR to clean up Gitops folder - remove the older modified pipelines and triggers, argocd apps, patches etc. (probably best to move under exploration folder Frederic will create in another PR)

Local Run:
 ```
[bnr@bnr scripts]$
DATA_DIR=/home/bnr/workspace/pipelines-service ./register.sh
Extracting files under the kubeconfig dir and reading the content in each file 
pcluster.kubeconfig
api-ci-ln-r421z0k-72292-origin-ci-int-gce-dev-rhcloud-com --- default/api-ci-ln-r421z0k-72292-origin-ci-int-gce-dev-rhcloud-com:6443/kube:admin --- pcluster.kubeconfig 
Checking if the GitOps cluster pod is available and Ready..
pod/openshift-gitops-server-78468747f9-v25dg condition met
Installing pipelines and triggers components on the cluster via Openshift GitOps... 
application.argoproj.io/pipelines-service created
Checking if the pipelines pod is available and Ready..........pod/tekton-pipelines-controller-7b7ddb4f77-5xwhs condition met

 Checking if the triggers pod is available and Ready....pod/tekton-triggers-controller-8655f6b4ff-89dm9 condition met

 Checking if the pipelines pod is available and Ready.pod/tekton-triggers-core-interceptors-5b468c98cc-rvfcp condition met

```

### CODE REVIEW

File: images/cluster-setup/register.sh [hunk 1]
```diff
@@ -0,0 +1,147 @@
...
+}
+
+gitops_installed_check() {
+#  printf "Checking if Openshift GitOps is installed on the cluster... \n"
+    for i in "${!clusters[@]}"; do
+      #checking if the gitops server pod is up and running
+      declare podname=""
+      declare -i i=0
+      declare -i timeout=60
+      printf "Checking if the GitOps cluster pod is available and Ready."
```
fgiloux, line 73: I am not sure that it is the best check.
The pod may be temporarily  unavailable (due to an upgrade, due to a node maintenance) and still GitOps installed and working.
The operator resource may be a better fit

fgiloux (commented):
I am glad that this PR has been raised. It is an important step and I appreciate that you also created documentation.
There are few things where I am not 100% clear. When we have progressed on them I will do a new review where I will have, among others, a better look at the script.

gitops/argocd/argocd.yaml
```diff
@@ -0,0 +1,33 @@
...
+  name: pipelines-service
+  namespace: openshift-gitops
+
+spec:
+  destination:
+    namespace: openshift-gitops
+    server: https://kubernetes.default.svc
+  source:
+    path: gitops/argocd/argo-apps
+    repoURL: https://github.com/openshift-pipelines/pipelines-service.git
```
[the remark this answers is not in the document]
bnallapeta: Same as above.

gitops/README.md
```diff
@@ -0,0 +1,54 @@
...
+2. Open and merge a Pull Request
+There are a couple of steps that needs to be done in order to use PAC in pipelines-service.
+a. Install Pipelines as Code on the Openshift Cluster.
+b. Setup a Github application by following the steps [here](https://pipelinesascode.com/docs/install/github_apps/).
+
+Once these two steps are done, the Github app you've setup from step2 is now talking to PAC and would watch PRs to trigger PipelineRun to automate the workflow described so far.
+
+The PR should have:
+- kubeconfig/s of the openshift clusters to be placed under gitops/credentials/kubeconfig/compute
+- overlay/kustomization.yaml to be created under gitops/environment/compute/<clustername>/overlay/kustomization.yaml
```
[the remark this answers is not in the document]
bnallapeta: As discussed, this work would be done post @Roming22 's work on environment and credentials folder re-structuring.

gitops/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+Following the phase1 architecture described [here](../docs/images/phase1.png), we structured GitOps to setup a Openshift cluster, i.e., 
+- Install pipelines and triggers components on the physical cluster via GitOps Application.
+- Log into kcp, register the clusters using the credentials for accessing kcp shared instance.
+
+Post this, user will be able to login to kcp workspace and start creating PipelineRuns :)
+
+## How to run
+
+Currently, we support two ways to get the cluster setup and ready to do:
```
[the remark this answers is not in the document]
bnallapeta: Not sure what I was going for earlier. Have updated it now :)

gitops/README.md
```diff
@@ -0,0 +1,54 @@
+# Pipelines-service
+
+## Why GitOps?
+
+We want to make the onboarding experience to use pipelines-service as easy and customizable as possible. With that in mind, pipelines-service is built around the principles of GitOps. Using kustomize, users will be able to setup, modify and update cluster resources without having to disrupt existing setup. We provide base kustomization.yaml files to help you get started and you can add your customizations in the overlay/kustomization.yaml files.
+
+## Pre-requisites
+- Openshift cluster
+- Openshift GitOps operator installed
+- Pipelines as Code installed and setup to talk to this repo
```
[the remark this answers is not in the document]
bnallapeta: It should be this. Added.
https://docs.openshift.com/container-platform/4.10/cicd/gitops/installing-openshift-gitops.html

images/cluster-setup/register.sh
bnallapeta: Yeah, I had done operator check first, but then had the same thought that the operator resource may be up, but the pod might not be :)
I can do both the checks if that is better.

Besides, I think if you compare just operator vs pods, pods are almost always alive even during upgrades (as k8s takes care of this to always serve traffic).

images/cluster-setup/install.sh
```diff
@@ -0,0 +1,147 @@
...
+        usage
+	      exit 1
+    fi
+}
+
+# populate clusters with the cluster names taken from the kubeconfig
+# populate contexts with the context name taken from the kubeconfig
+# populate kubeconfigs with the associated kubeconfig for each cluster name
+# only consider the first context for a specific cluster
+get_clusters() {
```
Roming22, line 43: That seems a bit complex. Couldn't it be simplified by retrieving the list of clusters like [this](https://github.com/Roming22/pipelines-service/blob/feat/register-gitops/images/argocd/register.sh#L78),

Roming22 (approved):
I have minor comments that should not hold the PR from being merged if it is otherwise ready. All of them can be addressed later on.

images/cluster-setup/install.sh
fgiloux: you can have multiple context and multiple clusters in a kubeconfig file

bnallapeta: I'm having difficulty following your code in the link you shared.
- In the process_clusters() func, why are we checking for kustomization.yaml by name?
- In the same func, we are passing cluster_dir to process_cluster() func. Why are we passing this path?
  cluster_dir returns something like this - /home/bnr/workspace/pipelines-service/gitops/environment/compute/base/argocd-rbac
- And then in process_cluster() func, you are checking for ${cluster_name}.yaml which may not be the case. Users would usually have *.kubeconfig files under credentials.
- Also, I am not sure if it is handling multiple cluster information inside one kubeconfig file.

images/cluster-setup/install.sh
```diff
@@ -0,0 +1,119 @@
...
+    KUBECONFIG="${DATA_DIR}/gitops/credentials/kubeconfig/compute/${kubeconfigs[$i]}" kubectl --context "${contexts[$i]}" apply -f "${DATA_DIR}/gitops/argocd/argocd.yaml"
+  done
+}
+
+postchecks() {
+  #checking if the pipelines and triggers pods are up and running
+  #pipelines controller
+  declare pipelines_podname=""
+  printf "Checking if the pipelines pod is available and Ready"
+  until [[ -n $pipelines_podname ]] ; do
```
fgiloux, line 78: What do you think of having a timeout? Waiting for 10mns should be enough most of the time. Making it configurable through an env variable would be nice for environments were pulling from the internet is very slow.

images/cluster-setup/install.sh
Roming22: @fgiloux I don't believe that having multiple clusters/contexts per kubeconfig would help managing the repository.  If a process can edit a kubeconfig to add/remove a context or cluster, we run into the risk of having a conflict on that file. I think that it would be best to have a single file per cluster/context.

@bnallapeta Every cluster has it's own kubeconfig, under `credential/compute/$cluster_name.yaml`. Every cluster has it's gitops configuration under `environment/compute/.../$cluster_name/kustomization.yaml`. Therefore:

1. To list the clusters that need to be setup, we just need to have the list of directories in `environment/compute` that have a `kustomization.yaml`.
2. $cluster_dir has 2 informations. The name of the cluster we want to setup, and the path holding the `kustomization.yaml`. We need the first info to get the right kubeconfig from `credentials/compute`, and the second to do the `kubectl apply`.
3. I used the yaml extension for the kubeconfig, because that's what they are, That might be a mistake, so let's put that aside. We would however require that for any `environment/compute/.../$cluster_name/kustomization.yaml` there is a corresponding `credential/compute/$cluster_name.[yaml|kubeconfig]`.
4. At the moment I don't see being able to handle multiple clusters/contexts in a single kubeconfig as desirable. As per my comment to Frederic, I think it could open the door to conflict issues as PR opened by humans and robots could start modifying the same area of the same file. With a single file per cluster/context, the most likely scenario is the file will either be added or deleted, and most likely will never be modified.

images/cluster-setup/install.sh
```diff
@@ -0,0 +1,119 @@
...
+get_clusters() {
+    clusters=()
+    contexts=()
+    kubeconfigs=()
+    printf "Extracting files under the kubeconfig dir and reading the content in each file \n"
+    files=("$(ls "$DATA_DIR/gitops/credentials/kubeconfig/compute")")
+    for kubeconfig in "${files[@]}"; do
+        printf "%s\n" "$kubeconfig"
+        subs=("$(KUBECONFIG=${DATA_DIR}/gitops/credentials/kubeconfig/compute/${kubeconfig} kubectl config view -o jsonpath='{range .contexts[*]}{.name}{","}{.context.cluster}{"\n"}{end}')")
+        for sub in "${subs[@]}"; do
```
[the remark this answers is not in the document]
Roming22: I believe they would be under `$DATA_DIR`, where `$DATA_DIR` is a git repository initialized from the content of `gitops/sre`.

images/cluster-setup/install.sh
fgiloux: @Roming22 we are not creating the kubeconfig files. They are provided by the persons or tools making0 the workload clusters available. When you log into multiple clusters or with multiple users without changing the KUBECONFIG environment variable you end up with multiple clusters and contexts in the kubeconfig file. Sure we could put additional restrictions to what is accepted as kubeconfig but I am missing the urge for it as Bharath already has code that can deal with it.

images/cluster-setup/install.sh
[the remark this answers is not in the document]
bnallapeta: Done.

bnallapeta:
@fgiloux @Roming22 I believe I have worked on all the comments in the PR. Please take a look.

gitops/argocd/tektoncd/openshift-operator.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: openshift-pipelines-operator
+  namespace: openshift-operators
```
fgiloux, line 5: No action: I would prefer the operators to be installed in their own namespaces (still separated from operands) rather than openshift-operators. It would require an OperatorGroup first to be created. It would however bring a few benefits:
- better segregation
- dedicated InstallPlan, which provides for instance the possibility to have different upgrade strategy by operator. This is really useful when one runs into troubles.

This can be addressed later on.

fgiloux (changes requested):
What happened to the gateway? We still need it.
As discussed a kustomization.yaml file under gitops/argocd/argo-apps may be helpful

bnallapeta:
> What happened to the gateway? We still need it. As discussed a kustomization.yaml file under gitops/argocd/argo-apps may be helpful

I have added gateway into exploration folder as discussed.
Have also created kustomization under argo-apps.

Local run is successful. Let me know if we are good with the changes. I will do a squash in the mean time.

```
[bnr@bnr pipelines-service]$ DATA_DIR=/home/bnr/workspace/pipelines-service ./images/cluster-setup/install.sh
Extracting files under the kubeconfig dir and reading the content in each file 
pcluster.kubeconfig
api-ci-ln-k9v2ibb-72292-origin-ci-int-gce-dev-rhcloud-com --- default/api-ci-ln-k9v2ibb-72292-origin-ci-int-gce-dev-rhcloud-com:6443/kube:admin --- pcluster.kubeconfig 
Installing tektoncd components on the cluster via Openshift GitOps... 
application.argoproj.io/pipelines-service configured
Checking if the Pipelines pod is available and Ready.......pod/tekton-pipelines-controller-78fb556895-n56td condition met

Checking if the Triggers pod is available and Ready....pod/tekton-triggers-controller-b67cdddb5-zpsts condition met

Checking if the Triggers interceptor pod is available and Ready.pod/tekton-triggers-core-interceptors-866596d594-kr6r9 condition met

```

Roming22 (approved):

fgiloux:
I have successfully tested. Please squash and I can merge. Thanks!

PR #86, opened 2022-05-31
Document the gitops model

Roming22:
A `template` dir is created so it can be used as a starting point
for an SRE team to create their own environment.

An `example` dir is created to show how the gitops repository should
be used.

The `.tekton` dir has been moved from `pac` to the `template` and
`example` directory.


fgiloux:
Hi @Roming22 
I am in disagreement with a couple of things:
- I am not for moving all the manifests that were moved to the SRE directory with your previous PR now to an example directory
- I am not for moving the pipelines as code workflow to an example directory

These things were already structured, part of our platform artifacts and should stay in our platform directories.

Here is how I see things. In the example directory:
- you can have overlays (not the base, or only custom base) that are generally created by people wanting to consume what we provide
- these overlays can:
    - either reference a remote branch: our gitops directory
    - reference local directories if the consumers decided to fork the repository
    
Happy to have a direct discussion with you on that.

bnallapeta (commented):
1. I agree with @fgiloux . Example folder should not contain anything of base. I think we don't need an example folder at all. We could document the same under README.md.
But if you insist, then let's keep the example folder light and should contain just the overlays.
2. .tekton should be under images or let's put it under root (pipelines-service). Not inside template.
3. Logically, I would assume that any and all of the user's configurations should go into one folder - in our case, environment/compute makes sense. So, should the flavors folder go inside environment/compute? 
4. Usually, kubeconfigs are not yaml files. They are either *.config or *.conf or *.kubeconfig. 
Need to change the default files in here - https://github.com/Roming22/pipelines-service/tree/my/refactor_gitops/gitops/sre/example/credentials/kubeconfig/compute
5. It must contain reference to something. Shouldn't be blank:
https://github.com/Roming22/pipelines-service/blob/my/refactor_gitops/gitops/sre/example/flavors/compute/base/kustomization.yaml
6.

Roming22:
Here are my answers, we can talk about it more at length after one of our stand-up.

The `example` directory is there to show how a environment that has been setup correctly would look like, I agree with Bharath that this information could be put in a README.md. Having the file structure allows one to navigate through the "real thing", which I found nice.

The `template` directory is the scaffolding required for an SRE team to create a new environment from scratch. I do not believe that this scaffolding needs to hold any manifest, except for a `flavors/base` that references the default configuration stored on our repository which includes those manifests. The template dir does not contains READMEs or any file that needs to be cleaned up, and does have the the .tekton dir required for CI/CD.

The `.tekton` folder can be removed from example, it was there because it's expected to be there for a properly setup environment.

The rational for moving it to `gitops/sre/template` is that spawning a new environment is as simple as copying the `template` directory somewhere and push the content to a new repository. I agree that the resource is technically a part of `pac` and therefore should be referenced there as well. I would think that adding a symlink in pac would be good enough of a solution.

`environment` contains the unique configuration of each clusters. `flavors` contains configuration shared across clusters. I would therefore like to keep them separate.

Let's ignore me using the "wrong" extension for the moment. That can easily be changed. However we would define that extension, and users/robots would be required to create kubeconfigs that respect the `$cluster_name.$extension` naming convention.

gitops/sre/example/credentials/kubeconfig/compute/README.md
```diff
@@ -0,0 +1,14 @@
+# Clusters credentials 
+
+This directory purpose is to contain the kubeconfig 
+files used to access and control the various clusters.
+
+The filename should match the directory name of
+the cluster in the `environment` directory.
+
```
fgiloux, line 8: no this is not what we have today. The directoy name should match the name of the cluster in the kubeconfig file.

gitops/sre/example/environment/compute/README.md
```diff
@@ -0,0 +1,17 @@
+# Cluster configuration
+
+This directory contains the `kustomization.yaml` that
+controls the various clusters' setup.
+
```
fgiloux, line 5: Not only. It should also contain overlays: patches, additional manifests but there is none here.

gitops/pac/.tekton
```diff
@@ -0,0 +1 @@
+../sre/template/.tekton
```
fgiloux, line 1: The link should be in the other direction. Keep the code where it is, next to the other components it interacts with and references it from the template/example directory.

fgiloux (changes requested):
In my opinion:
- examples should not be in the same gitops/sre directory where we keep our reference manifests. The rationals:
  - SRE teams should use the content of gitops/sre as base, not examples.
  - We need to be careful with changes we make to gitops/sre and how consumers can move from A to B. The stakes are alot lower with example.
- example and template is too much. We will need to keep things up-to-date and people wanting to consume our manifests can either use one or the other.
- examples really should be driven by use cases. What do we expect consumers to change/extend? Let's show them how they can do it.
- in the example directory tree there is no customization. In my opinion that's an important aspect of providing an example
- examples really should be driven by use cases. What do we expect consumers to change/extend? Let's show them how they can do it.
- I don't think that flavors should be part of a template.

fgiloux (commented):
A few minor documentation changes and a question:
Why has gitops/sre/credentials/kubeconfig/README.md been removed? I would have rather provided there additional information on the sub-directories

docs/sre/examples/single_cluster/credentials/kubeconfig/compute/my-cluster.kubeconfig
```diff
@@ -0,0 +1,16 @@
+# Kubeconfig for the clusters hosted by provider2.
```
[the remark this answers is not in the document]
Roming22: Done

docs/sre/examples/providers/credentials/kubeconfig/compute/p1.clusterA.kubeconfig
```diff
@@ -0,0 +1,16 @@
+# Kubeconfig for the clusters hosted by provider2.
```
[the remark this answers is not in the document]
Roming22: Done

PR #88, opened 2022-06-02
Added GitHub workflows to publish images to ghcr

ramessesii2:
## How would the packages look
image

#### ToDo: Add more images to the registry when #85 merges.

### DIFF
--- a/.github/workflows/argocd-registrar-image.yaml
+++ b/.github/workflows/argocd-registrar-image.yaml
@@ -1,0 +1,46 @@
+name: Build and Publish argocd-registrar Image
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "exploration/argocd-external/images/argocd-registrar/**"
+
+jobs:
+  build-push:
+
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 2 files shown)


.github/workflows/registrar-image.yaml
```diff
@@ -0,0 +1,44 @@
+name: Build and Publish kcp-registrar Image
+
+on:
+  push:
+    branches:
```
fgiloux, line 5: Could you add [path filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths)?
A new build is only required when something is modified in images/kcp-registrar

fgiloux (commented):
Thanks Satyam it looks good, just a small modification to limit the number of builds.
When this has been merged we will need to modify the PipelineRun to point to this new image.
Same will need to be done for exploration/argocd-external/images/argocd-registrar
and as you wrote Bharath's PR when it has merged

.github/workflows/registrar-image.yaml
ramessesii2: I'm on it.

PR #91, opened 2022-06-03
PLNSRVCE-188 : Conform to the mid term architecture - Part 2 

bnallapeta:
- This PR follows PR #85.
- While 85 has the script, files, and folders to make the changes to adhere to Phase 1 architecture, this PR has the Dockerfile and the PipelineRun yaml which will automate the workflow using PAC. 

Local Tests:

image

image

### DIFF
--- a/gitops/pac/.tekton/pln-svc-registration.yaml
+++ b/gitops/pac/.tekton/pln-svc-registration.yaml
@@ -1,0 +1,59 @@
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: setup-cluster-pipeline
+  annotations:
+    pipelinesascode.tekton.dev/task: "git-clone"
+    pipelinesascode.tekton.dev/on-event: "[push]"
+    pipelinesascode.tekton.dev/on-target-branch: "[main]"
+spec:
+  params:
+    - name: repo_url
+      value: "{{repo_url}}"
+    - name: revision
+      value: "{{revision}}"
+  pipelineSpec:
+    params:
+      - name: repo_url
+      - name: revision
+    workspaces:
+      - name: source
+    tasks:
+      - name: fetch-repository
+        taskRef:
+          name: git-clone
+        workspaces:
+          - name: output
+            workspace: source
+        params:
+          - name: depth
+            value: "500"
+          - name: url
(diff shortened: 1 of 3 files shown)


images/cluster-setup/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
+WORKDIR /
+RUN mkdir /workspace && chmod 777 /workspace && chown 65532:65532 /workspace
+ENV HOME /tmp/home
+RUN mkdir $HOME && chmod 777 $HOME && chown 65532:65532 $HOME
+COPY ./install.sh /usr/local/bin/install.sh
+RUN KUBE_VERSION=v1.24.0 && \
+    curl -L -o /usr/local/bin/kubectl "https://dl.k8s.io/release/$KUBE_VERSION/bin/linux/amd64/kubectl" && \
+    chmod +x /usr/local/bin/kubectl
```
fgiloux, line 9: Romain was keen on having unambiguous file rights: `chmod 755` instead of `chmod +x`

images/cluster-setup/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
+WORKDIR /
```
fgiloux, line 2: Could you overwrite the labels of the base image and set specifics?
Something like [this](https://github.com/openshift-pipelines/pipelines-service/blob/main/images/kcp-registrar/Dockerfile#L12-L26)
If you are short of time I can do it afterwards

gitops/pac/.tekton/pln-svc-registration.yaml
```diff
@@ -0,0 +1,59 @@
...
+          - fetch-repository
+        workspaces:
+          - name: source
+            workspace: source
+        taskSpec:
+          workspaces:
+            - name: source
+          steps:
+            - name: cluster-setup
+              image: quay.io/bnr/pipelines-setup:latest
```
fgiloux, line 46: no action: we can create a GitHub action and use ghcr as for the other images. Can be done in another PR.

fgiloux (approved):
A few minor changes that can also be handled later on

PR #96, opened 2022-06-04
Refactor cluster-setup install script

Roming22:
This an overview of the code changes that would need to be done if we agree on the gitops model described by PR #86.


images/cluster-setup/Dockerfile
```diff
@@ -7,7 +7,7 @@ COPY ./install.sh /usr/local/bin/install.sh
 RUN KUBE_VERSION=v1.24.0 && \
     curl -L -o /usr/local/bin/kubectl "https://dl.k8s.io/release/$KUBE_VERSION/bin/linux/amd64/kubectl" && \
     chmod 755 /usr/local/bin/kubectl
+RUN microdnf install findutils git
 USER 65532:65532
 VOLUME /workspace
 WORKDIR /workspace
-ENTRYPOINT ["/usr/local/bin/install.sh"]
```
fgiloux, line 13: Why are you removing the entrypoint? I would understand if it was replaced by a command but it is not the case

images/cluster-setup/install.sh
```diff
@@ -20,122 +20,127 @@ set -o pipefail
 
 usage() {
 
-    printf "Usage: DATA_DIR="/workspace" ./register.sh \n\n"
-
-    # Parameters
-    printf "The following parameters need to be passed to the script: \n"
-    printf "DATA_DIR: the location of the cluster files \n"
+  printf "
+Usage:
```
fgiloux, line 24: As explained in the other PR environment variables are a more convenient approach.

images/cluster-setup/install.sh
```diff
@@ -20,122 +20,127 @@ set -o pipefail
...
-            fi
-        done
-    done
+init() {
+  cluster_config_dirs=($(
+    find "${WORKSPACE_DIR}/environment/compute" \
+      -name kustomization.yaml \
+      -exec dirname {} \;
+  ))
+}
```
fgiloux, line 77: As explained in the other PR the existing approach is more flexible.

fgiloux (commented):
I am not positive on this

images/cluster-setup/Dockerfile
Roming22: This was necessary so I could make the image work with GitLab CI.  It might be because I'm unfamiliar with it, so I wouldn't look to hard at that specific change.

images/cluster-setup/install.sh
Roming22: I did not remove support for env variable. I just added support for arguments on top.

images/cluster-setup/install.sh
Roming22: I thought that we had an understanding, but it looks like I was wrong. Let's have another talk about it to make sure I understand where we stand on this topic.

images/cluster-setup/Dockerfile
Roming22: Fixed

images/cluster-setup/install.sh
Roming22: Reverted to old behavior.

images/cluster-setup/install.sh
```diff
@@ -45,92 +83,74 @@ get_clusters() {
     contexts=()
     kubeconfigs=()
     printf "Extracting files under the kubeconfig dir and reading the content in each file \n"
-    files=("$(ls "$DATA_DIR/gitops/sre/credentials/kubeconfig/compute")")
+    files=("$(ls "$WORKSPACE_DIR/credentials/kubeconfig/compute")")
     for kubeconfig in "${files[@]}"; do
-        printf "%s\n" "$kubeconfig"
-        subs=("$(KUBECONFIG=${DATA_DIR}/gitops/sre/credentials/kubeconfig/compute/${kubeconfig} kubectl config view -o jsonpath='{range .contexts[*]}{.name}{","}{.context.cluster}{"\n"}{end}')")
+        printf "  - %s\n" "$kubeconfig"
+        subs=("$(KUBECONFIG=${WORKSPACE_DIR}/credentials/kubeconfig/compute/${kubeconfig} kubectl config view -o jsonpath='{range .contexts[*]}{.name}{","}{.context.cluster}{"\n"}{end}')")
```
fgiloux, line 89: I am unsure why you removed gitops/sre:
This would break the PipelineRun:
https://github.com/openshift-pipelines/pipelines-service/blob/0abf79d5a7e76f6d8a3ed168df26dc395088fed1/gitops/pac/.tekton/pln-svc-registration.yaml#L47

images/cluster-setup/install.sh
```diff
@@ -45,92 +83,74 @@ get_clusters() {
...
                 kubeconfigs+=( "${kubeconfig}" )
-                printf "%s --- %s --- %s \n" "$cluster" "$context" "$kubeconfig"
+                printf "    - %s --- %s --- %s \n" "$cluster" "$context" "$kubeconfig"
             fi
         done
     done
+    printf "\n"
+}
 
+init_cluster_kubectl() {
```
fgiloux, line 104: could be renamed:
switch_cluster()

rational:
- it is not an initialisation
- it makes clear what the intent is

images/cluster-setup/install.sh
Roming22: The content of `WORKSPACE_DIR` should be a copy of `gitops/sre/template`, and therefore would not contain `gitops/sre`. You can have a look at my proof-of-concept repository if you want.

images/cluster-setup/install.sh
Roming22: Fixed.

images/cluster-setup/install.sh
fgiloux: let's not mix things. You can move that to the other PR then.

PR #107, opened 2022-06-14
Fix issue with KIND_EXPERIMENTAL_PROVIDER not being properly set

fgiloux:
For using kind with podman KIND_EXPERIMENTAL_PROVIDER needs to be set. It was not consistently done in the script and relying on the environment variable being preset.
Besides that PODMAN_CMD was redundant as CONTAINER_ENGINE is already set by the .utils script

### DIFF
--- a/local/kind/setup.sh
+++ b/local/kind/setup.sh
@@ -25,18 +25,21 @@ prechecks () {
     fi
 
     if [[ "${ALLOW_ROOTLESS}" == "true" ]]; then
-        echo "Rootless mode enabled"
+        printf "Rootless mode enabled\n"
     fi
 
     if [[ "${CONTAINER_ENGINE}" != "docker" && "${ALLOW_ROOTLESS}" != "true" ]]; then
-        KIND_CMD="sudo kind"
-        PODMAN_CMD="sudo podman"
+        KIND_CMD="sudo KIND_EXPERIMENTAL_PROVIDER=podman kind"
     else
-        KIND_CMD="kind"
-	if [[ "${CONTAINER_ENGINE}" == "docker" ]]; then
-	    PODMAN_CMD="docker"
+	if [[ "${CONTAINER_ENGINE}" != "docker" ]]; then
+            KIND_CMD="KIND_EXPERIMENTAL_PROVIDER=podman kind"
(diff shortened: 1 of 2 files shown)


local/kind/setup.sh
```diff
@@ -95,11 +97,11 @@ for cluster in "${CLUSTERS[@]}"; do
     if [[ -z "${clusterExists}" ]]; then
         echo "Creating kind cluster ${cluster}"
         cp "${cluster}.config" "${TMP_DIR}/${cluster}.config"
-        KIND_EXPERIMENTAL_PROVIDER=${KIND_EXPERIMENTAL_PROVIDER:-} ${KIND_CMD} create cluster \
+        ${KIND_CMD} create cluster \
             --config "${TMP_DIR}/${cluster}.config" \
             --kubeconfig "${TMP_DIR}/${cluster}.kubeconfig"
 
-        if [[ ${KIND_CMD} == "sudo kind" ]]; then
+        if [[ ${KIND_CMD} == "sudo KIND_EXPERIMENTAL_PROVIDER=podman kind" ]]; then
```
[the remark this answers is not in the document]
fgiloux: no it is not the same if -E is not used: man sudo =>

> -E, --preserve-env
                 Indicates to the security policy that the user wishes to preserve their existing environment variables.  The security policy may return an error if the user does not have permission to
                 preserve the environment.

Roming22, line 104: Would `[[ "${CONTAINER_ENGINE}" != "docker" ]]` or `[[ "${CONTAINER_ENGINE}" == "podman" ]]` achieve the same result? It might be easier to read.

bnallapeta (approved):
Works locally for me with the new experimental flag change. Also, Avinal has created another PR to document these common issues under troubleshooting.

local/kind/setup.sh
fgiloux: No it is not. Key here is that when sudo is used read rights need to be granted to the user.

Roming22 (approved):
Please resolve conflicts for the merge.

fgiloux:
> Please resolve conflicts for the merge.

Done.

PR #110, opened 2022-06-14
Clean up kcp_registrar

Roming22:
The logic is mostly unchanged, except:
    1. `register.sh` is copied last to speed up image creation time on the
      local dev env;
    2. the kubeconfig are copied to a tmp folder to workaround ro permissions
      in the container;
    3. all clusters are always processed.

Item #3 is possible because the cluster registration is idempotent as
of kcp release-0.5.
Item #3 is desirable because if anything bad happens while setting up
the cluster, the recovery would require a manual action to delete the
syncer on KCP.

### DIFF
--- a/images/kcp-registrar/Dockerfile
+++ b/images/kcp-registrar/Dockerfile
@@ -1,4 +1,5 @@
 # Build the binary
+# cf https://github.com/kcp-dev/kcp/issues/1092
 FROM golang:1.17 AS builder
 
 ARG KCP_BRANCH
@@ -33,14 +34,14 @@ RUN mkdir /workspace && chmod 777 /workspace && chown 65532:65532 /workspace
 RUN mkdir $HOME && chmod 777 $HOME
 COPY --from=builder workspace/kcp/bin/kubectl-kcp /usr/local/bin/kubectl-kcp
 RUN chmod 755 /usr/local/bin/kubectl-kcp
-COPY ./register.sh /usr/local/bin/register.sh
-RUN chmod 755 /usr/local/bin/register.sh
 RUN JQ_VERSION=1.6 && \
-    curl -sSL -o /usr/local/bin/jq https://github.com/stedolan/jq/releases/download/jq-$JQ_VERSION/jq-linux64 && \
+    curl --fail -sSL -o /usr/local/bin/jq https://github.com/stedolan/jq/releases/download/jq-$JQ_VERSION/jq-linux64 && \
     chmod 755 /usr/local/bin/jq
 RUN KUBE_VERSION=v1.24.0 && \
-    curl -L -o /usr/local/bin/kubectl "https://dl.k8s.io/release/$KUBE_VERSION/bin/linux/amd64/kubectl" && \
+    curl --fail -L -o /usr/local/bin/kubectl "https://dl.k8s.io/release/$KUBE_VERSION/bin/linux/amd64/kubectl" && \
     chmod 755 /usr/local/bin/kubectl
+COPY ./register.sh /usr/local/bin/register.sh
+RUN chmod 755 /usr/local/bin/register.sh
 USER 65532:65532
 VOLUME /workspace
 WORKDIR /workspace
--- a/images/kcp-registrar/register.sh
+++ b/images/kcp-registrar/register.sh
@@ -24,50 +24,50 @@ usage() {
 
     # Parameters
     printf "The following parameters need to be passed to the script:\n"
-    printf "KCP_ORG: the organistation for which the workload clusters need to be registered, i.e.: root:pipelines-service\n"
+    printf "KCP_ORG: the organization for which the workload clusters need to be registered, i.e.: root:pipelines-service\n"
     printf "KCP_WORKSPACE: the name of the workspace where the workload clusters get registered (created if it does not exist), i.e: compute\n"
     printf "DATA_DIR: the location of the cluster files\n"
     printf "          a single file with extension kubeconfig is expected in the subdirectory: credentials/kubeconfig/kcp\n"
     printf "          kubeconfig files for compute clusters are expected in the subdirectory: credentials/kubeconfig/compute \n\n"
 }
 
-prechecks () {
+exit_error() {
+    printf "[ERROR] %s" "$@" >&2
+    usage
+    exit 1
+}
+
+prechecks() {
     KCP_ORG=${KCP_ORG:-}
     if [[ -z "${KCP_ORG}" ]]; then
-        printf "KCP_ORG not set\n\n"
-        usage
-        exit 1
+        exit_error "KCP_ORG not set\n\n"
     fi
 
-    KCP_ORG=${KCP_WORKSPACE:-}
+    KCP_WORKSPACE=${KCP_WORKSPACE:-}
     if [[ -z "${KCP_WORKSPACE}" ]]; then
-        printf "KCP_WORKSPACE not set\n\n"
-        usage
-        exit 1
+        exit_error "KCP_WORKSPACE not set\n\n"
     fi
-    
+
     DATA_DIR=${DATA_DIR:-}
     if [[ -z "${DATA_DIR}" ]]; then
-        printf "DATA_DIR not set\n\n"
-        usage
-	exit 1
+        exit_error "DATA_DIR not set\n\n"
     fi
 }
 
 # populate kcp_kcfg with the location of the kubeconfig for connecting to kcp
 kcp_kubeconfig() {
-    if files=($(ls $DATA_DIR/credentials/kubeconfig/kcp/*.kubeconfig 2>/dev/null)); then
+    kubeconfig_dir="$DATA_DIR/credentials/kubeconfig"
+    if files=($(ls "$kubeconfig_dir/kcp/"*.kubeconfig 2>/dev/null)); then
         if [ ${#files[@]} -ne 1 ]; then
-            printf "A single kubeconfig file is expected at %s\n" "$DATA_DIR/credentials/kubeconfig/kcp"
-            usage
-            exit 1
+            exit_error "A single kubeconfig file is expected at $kubeconfig_dir/kcp\n\n"
         fi
-        kcp_kcfg="${files[0]}"
     else
-        printf "A single kubeconfig file is expected at %s\n" "$DATA_DIR/credentials/kubeconfig/kcp"
-        usage
-        exit 1
+        exit_error "A single kubeconfig file is expected at $kubeconfig_dir/kcp\n\n"
     fi
+    tmp_dir=$(mktemp -d)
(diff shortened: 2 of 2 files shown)


images/kcp-registrar/register.sh
```diff
@@ -20,54 +20,58 @@ set -o pipefail
 
 usage() {
 
-    printf "Usage: KCP_ORG="root:pipelines-service" KCP_WORKSPACE="compute" DATA_DIR="/workspace" ./register.sh\n\n"
+    printf "Usage: KCP_ORG="root:pipeline-svc" KCP_WORKSPACE="compute" DATA_DIR="/workspace" ./register.sh\n\n"
```
fgiloux, line 23: sorry but I don't see the rational for changing that. Readability is in my opinion more important than saving the typing of 4 characters, especially when tab completion allows not to type them.

images/kcp-registrar/register.sh
```diff
@@ -20,54 +20,58 @@ set -o pipefail
...
     printf "          kubeconfig files for compute clusters are expected in the subdirectory: credentials/kubeconfig/compute \n\n"
 }
 
-prechecks () {
+exit_error() {
+    printf "[ERROR] %s\n\n" "$@" >&2
+    usage
+    exit 1
+}
+
```
fgiloux, line 39: I don't think this modification is worth the time for review and validation. The gain is negligible at best and it enforces a formatting that may not always be welcome.

images/kcp-registrar/register.sh
```diff
@@ -20,54 +20,58 @@ set -o pipefail
...
 
+kcp_kubectl(){
+    KUBECONFIG="$kcp_kcfg" kubectl "$@"
+}
+
 # populate kcp_kcfg with the location of the kubeconfig for connecting to kcp
 kcp_kubeconfig() {
-    if files=($(ls $DATA_DIR/credentials/kubeconfig/kcp/*.kubeconfig 2>/dev/null)); then
+    credentials_dir="$DATA_DIR/credentials"
+    if files=($(ls "$credentials_dir/kubeconfig/kcp/"*.kubeconfig 2>/dev/null)); then
```
fgiloux, line 64: same here.

images/kcp-registrar/register.sh
```diff
@@ -20,54 +20,58 @@ set -o pipefail
...
-        usage
-	exit 1
+        exit_error "DATA_DIR not set"
     fi
 }
 
+kcp_kubectl(){
+    KUBECONFIG="$kcp_kcfg" kubectl "$@"
+}
+
```
fgiloux, line 60: same here

images/kcp-registrar/register.sh
```diff
@@ -78,65 +82,59 @@ get_clusters() {
...
-            KUBECONFIG=${DATA_DIR}/credentials/kubeconfig/compute/${kubeconfigs[$i]} kubectl apply --context ${contexts[$i]} -f /tmp/syncer-${clusters[$i]}.yaml 
-        fi
+        cluster_name="$(echo "${clusters[$i]}" | cut -d: -f1)"
+        printf "  - %s\n" "$cluster_name"
+        syncer_manifest="/tmp/syncer-$cluster_name.yaml"
+        kcp_kubectl kcp workload sync "$cluster_name" \
+            --syncer-image ghcr.io/kcp-dev/kcp/syncer:release-0.4 \
+            --resources deployments.apps,services,ingresses.networking.k8s.io,conditions.tekton.dev,pipelines.tekton.dev,pipelineruns.tekton.dev,pipelineresources.tekton.dev,runs.tekton.dev,tasks.tekton.dev,taskruns.tekton.dev \
+            >"$syncer_manifest"
+        KUBECONFIG="$credentials_dir/kubeconfig/compute/${kubeconfigs[$i]}" kubectl apply --context ${contexts[$i]} -f "$syncer_manifest"
```
fgiloux, line 128: There are implications in re-registering a cluster that has already been registered. Frankly I am not even sure that the kcp workload sync command would work if a workload cluster with the same cluster name already exists.
We can surely validate and refine our operational model regarding upgrades and handling of out-of-band changes but I don't think that this PR is the place for it.

images/kcp-registrar/Dockerfile
```diff
@@ -30,15 +30,16 @@ RUN mkdir /workspace && chmod 777 /workspace && chown 65532:65532 /workspace
...
-    curl -L -o /usr/local/bin/kubectl "https://dl.k8s.io/release/$KUBE_VERSION/bin/linux/amd64/kubectl" && \
+    curl --fail -L -o /usr/local/bin/kubectl "https://dl.k8s.io/release/$KUBE_VERSION/bin/linux/amd64/kubectl" && \
     chmod 755 /usr/local/bin/kubectl
+COPY ./register.sh /usr/local/bin/register.sh
+RUN chmod 755 /usr/local/bin/register.sh
 USER 65532:65532
 VOLUME /workspace
 WORKDIR /workspace
-CMD ["/usr/local/bin/register.sh"]
+ENTRYPOINT /usr/local/bin/register.sh
```
fgiloux, line 45: What is the rational for removing the command and overriding the default entrypoint?
"/bin/sh -c" is convenient for any troubleshooting exercise. Surely it is possible to override that in Kubernetes or with the container run command of the container engine but in my opinion this change makes the image more difficult to use.

images/kcp-registrar/Dockerfile
```diff
@@ -30,15 +30,16 @@ RUN mkdir /workspace && chmod 777 /workspace && chown 65532:65532 /workspace
 RUN mkdir $HOME && chmod 777 $HOME
 COPY --from=builder workspace/kcp/bin/kubectl-kcp /usr/local/bin/kubectl-kcp
 RUN chmod 755 /usr/local/bin/kubectl-kcp
-COPY ./register.sh /usr/local/bin/register.sh
-RUN chmod 755 /usr/local/bin/register.sh
+RUN microdnf install findutils git
```
fgiloux, line 33: This was not needed so far. What is the rational for that?

images/kcp-registrar/Dockerfile
```diff
@@ -30,15 +30,16 @@ RUN mkdir /workspace && chmod 777 /workspace && chown 65532:65532 /workspace
...
 RUN JQ_VERSION=1.6 && \
-    curl -sSL -o /usr/local/bin/jq https://github.com/stedolan/jq/releases/download/jq-$JQ_VERSION/jq-linux64 && \
+    curl --fail -sSL -o /usr/local/bin/jq https://github.com/stedolan/jq/releases/download/jq-$JQ_VERSION/jq-linux64 && \
     chmod 755 /usr/local/bin/jq
 RUN KUBE_VERSION=v1.24.0 && \
-    curl -L -o /usr/local/bin/kubectl "https://dl.k8s.io/release/$KUBE_VERSION/bin/linux/amd64/kubectl" && \
+    curl --fail -L -o /usr/local/bin/kubectl "https://dl.k8s.io/release/$KUBE_VERSION/bin/linux/amd64/kubectl" && \
     chmod 755 /usr/local/bin/kubectl
+COPY ./register.sh /usr/local/bin/register.sh
+RUN chmod 755 /usr/local/bin/register.sh
```
fgiloux, line 41: I would expect register.sh to be more stable than jq, kubectl that get regularly released. My feeling is that the change in the ordering will have a negative impact

fgiloux (changes requested):
Hi @Roming22 please see the comments I provided to this PR. At this point in time I am struggling with seeing the proposed value.

images/kcp-registrar/register.sh
Roming22: I changed it because our current org workspace is named pipeline-svc. If you don't agree I can revert that change, it is not a big deal.

images/kcp-registrar/register.sh
Roming22: I did validate that executing a `kcp workload sync` did not generate a failure.

images/kcp-registrar/Dockerfile
Roming22: When one does local dev, and iterates on a new feature, the files in the repo are most likely to change. Having to reinstall the application is slowing down the dev cycle. I believe this change is in the best interest of the developer.

Roming22:
@bnallapeta @fgiloux This PR fixes the failure that we have on our GitLab pipelines.

You can easily test yourself by checking out our `kcp-stable` repo and running:
```
> KCP_ORG=root:rh-sso-15826143 KCP_WORKSPACE=compute-pwjs8 DATA_DIR=path/to/kcp-stable ./register.sh
Switching to organization root:rh-sso-15826143
Current workspace is "root:rh-sso-15826143".
Switching to workspace compute-pwjs8
  - Use existing workspace
Current workspace is "root:rh-sso-15826143:compute-pwjs8".
Registering clusters to kcp
  - api-pipeline-stg-zs03-p1-openshiftapps-com
namespace/kcpsync0b7864f7bcce800fcf840cf14a5af7960cd218900e6337ec690e145e unchanged
serviceaccount/kcp-syncer unchanged
clusterrole.rbac.authorization.k8s.io/kcpsync0b7864f7bcce800fcf840cf14a5af7960cd218900e6337ec690e145e unchanged
clusterrolebinding.rbac.authorization.k8s.io/kcpsync0b7864f7bcce800fcf840cf14a5af7960cd218900e6337ec690e145e unchanged
secret/kcp-syncer-config configured
deployment.apps/kcp-syncer unchanged
```

Roming22:
I've reverted the bit that Frederic is not comfortable with at the moment, because it is important to merge that PR to move our sprint forward. I'll open a separate PR with that change, and we can discuss the impact over there.

@fgiloux I believe all the important points of your review were taken care of.

images/kcp-registrar/register.sh
```diff
@@ -24,50 +24,50 @@ usage() {
...
     else
-        printf "A single kubeconfig file is expected at %s\n" "$DATA_DIR/credentials/kubeconfig/kcp"
-        usage
-        exit 1
+        exit_error "A single kubeconfig file is expected at $kubeconfig_dir/kcp\n\n"
     fi
+    tmp_dir=$(mktemp -d)
+    cp -rf "$DATA_DIR/credentials" "$tmp_dir"
+    kubeconfig_dir="$tmp_dir/credentials/kubeconfig"
+    kcp_kcfg="$(ls "$kubeconfig_dir/kcp/"*.kubeconfig)"
```
fgiloux, line 70: Nitpicking: 
You can as well copy the single kubeconfig file ${files[0]} in the if case. The above let to think that there are multiple kubeconfig files, which is not the case.
If you want to copy everything, it would be better to copy first and make the check for file uniqueness on the content of tmp_dir

fgiloux (approved):
I hope we minimize the number of such PRs that are time consuming for the sake of individual coding preferences.

PR #111, opened 2022-06-15
Add more troubleshooting scenarios

avinal:
Added troubleshooting for the following:
- Error: can only create exec sessions on running containers: container state improper
- Failed to create kind cluster - "Reached target ._Multi-User System._|detected cgroup v1"
- Looping with "error: availableReplicas is not found"
- "Kind could not be found" when running with sudo

Signed-off-by: Avinal Kumar <avinkuma@redhat.com>

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -21,12 +21,12 @@ Prerequisites:
 ./local/kind/setup.sh
 ```
 
-By default Argo CD is installed on both clusters. It is possible to deactivate its installation by setting an environment variable `NO_ARGOCD=true`
+By default, Argo CD is installed on both clusters. It is possible to deactivate its installation by setting an environment variable `NO_ARGOCD=true`
 
 ---
 **_NOTES:_**
 
-1. Podman is used per default as container engine if available on the machine. If both podman and docker are available it is possible to force the use of docker by setting an environment variable `CONTAINER_ENGINE=docker`.
+1. Podman is used as the default container engine, if available on the machine. If both podman and docker are available it is possible to force the use of docker by setting an environment variable `CONTAINER_ENGINE=docker`.
 2. Podman defaults to running as sudo (root). To run podman in rootless mode, use the environment variable `ALLOW_ROOTLESS=true`. See the [kind documentation](https://kind.sigs.k8s.io/docs/user/rootless/) for the prerequisites.
 
 ---
@@ -50,7 +50,7 @@ The script will output the location of the kubeconfig file that can be used to i
 ```
 
 ---
-**_NOTE:_**  podman is used per default as container engine if available on the machine. If both podman and docker are available it is possible to force the use of docker by setting an environment variable `CONTAINER_ENGINE=docker`.
+**_NOTE:_** Podman is used as the default container engine, if available on the machine. If both podman and docker are available it is possible to force the use of docker by setting an environment variable `CONTAINER_ENGINE=docker`.
 
 ---
 
@@ -62,7 +62,7 @@ Here is an example for running the registration image in a development environme
 podman run --env KCP_ORG='root:pipelines-service' --env KCP_WORKSPACE='compute' --env DATA_DIR='/workspace' --privileged --volume /home/myusername/plnsvc:/workspace quay.io/myuser/pipelines-kcp
 ```
 
-Make sure that iptables/firewalld are not preventing the communication (tcp initiated from the kind clusters) between the containers running on the kind network and the kcp process on the host.
+Make sure that iptables/firewalld is not preventing the communication (tcp initiated from the kind clusters) between the containers running on the kind network and the kcp process on the host.
 
 To check iptables rules:
 
@@ -82,7 +82,7 @@ A gateway can be installed to expose endpoints running on the workload clusters 
 
 ## GitOps
 
-Argo CD is by default installed on both clusters. Alternatively it can be installed afterwards by running the following:
+Argo CD is by default installed on both clusters. Alternatively, it can be installed afterward by running the following:
 
 ```bash
 KUBECONFIG=/path-to/config.kubeconfig ./local/argocd/setup.sh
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -126,3 +126,39 @@ On Fedora, this can be fixed by adding a `.conf` file to `/etc/sysctl.d/ increas
 fs.inotify.max_user_watches=2097152
 fs.inotify.max_user_instances=256
 ```
+
+### Error: can only create exec sessions on running containers: container state improper
+
+This may happen when the containers have stopped. This can be fixed by deleting the already created clusters and starting the script again.
+
+```bash
+sudo kind delete cluster us-east1 us-west1
```
ramessesii2, line 135: Shouldn't it be `kind delete clusters us-west1 us-east1 `?

ramessesii2 (changes requested):
Thanks for the PR. LGTM

DEVELOPMENT.md
avinal: Ah, yes, changing. Thanks 😅

DEVELOPMENT.md
```diff
@@ -126,3 +126,39 @@ On Fedora, this can be fixed by adding a `.conf` file to `/etc/sysctl.d/ increas
 fs.inotify.max_user_watches=2097152
 fs.inotify.max_user_instances=256
 ```
+
+### Error: can only create exec sessions on running containers: container state improper
+
+This may happen when the containers have stopped. This can be fixed by deleting the already created clusters and starting the script again.
```
fgiloux, line 132: > the containers have stopped

=> the kind container has stopped.
Note that kind containers are ephemeral and do not survive a reboot of the host.

DEVELOPMENT.md
```diff
@@ -126,3 +126,39 @@ On Fedora, this can be fixed by adding a `.conf` file to `/etc/sysctl.d/ increas
...
+
+This is happening probably due to [Docker Desktop 4.3.0+ using Cgroups v2](https://kind.sigs.k8s.io/docs/user/known-issues/#failure-to-create-cluster-with-docker-desktop-as-container-runtime).
+
+To fix this issue, please ensure that the `KIND_EXPERIMENTAL_PROVIDER=podman` is set properly. You should be able to see _enabling experimental podman provider_ while running kind commands or the scripts. If the issue remains even after this, please try to disable docker and try again.
+
+A final fix is to uninstall docker and use podman only.
+
+### Looping with "error: availableReplicas is not found"
+
+This issue is related to the Argo CD installation. It waits for the ingress controller to be ready before creating the ingress object for it. You can wait or if you don't need Argo CD you can pass `NO_ARGOCD=true` before running the script.
```
fgiloux, line 148: This is not really an issue per se. It is the expected behavior as the script checks whether the ingress controller is available. It is expected to see a few lines like that with any installation. It is becoming an issue when the ingress controller does not become ready on time or never becomes ready.
I think it would be good to state first that a few lines like the one above are expected.

DEVELOPMENT.md
```diff
@@ -126,3 +126,39 @@ On Fedora, this can be fixed by adding a `.conf` file to `/etc/sysctl.d/ increas
...
+
+The other solution is to increase `fs.inotify.max_user_instances`. This can be done either by the solution described above or by running the following command.
```
fgiloux, line 150: I would put some explanation: it has been noticed that a common reason for the ingress controller not becoming ready is scarcity of file monitors.... this is why increasing `fs.inotify.max_user_instances` may help.

fgiloux (changes requested):
Thanks @avinal for putting that together. It will surely be helpful. I have a few comments that, I hope, can easily get addressed. I am looking forward to this PR getting merged.

DEVELOPMENT.md
avinal: Sure, I updated it.

DEVELOPMENT.md
```diff
@@ -50,7 +50,7 @@ The script will output the location of the kubeconfig file that can be used to i
 ```
 
 ---
-**_NOTE:_**  podman is used per default as container engine if available on the machine. If both podman and docker are available it is possible to force the use of docker by setting an environment variable `CONTAINER_ENGINE=docker`.
+**_NOTE:_**  The podman is used per default as container engine if available on the machine. If both podman and docker are available it is possible to force the use of docker by setting an environment variable `CONTAINER_ENGINE=docker`.
```
Roming22, line 53: I don't think that the podman binary needs to be prefixed by `The`. I would revert that specific change. As a matter of fact, it is a repeat of the note line 29, so if possible I would reference that instead of duplicating the information.

Roming22 (commented):
Thanks for this PR. Just a couple of small changes and it should be good to go.

DEVELOPMENT.md
```diff
@@ -126,3 +126,39 @@ On Fedora, this can be fixed by adding a `.conf` file to `/etc/sysctl.d/ increas
...
+
+If it is too long and the ingress controller is still not ready, that is probably due to the scarcity of file monitors on the system. In that case, increasing `fs`.inotify.max_user_instances` may help. This can be done either by the solution described [above](#ingress-router-pods-crashloop---too-many-open-files) or by running the following command.
+
+```bash
+sudo sysctl -w fs.inotify.max_user_instances=256
+```
+
+### "Kind could not be found" when running with sudo
+
+This issue may occur if the kind was installed using `go install`. This way the kind executable is placed in `GOBIN` (in the home directory) and thus not available to root.
```
fgiloux, line 158: please remove "the" before kind in "if the kind was installed". "kind" is the name of the app.

DEVELOPMENT.md
fgiloux: +1 on not having "The". I am not too concerned about the duplication. It is sometimes better not to break the flow.

DEVELOPMENT.md
```diff
@@ -21,12 +21,12 @@ Prerequisites:
...
 ```
 
-By default Argo CD is installed on both clusters. It is possible to deactivate its installation by setting an environment variable `NO_ARGOCD=true`
+By default, Argo CD is installed on both clusters. It is possible to deactivate its installation by setting an environment variable `NO_ARGOCD=true`
 
 ---
 **_NOTES:_**
 
-1. Podman is used per default as container engine if available on the machine. If both podman and docker are available it is possible to force the use of docker by setting an environment variable `CONTAINER_ENGINE=docker`.
+1. Podman is used per default as the container engine is available on the machine. If both podman and docker are available it is possible to force the use of docker by setting an environment variable `CONTAINER_ENGINE=docker`.
```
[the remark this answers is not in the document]
avinal: Oh, I misinterpreted the meaning of the complete sentence. Will change that.

DEVELOPMENT.md
avinal: Ok got it

DEVELOPMENT.md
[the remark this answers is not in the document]
avinal: I tried to rephrase the sentence for clarity.

fgiloux (approved):

avinal:
Thanks 🚀

PR #116, opened 2022-06-17
Fix a couple of issues with the registration

fgiloux:
- KCP_TAG variable has been added to the script but not explicitely declared. It is useful when running the script locally and for documentation purpose
- KCP_WORKSPACE was missing in the PipelineRun

### DIFF
--- a/gitops/sre/.tekton/kcp-registration.yaml
+++ b/gitops/sre/.tekton/kcp-registration.yaml
@@ -50,6 +50,8 @@ spec:
                   value: $(workspaces.source.path)
                 - name: KCP_ORG
                   value: "pipelines-service"
+                - name: KCP_WORKSPACE
+                  value: "compute"
   workspaces:
     - name: source
       volumeClaimTemplate:
--- a/images/kcp-registrar/register.sh
+++ b/images/kcp-registrar/register.sh
@@ -20,12 +20,13 @@ set -o pipefail
 
 usage() {
 
-    printf "Usage: KCP_ORG="root:pipelines-service" KCP_WORKSPACE="compute" DATA_DIR="/workspace" ./register.sh\n\n"
+    printf "Usage: KCP_ORG="root:pipelines-service" KCP_WORKSPACE="compute" KCP_SYNC_TAG="release-0.5" DATA_DIR="/workspace" ./register.sh\n\n"
 
     # Parameters
     printf "The following parameters need to be passed to the script:\n"
     printf "KCP_ORG: the organization for which the workload clusters need to be registered, i.e.: root:pipelines-service\n"
     printf "KCP_WORKSPACE: the name of the workspace where the workload clusters get registered (created if it does not exist), i.e: compute\n"
+    printf "KCP_SYNC_TAG: the tag of the kcp syncer image to use (preset in the container image at build time and leveraged by the PipelineRun)\n"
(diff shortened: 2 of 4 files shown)


gitops/sre/.tekton/kcp-registration.yaml
```diff
@@ -50,6 +50,8 @@ spec:
                   value: $(workspaces.source.path)
                 - name: KCP_ORG
                   value: "pipelines-service"
+                - name: KCP_WORKSPACE
```
bnallapeta, line 53: Don't we have to add KCP_TAG here too?

images/kcp-registrar/register.sh
```diff
@@ -26,6 +26,7 @@ usage() {
     printf "The following parameters need to be passed to the script:\n"
     printf "KCP_ORG: the organization for which the workload clusters need to be registered, i.e.: root:pipelines-service\n"
     printf "KCP_WORKSPACE: the name of the workspace where the workload clusters get registered (created if it does not exist), i.e: compute\n"
+    printf "KCP_TAG: the tag of the kcp syncer image to use (preset in the image when using Pipelines as Code)\n"
```
bnallapeta, line 29: We have to document KCP_TAG at this place as well.
https://github.com/openshift-pipelines/pipelines-service/blob/main/docs/kcp-registration.md

bnallapeta, line 29: wdym its preset when using PAC? Do you mean that its set as an env var [here](https://github.com/openshift-pipelines/pipelines-service/blob/main/gitops/sre/.tekton/kcp-registration.yaml#L48)?

bnallapeta (commented):
Minor changes.

bnallapeta:
@fgiloux On reflection, how about renaming KCP_TAG to a more appropriate name - SYNCER_TAG?

gitops/sre/.tekton/kcp-registration.yaml
fgiloux: No we don't have because of [this](https://github.com/openshift-pipelines/pipelines-service/blob/main/images/kcp-registrar/Dockerfile#L5-L6)

images/kcp-registrar/register.sh
fgiloux: No, it is set directly in the image (cf above). I added this comment as I thought it was not transparent.

images/kcp-registrar/register.sh
```diff
@@ -127,6 +133,8 @@ register() {
                 --syncer-image ghcr.io/kcp-dev/kcp/syncer:$KCP_TAG \
                 --resources deployments.apps,services,ingresses.networking.k8s.io,conditions.tekton.dev,pipelines.tekton.dev,pipelineruns.tekton.dev,pipelineresources.tekton.dev,runs.tekton.dev,tasks.tekton.dev,taskruns.tekton.dev \
                 >"$syncer_manifest"
+	    echo ${kubeconfigs[$i]}
```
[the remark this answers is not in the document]
fgiloux: tests leftover I will remove them

gitops/sre/.tekton/kcp-registration.yaml
fgiloux: This is actually wrong. The Tag is only set in the builder image.

bnallapeta (approved):
LGTM.

Issue #105, opened 2022-06-13
Document workflow for infrastructure provisioning


fgiloux:
@Roming22 this is a follow-up of your PR #86. I feel that

-  gitops/README.md should be extended to provide instructions on
    - the structure of our repo with the availability of kustomize base directories and their location
    - the standard approach for consuming it by either forking (especially in disconnected environment) or copying the sre directory
    - the way to create overlays using local base (fork) or remote base (copy) 
    - the presence of examples under /doc/sre/examples
- . /README.md should introduce and link the gitops/README.md under "How do I start?"

PR #117, opened 2022-06-18
Add gateway deployment for Pipelines as Code

fgiloux:
This provides the manifest and a workflow for deploying a gateway.
There are however limitations and issues:
- ingress is broken with kcp 0.5
- there is no controller watching EventListeners, which means that the gateway is statically configured for Pipelines as Code but in its current state does not support Tekton Triggers
- PipelineRuns triggered by PaC are not visible in any kcp workspace

These aspects will be addressed in phase 2.


gitops/sre/.tekton/gateway-deployment.yaml
```diff
@@ -0,0 +1,61 @@
...
+        taskSpec:
+          workspaces:
+            - name: source
+          steps:
+            - name: gateway-deployment
+              image: ghcr.io/openshift-pipelines/gateway-deployment:main
+              workingDir: $(workspaces.source.path)
+              env:
+                - name: DATA_DIR
+                  value: $(workspaces.source.path)/gitops/sre
```
bnallapeta, line 50: We have set this value as  $(workspaces.source.path) for the other two images. And in the script, it is indeed referenced like this. So, could you please change the value for the cluster-setup.yaml and kcp-registration.yaml too?
files=($(ls $DATA_DIR/credentials/kubeconfig/compute))

images/gateway/install.sh
```diff
@@ -0,0 +1,111 @@
...
+	KUBECONFIG=${kcp_kcfg} kubectl kcp workspace use ${KCP_WORKSPACE}
+
+    else
+       printf "creating workspace %s\n" "${KCP_WORKSPACE}"
+       KUBECONFIG=${kcp_kcfg} kubectl kcp workspace create "${KCP_WORKSPACE}" --enter
+    fi
+}
+
+install_gateway () {
+    KUBECONFIG=${kcp_kcfg}
```
bnallapeta, line 92: Since you are using KUBECONFIG directly in line 94, this can be removed.

images/gateway/install.sh
```diff
@@ -0,0 +1,111 @@
...
+    CONFIG_DIR="${DATA_DIR}/environment/kcp/gateway"
+    KUBECONFIG=${kcp_kcfg} kubectl apply -k  "${CONFIG_DIR}"
+}
+
+main() {
+printf "Switching to organization %s\n" ${KCP_ORG}
```
bnallapeta, line 98: I think this line is here by mistake?

images/gateway/install.sh
```diff
@@ -0,0 +1,111 @@
...
+    prechecks
+    kcp_kubeconfig
+    printf "Switching to organization %s\n" ${KCP_ORG}
```
bnallapeta, line 101: Double quote KCP_ORG, KCP_WORKSPACE everywhere

images/kcp-registrar/register.sh
```diff
@@ -39,7 +39,7 @@ prechecks () {
         exit 1
     fi
 
-    KCP_ORG=${KCP_WORKSPACE:-}
+    KCP_WORKSPACE=${KCP_WORKSPACE:-}
```
bnallapeta, line 42: This file is old. Rebasing should remove this file.

images/gateway/install.sh
```diff
@@ -0,0 +1,111 @@
...
+    else
+       printf "creating workspace %s\n" "${KCP_WORKSPACE}"
+       KUBECONFIG=${kcp_kcfg} kubectl kcp workspace create "${KCP_WORKSPACE}" --enter
+    fi
+}
+
+install_gateway () {
+    KUBECONFIG=${kcp_kcfg}
+    CONFIG_DIR="${DATA_DIR}/environment/kcp/gateway"
+    KUBECONFIG=${kcp_kcfg} kubectl apply -k  "${CONFIG_DIR}"
```
Roming22, line 94: double space after `-k`.

images/gateway/install.sh
Roming22, line 101: I'd move the print statement in the function itself, but that might be a personal preference.

gitops/sre/.tekton/gateway-deployment.yaml
bnallapeta: We do not need the gitops/sre as folders 'credentials' and 'environment' are placed at the root level.
Please ignore my previous comment. And in this file, we need to remove '/gitops/sre'.

fgiloux: > We have set this value as $(workspaces.source.path) for the other two images. And in the script, it is indeed referenced like this. So, could you please change the value for the cluster-setup.yaml and kcp-registration.yaml too?

This PR is focused on the gateway. It is unfortunate that changes made by PR #86 broke other things.

> And in this file, we need to remove '/gitops/sre'.

I disagree. The location of the credentials directory is under gitops/sre/credentials, which is exactly what you get if you fork this repository. Other approaches of creating a gitops repository, which may not require a gitops/sre directory are acceptable but then the persons creating such a repository can adjust the environment variable to their custom structure.

The documentation, which should have come with PR #86  is unfortunately missing. That's the reason I created [issue #105 ](https://github.com/openshift-pipelines/pipelines-service/issues/105). It is unfortunate that it has not been worked on.

images/gateway/install.sh
fgiloux: true. Thanks.

images/gateway/install.sh
fgiloux: True. Thanks

images/gateway/install.sh
fgiloux: ok

fgiloux:
@bnallapeta @Roming22 anything preventing the merge of this PR?

gitops/sre/.tekton/gateway-deployment.yaml
bnallapeta: - Currently, both the cluster-setup and kcp-registrar PR have workingDir: $(workspaces.source.path). 
- And even in the register.sh script for kcp-registrar, we are using $DATA_DIR/credentials/kubeconfig where DATA_DIR will point to wherever credentials and environment are present.

https://github.com/openshift-pipelines/pipelines-service/blob/main/gitops/sre/.tekton/kcp-registration.yaml#L47
https://github.com/openshift-pipelines/pipelines-service/blob/main/gitops/sre/.tekton/pln-svc-registration.yaml#L47

docs/gateway.md
```diff
@@ -70,3 +70,10 @@ here the backend is the service `httpecho` in the `trigger` namespace listening
 **_NOTE:_**  This is only needed for phase 1. This component will get removed when we move to phase 2 and have the event listeners provisioned through kcp.
 
 ---
+
```
bnallapeta, line 73: We also need to update content under Configuration section.
eg: server s1 httpecho.trigger.svc.cluster.local:80 -> server el-pac el-pipelines-as-code-interceptor.openshift-pipelines.svc.cluster.local:8080

fgiloux: done

gitops/sre/.tekton/gateway-deployment.yaml
fgiloux: changed it for the sake of consistency with the other PipelineRuns

fgiloux:
@bnallapeta I should have addressed your requets

bnallapeta (approved):
LGTM. Merging the PR.

PR #128, opened 2022-06-29
Document kcp plugin version in ckcp/README

### LINKED ISSUE #124: KCP Plugin installation hint lacks version/branch
On https://github.com/openshift-pipelines/pipelines-service/tree/main/ckcp

I am seeing I need to install the `kcp plugin`, but doing so from main did not complete the `Workspace: ` step. I had to build from the `release-0.5 branch` (found this version was used on the commits)

bnallapeta:
Fixes #124

### DIFF
--- a/ckcp/README.md
+++ b/ckcp/README.md
@@ -35,6 +35,7 @@ Before you execute the script,
 4. You need to install [yq](https://mikefarah.gitbook.io/yq/#install)
 5. You need to install [kubectl cert-manager plugin](https://cert-manager.io/docs/usage/kubectl-plugin/#installation)
 6. You need to install [kubectl kcp plugin](https://github.com/kcp-dev/kcp/blob/main/docs/kubectl-kcp-plugin.md)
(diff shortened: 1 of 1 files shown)


ramessesii2 (approved):
lgtm

ckcp/README.md
```diff
@@ -35,6 +35,7 @@ Before you execute the script,
 4. You need to install [yq](https://mikefarah.gitbook.io/yq/#install)
 5. You need to install [kubectl cert-manager plugin](https://cert-manager.io/docs/usage/kubectl-plugin/#installation)
 6. You need to install [kubectl kcp plugin](https://github.com/kcp-dev/kcp/blob/main/docs/kubectl-kcp-plugin.md)
+   Note: ckcp uses the official kcp image in order to run kcp in a pod (latest released versions). The current version is release-0.5. Make sure to checkout this branch before installing the plugin.
```
fgiloux, line 38: I would not mention the current version. We would otherwise need to update this page every month and we may just miss it. We could mention as you did that it is advisable (especially now as kcp API is still evolving quickly) to have the same version for the kcp plugin and kcp core API (ckcp). We may mention that ckcp version can be found in ckcp/openshift/overlays/dev/kustomization.yaml otherwise we would need to extend the script (openshift_dev_setup.sh) to return the version when called with -v --version option

bnallapeta: @fgiloux yes, I too was conflicted about mentioning the version or not. I shall change it to this :

ckcp version can be found in ckcp/openshift/overlays/dev/kustomization.yaml.

bnallapeta: @fgiloux  Done. PTAL

PR #130, opened 2022-06-29
fix: KCP registration process may fail when switching workspace

Roming22:
The ServiceAccount used may not have access to the root workspace. Switching
workspaces is unnecessary if we're already in the target workspace.

### DIFF
--- a/images/kcp-registrar/register.sh
+++ b/images/kcp-registrar/register.sh
@@ -139,10 +139,14 @@ register() {
 
 prechecks
 kcp_kubeconfig
-printf "Switching to organization %s\n" "${KCP_ORG}"
-switch_org
-printf "Switching to workspace %s\n" "${KCP_WORKSPACE}"
-switch_ws
+if [[ "$(KUBECONFIG=${kcp_kcfg} kubectl kcp workspace current | cut -d\" -f2)" == "$KCP_ORG:$KCP_WORKSPACE" ]]; then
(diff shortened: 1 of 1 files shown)


fgiloux:
If I am understanding you correctly you are saying that the credentials you are using may not give you access to the organization workspace as you are trying to create the workspace directly under root.
Assuming that in such a case the workspace would be already existing and why `if [[ "$(KUBECONFIG=${kcp_kcfg} kubectl kcp workspace current | cut -d\" -f2)" == "$KCP_ORG:$KCP_WORKSPACE" ]]; then`may work. I think it would make sense not to require to be already in the correct workspace for the code to work for your scenario.
I did a bit of testing with kcp 0.5 and latest and unfortunately it is possible to switch to a workspace that  does not exist (and to create resources there), which is IMHO a flaw in kcp. I will check whether it is tracked somewhere.

TL;DR: I am fine with your change for now but we may need to follow up later on to make it more robust

Roming22:
@fgiloux What I'm saying is that if a user creates the workspace `root:org1:workspace1` and creates a service account in that workspace, the serviceaccount won't have the right to list the workspaces in `root:org1`, so will try to create the workspace `root:org1:workspace1`, which it does not have the right to do.

If you want the code to be a bit more robust, I could change the the if statement to:
```if ! KUBECONFIG="$kcp_kcfg" kubectl kcp workspace use "$KCP_ORG:$KCP_WORKSPACE"; then```

Just let me know.

Roming22:
@fgiloux I agree with your proposal.

PR #129, opened 2022-06-29
Fixed path issue with Github actions gateway-deployment image building

bnallapeta:
Latest gateway PR merge caused Github actions image build to fail due to a path issue. Fixed the path and renamed the folder accordingly.

https://github.com/openshift-pipelines/pipelines-service/actions/runs/2584339525

PR #131, opened 2022-06-30
Add labels to Dockerfile of cluster-setup and gateway-deployment

bnallapeta:


fgiloux:
It looks good to me. Let's first agree on #129 to make sure things are aligned.

PR #136, opened 2022-07-04
Revert TektonConfig until issue is resolved

avinal:
It seems TektonConfig is not being applied peoperly due to a permission issue. This PR reverts TektonConfig until the issue is resolved.

Signed-off-by: Avinal Kumar <avinkuma@redhat.com>

### DIFF
--- a/gitops/argocd/tektoncd/kustomization.yaml
+++ b/gitops/argocd/tektoncd/kustomization.yaml
@@ -2,4 +2,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - openshift-operator.yaml
-  - tekton-config.yaml
+  # Due to some permission issue probably, the TektonConfig is not being applied properly, commenting till the issue is resolved. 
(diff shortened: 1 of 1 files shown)


gitops/argocd/tektoncd/kustomization.yaml
```diff
@@ -2,4 +2,3 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - openshift-operator.yaml
-  - tekton-config.yaml
```
bnallapeta, line 5: Better to comment this line out and include a comment mentioning the reason for it (as compared to deleting the line). Even if another dev picks up this issue, they will have context.

avinal: Ok sure

bnallapeta (approved):
Merging this PR as a temp workaround so that ckcp can be consumed without custom pruning options. Tekton will still continue to do pruning with the default settings ootb.

PR #137, opened 2022-07-04
Sync NetworkPolicy resources with ckcp

ramessesii2:
# Why this PR?
- NetworkPolicy needs to be synced with kcp for us to be able to apply the NetworkPolicy resources in a kcp workspace.
 
### DIFF
--- a/ckcp/openshift_dev_setup.sh
+++ b/ckcp/openshift_dev_setup.sh
@@ -23,6 +23,7 @@ CR_TOSYNC=(
             runs.tekton.dev
             tasks.tekton.dev
             repositories.pipelinesascode.tekton.dev
+            networkpolicies.networking.k8s.io
           )
 
 usage() {


ckcp/openshift_dev_setup.sh
```diff
@@ -23,6 +23,7 @@ CR_TOSYNC=(
             runs.tekton.dev
             tasks.tekton.dev
             repositories.pipelinesascode.tekton.dev
+            networkpolicies.networking.k8s.io
```
Roming22, line 26: You need to add it to `kcp-registrar/register.sh` as well.

Roming22 (changes requested):

bnallapeta:
@ramessesii2 @Roming22 
- I am still not sure if we should provide NetPol to be synced by default. Is this a common enough use case for most of our customers to define their own network policies?
- If we do, do we also provide the default network policy we enforce? Or do we leave it as is?

ckcp/openshift_dev_setup.sh
ramessesii2: added comment in the end as per our discussion during the standup.

ramessesii2:
For retrospective - 
We decide to let the users have the option to add more resources according to requirement.
These changes are being made in #145.
Closing this PR.

PR #139, opened 2022-07-04
PLNSRVCE-420 : Implement Tekton results for openshift development

CathalOConnorRH:
Implement Tekton Results and Postgresql DB for openshift development

Signed-off-by: Cathal O'Connor cathal@redhat.com


ckcp/openshift_dev_setup.sh
```diff
@@ -309,8 +309,25 @@ install_openshift_pipeline() {
   APP="argocd"
 
   echo -n "  - openshift-pipeline application: "
+
+  # Generate postgres secret
+  kubectl create secret generic tekton-results-postgres --namespace="openshift-pipelines" --from-literal=POSTGRES_USER=postgres --from-literal=POSTGRES_PASSWORD=$(openssl rand -base64 20) --dry-run=client -o yaml | kubectl apply -f - >/dev/null 2>&1
```
ramessesii2, line 314: This script stops at this point -
```bash
+ echo -n '  - openshift-pipeline application: '
  - openshift-pipeline application: + kubectl apply -f -
++ openssl rand -base64 20
+ kubectl create secret generic tekton-results-postgres --namespace=openshift-pipelines --from-literal=POSTGRES_USER=postgres --from-literal=POSTGRES_PASSWORD=x9+6KZHuEsmghDHq0Z5EPYczrbs= --dry-run=client -o yaml
```
To me, the possible cause seems to be the fact that without `Openshift-Pipelines` operator, there's no namespace `--namespace=openshift-pipelines` which would be created later with `tektoncd` argo-app [here](https://github.com/openshift-pipelines/pipelines-service/pull/139/files#diff-0ccaa908000d6cb2cc1643e32a307cf89a8cbf5bd36ceab4557ef50817def2e1R329)
Would it be a good idea to create a different function to configure tekton-results and handle this chronology?

gitops/argocd/argo-apps/tekton-results.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
```
ramessesii2, line 2: For the customers to be able to use Tekton-results, we'll need to update `/images/cluster-setup/install.sh` as well.

ramessesii2 (commented):
Overall looks good to me.

ckcp/openshift_dev_setup.sh
CathalOConnorRH: Updated to install from a separate function

gitops/argocd/argo-apps/tekton-results.yaml
CathalOConnorRH: As discussed, this will be completed with PLNSRVCE-422 but it's dependent on PLNSRVCE-421 being completed first.

bnallapeta:
@CathalOConnorRH 
1. The directory structure we have setup for using ArgoCD follows "app of apps" pattern. Under gitops/argocd, we only maintain argocd.yaml as the parent or root app and individual components go under argocd-apps (which is called from argocd.yaml).

So, tekton-results.yaml also should go under gitops/argocd/argo-apps. The manifests (tekton-results folder) however are good to be under gitops/argocd. Please make this change and update any path calling the script accordingly.

2. Some Readme on Tekton Results would be nice. Minimally, you could mention what results is for and the end result we could see out of this. Additionally, if there is any setup they would need to do or some configuration they can update etc.

Roming22:
@CathalOConnorRH My understanding is that this PR includes the manifests to deploy the database on the target cluster. My understanding is that this is not the "production" model, which will use a managed service as the DB backend.

If that's correct, you'll need to split the manifest into 2:
- Under _ckcp: Create a folder with the manifests to deploy the DB.
- Under _gitops/argocd/tekton-results_: manifests to deploy the tekton-result api and configure the connection to the DB.

gitops/argocd/tekton-results/statefulset.yaml
```diff
@@ -0,0 +1,58 @@
...
+        app.kubernetes.io/part-of: tekton-results
+        app.kubernetes.io/version: v0.3.0
+    spec:
+      containers:
+        - envFrom:
+            - configMapRef:
+                name: tekton-results-postgres
+            - secretRef:
+                name: tekton-results-postgres
+          image: postgres@sha256:6647385dd9ae11aa2216bf55c54d126b0a85637b3cf4039ef24e3234113588e3
```
Roming22, line 30: Could we use a more user friendly tag, such as a version number?

gitops/argocd/tekton-results/deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+          name: tekton-results-api
+          volumeMounts:
+            - mountPath: /etc/tls
+              name: tls
+              readOnly: true
+      serviceAccountName: tekton-results-api
+      volumes:
+        - name: tls
+          secret:
+            secretName: tekton-results-tls
```
adambkaplan, line 61: Noticing that our GitOps approach doesn't make syncing the TLS secret easy. However, we do have access to Cert-manager as an operator, which makes it simple to create declarative, self-signed certs. We'd configure this as follows:

- Add a Subscription to openshift-cert-manager-operator - probably under `gitops/argocd/cert-manager` (with a corresponding Application in `argo-apps`)
- Create a Self-signed Issuer  https://cert-manager.io/docs/configuration/selfsigned/
- Create a Certificate using the self-signed issuer: https://cert-manager.io/docs/usage/certificate/

gitops/argocd/tekton-results/statefulset.yaml
avinal: This image name is directly taken from the last stable release version on Tekton Results. It would be preferable to keep that since there have been some problems with incompatible Postgres images.

gitops/argocd/tekton-results/deployment.yaml
bnallapeta: @adambkaplan We are already using cert-manager in ckcp [here](https://github.com/openshift-pipelines/pipelines-service/tree/main/ckcp/openshift-operators/cert-manager). If we need to extend this for results as well, we could create a folder called operators under the root directory and place cert-manager, gitops operators there. In the future, we can extend that to other operators as well.

fgiloux: @bnallapeta if I am understanding correctly cert-manager becomes part of the platform if it is used for Tekton results. As such it should be deployed through gitops.
@adambkaplan self-signed certificates are not really suitable for production use if the endpoint is accessed by a user or an external system. Is the idea that the endpoint would be accessed through the gateway / kcp loadbalancer configured with reencrypt, which would then have a certificate signed by a trusted authority and trust the CA used by cert-manager?

CathalOConnorRH: @fgiloux The certs are used by both the grpc and rest service that will be part of tekton-results. This is expected to be accessible by the user behind proper routes with reencrypt enabled. 

is cert-manager part of the deployment on all clusters or only as part of ckcp installs ?

adambkaplan: > @adambkaplan self-signed certificates are not really suitable for production use if the endpoint is accessed by a user or an external system. Is the idea that the endpoint would be accessed through the gateway / kcp loadbalancer configured with reencrypt, which would then have a certificate signed by a trusted authority and trust the CA used by cert-manager?

The gRPC server will likely need to be accessed through the gateway or equivalent REST service, in which case yes we would have a cert signed with a trusted CA. @sayan-biswas is researching this.

bnallapeta: @CathalOConnorRH Currently, it is part of just ckcp. But if we need to make it part of Tekton Results, then we could do [this](https://github.com/openshift-pipelines/pipelines-service/pull/139#discussion_r919682468) and deploy it via GitOps as @fgiloux has mentioned.

CathalOConnorRH: @bnallapeta I've opened a PR to move cert-manager install [here](https://github.com/openshift-pipelines/pipelines-service/pull/163)

Marking this as draft as there are external requirements and further changes needed to this PR before tekton-results can be merged.

CathalOConnorRH:
Closing this in favour of https://github.com/openshift-pipelines/pipeline-service/pull/227

PR #142, opened 2022-07-05
fix: Wait for resources when install pac

Roming22:

### DIFF
--- a/gitops/pac/setup.sh
+++ b/gitops/pac/setup.sh
@@ -73,9 +73,17 @@ mk_tmpdir () {
 }
 
 kustomize () {
-
     GIT_TOKEN=$(echo "${GIT_TOKEN}" | base64)
     WEBHOOK_SECRET=$(echo "${WEBHOOK_SECRET}" | base64)
+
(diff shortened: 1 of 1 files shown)


gitops/pac/setup.sh
```diff
@@ -73,10 +73,18 @@ mk_tmpdir () {
 }
 
 kustomize () {
-
     GIT_TOKEN=$(echo "${GIT_TOKEN}" | base64)
     WEBHOOK_SECRET=$(echo "${WEBHOOK_SECRET}" | base64)
 
+    # This resource may take some time to spawn, which will cause Kustomize to fail when
```
bnallapeta, line 79: As far as I know, we are not calling this script (gitops/pac/setup.sh) in any other script. So, if we are running this manually, it is expected that you have PaC installed prior to it. And if that's the case, I don't see why we would encounter issues like - "pipelinesascode.tekton.dev" not found in "kubectl api-resources"

bnallapeta:
@Roming22 These changes are also present in #143. Do you mind if we close this one? Anyways, this change is also related to the changes in 143. So let's work on a single PR.

PR #143, opened 2022-07-05
Use access_setup to setup KCP and the compute

Roming22:
Update openshift_dev_setup to leverage the scripts used to deploy the
service on production clusters.

### DIFF
--- a/ckcp/openshift_dev_setup.sh
+++ b/ckcp/openshift_dev_setup.sh
@@ -2,7 +2,6 @@
 
 #quit if exit status of any cmd is a non-zero value
 set -euo pipefail
-# set -x
 
 SCRIPT_DIR="$(
   cd "$(dirname "$0")" >/dev/null
@@ -13,16 +12,14 @@ GITOPS_DIR="$(dirname "$SCRIPT_DIR")/gitops"
 CKCP_DIR="$(dirname "$SCRIPT_DIR")/ckcp"
 WORK_DIR="$SCRIPT_DIR/work"
 KUBECONFIG=${KUBECONFIG:-$HOME/.kube/config}
-KUBECONFIG_KCP="$WORK_DIR/kubeconfig/admin.kubeconfig"
-KUBECONFIG_MERGED="merged-config.kubeconfig:$KUBECONFIG:$KUBECONFIG_KCP"
-CR_TOSYNC=(
+KUBECONFIG_KCP="$WORK_DIR/credentials/kubeconfig/kcp/admin.kubeconfig.base"
+CR_TO_SYNC=(
             conditions.tekton.dev
             pipelines.tekton.dev
             pipelineruns.tekton.dev
             pipelineresources.tekton.dev
             runs.tekton.dev
             tasks.tekton.dev
-            repositories.pipelinesascode.tekton.dev
           )
 
 usage() {
@@ -33,46 +30,20 @@ Usage:
 Setup the pipeline service on a cluster running on KCP.
 
 Optional arguments:
-    -a, --app APP
-        Install APP in the cluster. APP must be pipelines.
-        The flag can be repeated to install multiple apps.
-    --all
-        Install all applications.
     -d, --debug
         Activate tracing/debug mode.
     -h, --help
         Display this message.
 
 Example:
-    ${0##*/} --all
+    ${0##*/}
 " >&2
 
 }
 
 parse_args() {
-  local default_list="openshift-gitops ckcp"
-  local pipeline_list="openshift-pipeline"
-  APP_LIST="$default_list"
-  cluster_type="openshift"
-
   while [[ $# -gt 0 ]]; do
     case $1 in
-    -a | --app)
-      shift
-      case $1 in
-      pipelines)
-        APP_LIST="$APP_LIST $pipeline_list"
-        ;;
-      *)
-        echo "[ERROR] Unsupported app: $1" >&2
-        usage
-        exit 1
-        ;;
-      esac
-      ;;
-    --all)
-      APP_LIST="$default_list $pipeline_list"
-      ;;
     -d | --debug)
       set -x
       ;;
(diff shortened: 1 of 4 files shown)


Roming22:
@xinredhat I removed `repositories.pipelinesascode.tekton.dev` from the synced resources because we don't sync it on a regular cluster, so I'm unsure why we would need to sync it for dev purposes.

Let me know if that's a problem and we can discuss the best solution.

xinredhat:
> ynced

@Roming22 it's ok for me. we don't need it for now

ckcp/openshift_dev_setup.sh
```diff
@@ -13,16 +12,14 @@ GITOPS_DIR="$(dirname "$SCRIPT_DIR")/gitops"
 CKCP_DIR="$(dirname "$SCRIPT_DIR")/ckcp"
 WORK_DIR="$SCRIPT_DIR/work"
 KUBECONFIG=${KUBECONFIG:-$HOME/.kube/config}
-KUBECONFIG_KCP="$WORK_DIR/kubeconfig/admin.kubeconfig"
-KUBECONFIG_MERGED="merged-config.kubeconfig:$KUBECONFIG:$KUBECONFIG_KCP"
+KUBECONFIG_KCP="$WORK_DIR/credentials/kubeconfig/kcp/admin.kubeconfig.base"
 CR_TOSYNC=(
```
bnallapeta, line 16: I know you didn't work on this, but could you please change this to CR_TO_SYNC?

ckcp/openshift_dev_setup.sh
```diff
@@ -265,53 +248,37 @@ patches:
...
+    --kcp-workspace "$kcp_workspace" \
+    --work-dir "$WORK_DIR"
+  KUBECONFIG_KCP="$WORK_DIR/credentials/kubeconfig/kcp/ckcp-ckcp.default.compute.kubeconfig"
 }
 
-install_openshift_pipeline() {
-  APP="argocd"
+install_compute(){
+  echo "  - Setup compute access:"
+  "$SCRIPT_DIR/../images/access-setup/content/bin/setup_compute.sh" \
```
bnallapeta, line 267: this line which calls the setup_compute.sh script has issues.
[this](https://github.com/openshift-pipelines/pipelines-service/blob/main/images/access-setup/content/bin/setup_compute.sh#L112) line gets the compute including :6443 at the end which then is used at [this](https://github.com/openshift-pipelines/pipelines-service/blob/main/images/access-setup/content/bin/setup_compute.sh#L120) line that throws the below error:

```
    - Generate kubeconfig: W0706 22:40:30.310494  285476 loader.go:221] Config not found: /home/bnr/workspace/pipelines-service/ckcp/work/credentials/kubeconfig/compute/api-ci-ln-qwgzqbb-72292-origin-ci-int-gce-dev-rhcloud-com, 6443.kubeconfig
```

Fix : remove everything after colon [here](https://github.com/openshift-pipelines/pipelines-service/blob/main/images/access-setup/content/bin/setup_compute.sh#L112) 

`"$(yq '.contexts[] | select(.name == "'"$current_context"'") | .context.cluster' "/home/bnr/.kube/config")" | sed 's/:.*//'`

ckcp/openshift_dev_setup.sh
```diff
@@ -265,53 +248,37 @@ patches:
...
+    --work-dir "$WORK_DIR"
 
-  echo -n "  - openshift-pipeline application: "
-  kubectl apply -f "$GITOPS_DIR/$APP/$APP.yaml" --wait >/dev/null 2>&1
-  argocd app wait pipelines-service tektoncd --timeout=90 >/dev/null 2>&1
-  echo "OK"
+  echo "  - Deploy compute:"
+  "$SCRIPT_DIR/../images/cluster-setup/install.sh" --workspace-dir "$WORK_DIR"
+
+  echo "  - Setup Pipelines as Code:"
```
bnallapeta, line 274: Perhaps, 'Installing Pipelines as Code' is more apt here.

ckcp/openshift_dev_setup.sh
```diff
@@ -265,53 +248,37 @@ patches:
...
+  GITOPS_REPO="https://example.git.com/my/repo" GIT_TOKEN="placeholder_token" \
```
bnallapeta, line 275: Could you please put a comment before this line with the reason to call the script with the dummy values?  something like

#Passing dummy values to the parameters of the pac/setup.sh script because we only want to install the runner side of resources.

ckcp/openshift_dev_setup.sh
```diff
@@ -33,11 +30,6 @@ Usage:
 Setup the pipeline service on a cluster running on KCP.
 
 Optional arguments:
-    -a, --app APP
-        Install APP in the cluster. APP must be pipelines.
-        The flag can be repeated to install multiple apps.
-    --all
-        Install all applications.
```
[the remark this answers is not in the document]
Roming22: Done

ckcp/openshift_dev_setup.sh
Roming22: Done

ckcp/openshift_dev_setup.sh
Roming22: Done

ckcp/openshift_dev_setup.sh
Roming22: Done

Roming22:
> @Roming22 So technically, having PaC installed and setup provides the capability to trigger pipelines when a new cluster is added. In the context of ckcp, what is the intent behind installing PaC?
Making sure that we can provide a test environment that's is indistinguishable from a production environment. 

> Are we expecting developers using ckcp to add clusters to the setup?
Most of the case, I would not expect them to. But there's no reason they should not be able to, e.g. to test multi-cluster setup.

> The sre directory structure is auto-created under ckcp/work once we run the script. So devs will need to follow the same process as in gitops/readme to accomplish this?
The goal is that the development environment looks exactly like a production environment so we can dogfood all the production scripts during the development phase. If they wanted to add additional clusters, they would therefore do it the same way as for production, except that they would probably call the `cluster-setup/install.sh` and `kcp-registrar/register.sh` manually instead of going through the hassle of setting up a repository with a working CI/CD (but they could if they wanted to).

bnallapeta:
@Roming22 Thank you for the explanation. Yes, it makes sense. By this, we are providing an option to do it the PaC way if they want to all the while with the option to run manual scripts and maintaining consistency with the production env.

bnallapeta:
@Roming22 I missed this one. You need to update the README as well as you changed the script options. This block -https://github.com/openshift-pipelines/pipelines-service/blob/main/ckcp/README.md?plain=1#L40

ckcp/openshift_dev_setup.sh
```diff
@@ -265,53 +248,39 @@ patches:
...
+  "$SCRIPT_DIR/../images/cluster-setup/install.sh" --workspace-dir "$WORK_DIR"
+
+  echo "  - Install Pipelines as Code:"
+  # Passing dummy values to the parameters of the pac/setup.sh script
+  # because we only want to install the runner side of resources.
+  GITOPS_REPO="https://example.git.com/my/repo" GIT_TOKEN="placeholder_token" \
+    WEBHOOK_SECRET="placeholder_webhook" \
+    "$GITOPS_DIR/pac/setup.sh"
+
+  echo -n "  - Register compute to KCP"
```
ramessesii2, line 281: I get this while running the script
```
- Register compute to KCPWorkspace: root:default:computeapi-ci-ln-31z9c6t-72292-origin-ci-int-gce-dev-rhcloud-com
```
This line is partially from this script i.e. - `- Register compute to KCP` and the rest of it comes from `register.sh`. 
->  If the intent is to show the cluster info in the same line, please add a space after `to KCP`.

ramessesii2 (approved):
This works great for me. Minor comments related to aesthetics.

ckcp/openshift_dev_setup.sh
Roming22: Fixed

bnallapeta (approved):

Issue #140, opened 2022-07-05
SRE team should have the option to provide custom resources to sync to the kcp syncer


ramessesii2:
## Current State
The [`register.sh`](https://github.com/openshift-pipelines/pipelines-service/blob/main/images/kcp-registrar/register.sh#L133) script and [`openshift_dev_setup.sh`](https://github.com/openshift-pipelines/pipelines-service/blob/main/ckcp/openshift_dev_setup.sh#L18) currently has a flag/parameter (resources) to sync the list of resources between kcp and workload clusters. For now, we've a fixed number of resources to sync.

## Desired State
With different customers, the resources to sync may vary and we want the customer's SRE team to have control over the resources to be synced based on their requirement. Provide an additional flag (eg: CR_TO_SYNC) in register.sh and openshift_dev_setup.sh scripts so that users can provide their custom resources to be synced.

fgiloux:
@ramessesii2 although I do agree that we may need to make it configurable I tend to disagree that we may want to support different resources to be synced for different customers.
My point of view is that the service offered is built with a defined set of resources getting synced and may not work properly if some of these resources are not synced and may not be secure if additional resources that have not been validated get synced.
I am happy to listen to your arguments to convince me otherwise.

ramessesii2:
Hi @fgiloux, I wanted this change so that we could have other resources like NetworkPolicy synced as and when there's necessity. We could have this as hard-coded and as one of the resources that Pipelines-Service maintains but the other argument against it was - "not every user would require this resource".  
> if some of these resources are not synced and may not be secure if additional resources that have not been validated get synced.

Indeed, I completely agree with this statement as well.

ramessesii2:
Closing this issue.
For retrospective - Pipelines-Service is going to support syncing only defined resources between kcp workspaces and computes.

PR #145, opened 2022-07-06
feature: Add option to sync additional resources

Roming22:
Users may need to have the kcp syncer sync custom resources (e.g. NetworkPolicies).
A new KCP_EXTRA_CRDS variable is introduced to customize the extra resources to synchronize.

### DIFF
--- a/ckcp/openshift_dev_setup.sh
+++ b/ckcp/openshift_dev_setup.sh
@@ -30,6 +30,9 @@ Usage:
 Setup the pipeline service on a cluster running on KCP.
 
 Optional arguments:
+    --extra-crds
+        Comma separated list of extra CRDs that need to be synchronized
+        by the KCP syncer.
     -d, --debug
         Activate tracing/debug mode.
     -h, --help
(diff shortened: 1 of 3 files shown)


bnallapeta:
@Roming22 LGTM. Also, we want to make the same changes with ckcp as well right?

ramessesii2 (commented):
Thank you for this PR and LGTM.  
+1 for Bharath's comment

Roming22:
@bnallapeta @ramessesii2 Now that #143 has been merged, I've added the networkpolicies to `ckcp/openshift_dev_setup.sh`.

Roming22:
Added the `--extra-crds` option to `ckcp/openshift_dev_setup.sh` as discussed.

ckcp/openshift_dev_setup.sh
```diff
@@ -30,6 +30,9 @@ Usage:
 Setup the pipeline service on a cluster running on KCP.
 
 Optional arguments:
+    --extra-crds
+        Comma separated list of extra CRDs that need to by synchronized
```
xinredhat, line 34: typo? `by` -> `be`

Roming22: Fixed

fgiloux:
As per my comment on issue #140 I have some concerns with the approach. The key point is that we need to be careful in what configuration we allow for core components. The rational is that things need to be validated, quality engineering needs to be done and we quickly reach an explosion of combinations when multiple and here unbounded configurations are allowed for core components, whose behavior is key for other components of the platform to work as expected. I would not have similar concerns for components that are at the periphery of the platform.

PR #146, opened 2022-07-07
added ci jobs for code analysis

xinredhat:
**Changes**
This is to provide a CI job for code analysis against pipelines-service repo.

1. Dockerfile is under directory `images/shellcheck`
2. Github Action workflow `.github/workflows/shellcheck-image.yaml` is for building and publishing the shellcheck image
3. CI job `.tekton/code-analysis.yaml`


bnallapeta:
@xinredhat As discussed, please raise a single PR with this as well as the CI Pipeline changes. Merging this PR (shellcheck image) and not using it anywhere is something we should avoid in our code base.

.github/workflows/shellcheck-image.yaml
```diff
@@ -0,0 +1,47 @@
...
+      sha_short: ${{ steps.vars.outputs.sha_short }}
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Get the short sha
+        id: vars
+        run: echo "::set-output name=sha_short::$(echo ${{ github.sha }} | cut -b -7)"
+
+      # Build and push a kcp-registrar image, tagged with latest and the commit SHA.
```
Roming22, line 28: Update comment

images/shellcheck/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
+LABEL build-date= \
+    com.redhat.build-host= \
+    description="This image provides shecllcheck binary." \
```
Roming22, line 4: Typo. Repeats multiple times in the file.

Roming22 (commented):
I'm not sure I'm comfortable with the idea that resources used for testing are located in the same place as 'production' resources. Let's discuss it during stand-up.

images/shellcheck/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+    summary="Provides the 0.7.1 shecllcheck" \
+    url="https://github.com/openshift-pipelines/pipelines-service/tree/main/images/shellcheck" \
+    vcs-ref=  \
+    vcs-type="git" \
+    vendor="Pipelines Service" \
+    version="0.1"
+WORKDIR /
+RUN mkdir /workspace && chmod 777 /workspace && chown 65532:65532 /workspace
+ENV HOME /tmp/home
+RUN mkdir $HOME && chmod 777 $HOME && chown 65532:65532 $HOME
```
ramessesii2, line 20: Can we have as few `RUN` inside dockerfile as possible. It is because -
```
The RUN instruction will execute any commands in a new layer on top of the current image and commit the results. The resulting committed image will be used for the next step in the Dockerfile.

Layering RUN instructions and generating commits conforms to the core concepts of Docker where commits are cheap and containers can be created from any point in an image’s history, much like source control.
```
You can refer this as well https://docs.docker.com/engine/reference/builder/#run

xinredhat:
> I'm not sure I'm comfortable with the idea that resources used for testing are located in the same place as 'production' resources. Let's discuss it during stand-up.

@Roming22 I agree with you. My thought is moving shellcheck installation into CI job task. WDYT?

bnallapeta:
> I'm not sure I'm comfortable with the idea that resources used for testing are located in the same place as 'production' resources. Let's discuss it during stand-up.

@Roming22 Not sure I understand this correctly. Could you elaborate on this a bit more?
Just to provide a couple examples of many, here are repos placing their 'testing' workflows in the same repo as the production code.

https://github.com/github/super-linter/blob/main/.github/workflows/trivy.yml
https://github.com/stackrox/kube-linter/blob/main/.github/workflows/build.yaml

Roming22:
> Could you elaborate on this a bit more?

I'm perfectly fine with the testing workflow being part of the repo (and I would be worried if that was not the case). However resources used for testing should be in a separate folder instead of being mixed with the "production" code. So in this case, if we need to create images for testing, I would expect those images to be under something like `test/images`.

Roming22:
> My thought is moving shellcheck installation into CI job task. WDYT?

Installing shellcheck should be fast enough that I don't see repeating the action every time being an issue.

bnallapeta:
> > Could you elaborate on this a bit more?
> 
> I'm perfectly fine with the testing workflow being part of the repo (and I would be worried if that was not the case). However resources used for testing should be in a separate folder instead of being mixed with the "production" code. So in this case, if we need to create images for testing, I would expect those images to be under something like `test/images`.

@Roming22  Ahh, you were talking about the folder segregation. I went into a different train of thought :)
Yeah, I am +1 for the folder segregation.

xinredhat:
> 

Fixed. the image path is changed to `ghcr.io/openshift-pipelines/test/shellcheck`.

images/shellcheck/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
```
Roming22, line 1: This file should be under `test/images/shellcheck`, to keep the test code separate from the "production" code.

xinredhat: thanks. fixed.

.github/workflows/shellcheck-image.yaml
```diff
@@ -0,0 +1,47 @@
+name: Build and Publish shellcheck Image
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "images/shellcheck/**"
```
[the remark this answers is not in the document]
xinredhat: fixed.

.tekton/code-analysis.yaml
```diff
@@ -0,0 +1,74 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: code-analysis-pipeline
```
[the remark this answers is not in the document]
xinredhat: fixed.

bnallapeta:
@xinredhat As discussed, do test it locally as I was getting 'workspace binding "source" does not match any declared workspace' error. Let me know once you fix it.

bnallapeta (approved):
Verified it by running it on a test repo. Shellcheck works as expected.

PR #147, opened 2022-07-07
removed the dependency of kubectl cert-manager plugin

xinredhat:
**Changes**
Removed  kubectl `cert-manager` plugin and check if the service is ready or not by using `kubectl` directly.


bnallapeta (approved):
LGTM. @Roming22 PTAL and merge the PR if you're okay with it too.

ckcp/openshift_dev_setup.sh
```diff
@@ -176,12 +176,18 @@ install_cert_manager() {
...
     kubectl apply -f "$CKCP_DIR/argocd-apps/$APP.yaml" >/dev/null 2>&1
     argocd app wait "$APP" >/dev/null 2>&1
-    # Wait until cert manager is ready
-    kubectl cert-manager check api --wait=5m  >/dev/null 2>&1
   fi
+
+  # Wait until cert-manager is ready
+  pods_cmd="kubectl -n openshift-cert-manager get pods -o=jsonpath='{range .items[*]}{.metadata.name}{\" \"}{.status.containerStatuses[*].ready}{\"\n\"}{end}'"
+  until [[ $(eval $pods_cmd | wc -l) -eq 3 ]]; do 
+    sleep 10
```
Roming22, line 189: `sleep 5`. Please add a `echo -n "."` as well so user can see something's happening.

PR #151, opened 2022-07-13
Configure SyncTarget workspace and update APIBinding documentation

Roming22:
The process to bind to an APIExport now requires the right RBAC to be
granted to the user(s).

### DIFF
--- a/gitops/README.md
+++ b/gitops/README.md
@@ -76,7 +76,7 @@ _Note: Please run these scripts only if you intend not to use PaC to automate th
     b. To access and setup kcp
 
     ```
-    $ KCP_ORG="root:pipelines-service" KCP_WORKSPACE="compute" DATA_DIR="/home/workspace/pipelines-service/gitops/sre" ../images/kcp-registrar/register.sh
+    $ KCP_ORG="root:pipelines-service" KCP_WORKSPACE="compute" WORKSPACE_DIR="/home/workspace/pipelines-service/gitops/sre" ../images/kcp-registrar/register.sh
     ```
 
 ### Pipelines As Code (PaC)
@@ -172,6 +172,8 @@ spec:
   - rev-158351.deployments.apps
 
 ```
+
+Users must be authorized to bind APIExports on the workspace hosting the SyncTarget, or they will get an `unable to create APIImport: missing verb='bind' permission on apiexports` error. The default configuration will grant the `bind` permission to all authenticated users (c.f. `gitops/kcp/registration`). This configuration can be customized by creating the appropriate manifest(s) in the `environment/kcp` folder (c.f. [this example](../docs/sre/examples/providers/environment/kcp/kustomization.yaml)).
 
 Below snippet shows the APIBinding used in a new workspace named 'user'.
 
--- a/images/kcp-registrar/register.sh
+++ b/images/kcp-registrar/register.sh
@@ -20,16 +20,81 @@ set -o pipefail
 
 usage() {
 
-    printf "Usage: KCP_ORG="root:pipelines-service" KCP_WORKSPACE="compute" KCP_SYNC_TAG="release-0.5" DATA_DIR="/workspace" ./register.sh\n\n"
+    printf "
+Usage:
+    %s [options]
 
-    # Parameters
-    printf "The following parameters need to be passed to the script:\n"
-    printf "KCP_ORG: the organization for which the workload clusters need to be registered, i.e.: root:pipelines-service\n"
-    printf "KCP_WORKSPACE: the name of the workspace where the workload clusters get registered (created if it does not exist), i.e: compute\n"
-    printf "KCP_SYNC_TAG: the tag of the kcp syncer image to use (preset in the container image at build time and leveraged by the PipelineRun)\n"
-    printf "DATA_DIR: the location of the cluster files\n"
-    printf "          a single file with extension kubeconfig is expected in the subdirectory: credentials/kubeconfig/kcp\n"
-    printf "          kubeconfig files for compute clusters are expected in the subdirectory: credentials/kubeconfig/compute \n\n"
+Deploy Pipelines Service on the clusters as per the configuration in WORKSPACE_DIR.
+
+Mandatory arguments:
+    --kcp-org KCP_ORG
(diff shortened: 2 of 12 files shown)


gitops/README.md
```diff
@@ -173,6 +173,39 @@ spec:
...
+      - bind
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: apibinding:kubernetes:all-users
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: apibinding-kubernetes
```
xinredhat, line 201: this ClusterRole name doesn't match with `apibinding:kubernetes` defined in ClusterRole.

Roming22: Fixed.

Roming22:
@bnallapeta Please have a look and merge if you're OK with the changes.

Roming22:
> In my opinion when we create the workspace for the registration of the workload clusters we should create the clusterrole and clusterrolebinding you have documented here. In the documentation (should also be in the operational manual) we should instruct users how they can use kustomize and overlays to restrict access to the workload clusters to a smaller group than system:authenticated

I agree that it would make the UX better. However, I'm a bit wary about having the default security setting be "opened to all".

fgiloux:
> I agree that it would make the UX better. However, I'm a bit wary about having the default security setting be "opened to all".

Obfuscation is usually not seen as security.
Moreover this gives the opportunity to get feedback on how security can get improved, which is not possible when things are hidden.

gitops/README.md
```diff
@@ -173,6 +173,39 @@ spec:
...
+kind: ClusterRole
+metadata:
+  name: apibinding:kubernetes
+rules:
+  - apiGroups:
+      - apis.kcp.dev
+    resources:
+      - apiexports
+    resourceNames:
+      - kubernetes # value of the spec/indentity/secretRef/name field of the APIExport
```
bnallapeta, line 190: Do you mean 'identity'?

Roming22: Fixed

bnallapeta:
@Roming22 
1. There's a conflict currently.
2. I don't remember seeing so many files earlier. Is there another commit here by mistake?

Roming22:
@bnallapeta 

> 1. Fixed
> 2. Originally, I was just updating the docs. Frederic suggested automating the process. This required some changes to make it customizable. I've updated the commit message accordingly.

Roming22:
@bnallapeta I've processed your latest comment, let me know if that's not what you were expecting to see.

images/kcp-registrar/register.sh
```diff
@@ -20,16 +20,81 @@ set -o pipefail
...
+        Example: 'compute'.
+        Can be set through \$KCP_WORKSPACE.
+    --kcp-sync-tag KCP_SYNC_TAG
+        Tag of the kcp syncer image to use (preset in the container image at build time
+        and leveraged by the PipelineRun).
+        Example: 'v0.6.0-alpha0'
+        Can be set through \$KCP_SYNC_TAG.
+
+Optional arguments:
+    -w, --workspace-dir WORKSPACE_DIR
```
fgiloux, line 46: You need to be careful with such changes. Besides the fact that it has nothing to do with the stated PR goal and that it is disputable to see that as an "improvement" this is technically breaking the API.
DATA_DIR is referenced in multiple places:
- docs/kcp-registration.md
- gitops/sre/.tekton/kcp-registration.yaml
- gitops/pac/.tekton/kcp-registration.yaml
- DEVELOPMENT.md

images/kcp-registrar/register.sh
```diff
@@ -20,16 +20,81 @@ set -o pipefail
...
+        Kubeconfig files for compute clusters are expected in the subdirectory: credentials/kubeconfig/compute
+        Default: \$WORKSPACE_DIR or current directory if the environment
+        variable is unset.
+    -d, --debug
+        Activate tracing/debug mode.
+    -h, --help
+        Display this message.
+
+Example:
+    KCP_SYNC_TAG='v0.6.0-alpha0' %s --kcp-org 'root:my_org' --kcp-workspace 'my_workspace' --workspace_dir /path/to/my_dir
```
fgiloux, line 58: we should use the release image for the example

images/kcp-registrar/register.sh
```diff
@@ -132,21 +199,38 @@ register() {
...
-prechecks
-kcp_kubeconfig
-if [[ "$(KUBECONFIG=${kcp_kcfg} kubectl kcp workspace current | cut -d\" -f2)" == "$KCP_ORG:$KCP_WORKSPACE" ]]; then
-    printf "Workspace: %s" "$KCP_ORG:$KCP_WORKSPACE"
-else
-    printf "Switching to organization %s\n" "${KCP_ORG}"
-    switch_org
-    printf "Switching to workspace %s\n" "${KCP_WORKSPACE}"
-    switch_ws
+set_kcp_workspace() {
```
fgiloux, line 208: I am really not convinced by the name of the function and the log message. What the function is doing is applying a set of manifests. These manifests:
- currently are setting up RBAC
- aiming to configure access to a compute pool

fgiloux (changes requested):

images/kcp-registrar/register.sh
Roming22: Done

images/kcp-registrar/register.sh
Roming22: While at the moment it's only granting access to the APIExport, it can actually configure any resource in the kcp workspace if the manifests are added to the `environment/kcp` directory.

If you have an idea for a better name, share it with me and I'll update it.

Roming22:
@fgiloux @bnallapeta Is there anything holding the merge of this PR?

images/kcp-registrar/register.sh
fgiloux: proposal: configure_compute_ws()

fgiloux:
> Is there anything holding the merge of this PR?

@Roming22, we would have quicker iterations if we keep the PR scope focused on 1 thing.
As a result I have changed the title of the PR. Feel free to amend if you think that it could be better described.

PR #126, opened 2022-06-28
Implement tekton-chains

Roming22:
Configuration is taken verbatim from https://github.com/redhat-appstudio/infra-deployments/tree/ef1f8459a0ab86964c1421f7475fb41626cec905/components/build/tekton-chains

Signed-off-by: Romain Arnaud <rarnaud@redhat.com>

PR can be tested by running `./ckcp/openshift_dev_setup.sh` followed by `./ckcp/test.sh chains`.

PR #152, opened 2022-07-13
feature: Install rekor when using ckcp

Roming22:
Rekor is required to store the data for tekton-chains.

The content of this PR will be required when we'll want to deploy tekton-chains.

### DIFF
--- a/ckcp/openshift_dev_setup.sh
+++ b/ckcp/openshift_dev_setup.sh
@@ -76,7 +76,7 @@ parse_args() {
 }
 
 init() {
-  APP_LIST="openshift-gitops ckcp compute"
+  APP_LIST="openshift-gitops rekor ckcp compute"
   cluster_type="openshift"
 
   # Create SRE repository folder
@@ -101,6 +101,8 @@ init() {
   kcp_org="root:default"
   kcp_workspace="pipelines-service-compute"
   kcp_version="$(yq '.images[] | select(.name == "kcp") | .newTag' "$SCRIPT_DIR/openshift/overlays/dev/kustomization.yaml")"
+
+  domain_name=$(kubectl get ingresses.config.openshift.io cluster -o json|jq -r ".spec.domain")
 }
 
 # To ensure that dependencies are satisfied
@@ -222,7 +224,6 @@ install_ckcp() {
   cp -rf "$ckcp_dev_dir" "$ckcp_temp_dir"
 
   local ckcp_route
-  domain_name="$(kubectl get ingresses.config/cluster -o jsonpath='{.spec.domain}')"
   ckcp_route="ckcp-ckcp.$domain_name"
   echo "
 patches:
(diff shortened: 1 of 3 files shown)


ckcp/openshift_dev_setup.sh
```diff
@@ -281,6 +281,30 @@ patches:
   KUBECONFIG_KCP="$WORK_DIR/credentials/kubeconfig/kcp/ckcp-ckcp.default.compute.kubeconfig"
 }
 
+install_rekor() {
+  # Get our domain from the ingresses config of the cluster
+  echo -n "  - Domain: "
+  domain=$(kubectl get ingresses.config.openshift.io cluster -o json|jq -r ".spec.domain")
```
xinredhat, line 287: There is the same statement in the function `install_ckcp`([link](https://github.com/openshift-pipelines/pipelines-service/blob/main/ckcp/openshift_dev_setup.sh#L212)). To avoid the duplicate statement in anywhere, would you mind moving it into one place, such as `init` function? So that we can use `domain` variable in other places.

ckcp/openshift_dev_setup.sh
```diff
@@ -281,6 +281,30 @@ patches:
...
+  kubectl apply -k "$rekor_dir/$cluster_type"
+
+  rekor_url="https://rekor-server.$domain/api/v1/log"
+  echo -n "  - Wait for api on $rekor_url: "
+  while ! curl --fail --insecure --output /dev/null --silent "$rekor_url"; do
+    echo -n .
+    sleep 3
+  done
+  echo "OK"
+  exit 0
```
xinredhat, line 305: Do we really need to explicitly execute `exit 0`?

ckcp/openshift_dev_setup.sh
Roming22: Done

ckcp/openshift_dev_setup.sh
Roming22: Good catch. I forgot to remove it after testing. Fixed.

bnallapeta:
@Roming22 Does #126 needs to be merged before this? If so, please take a look at that one as it it currently marked as Draft.

Roming22:
@bnallapeta This PR can be merged idependently from #126.

Roming22:
@bnallapeta 
1. We're not sure which rekor instance is going to be used for staging/prod in the long term. Right now they are using a public rekor instance, hence why we only need it in `ckcp` at the moment.
2. Done.
3. I'm not sure I understand which path you're pointing me at. Can you please either specify the file/line or open a comment?
4. I've lifted the install straight from the `hack` scripts in `infra-deployments`. I had a look at the rekor operator, but the [git repo](https://github.com/sigstore/rekor-operator) clearly state that it's not ready to be used.

bnallapeta:
@Roming22 
In general, I am not confident on what is the correct installation method or the usability thereafter. But if we want to stay in line with the way HACBS is doing it, I guess that should be good. 
(ignore my comment on the path. Looks right)

@fgiloux Please provide any thoughts (merge the PR if all is well :p).

Roming22:
After discussion, there is no need to deploy rekor on the cluster. For testing, a public rekor instance is good enough.

PR #154, opened 2022-07-15
Upgrade kcp to release-0.6

Roming22:

### DIFF
--- a/.github/workflows/kcp-registrar-image.yaml
+++ b/.github/workflows/kcp-registrar-image.yaml
@@ -36,7 +36,7 @@ jobs:
           containerfiles: |
             ./images/kcp-registrar/Dockerfile
           build-args: |
-            KCP_BRANCH=release-0.5
+            KCP_BRANCH=release-0.6
 
       - name: Push to ghcr.io
         id: push-to-ghcr
--- a/ckcp/openshift/overlays/dev/kustomization.yaml
+++ b/ckcp/openshift/overlays/dev/kustomization.yaml
@@ -6,6 +6,6 @@ resources:
 images:
 - name: kcp
   newName: ghcr.io/kcp-dev/kcp
-  newTag: release-0.5
+  newTag: release-0.6
 
 namespace: ckcp
--- a/ckcp/openshift_dev_setup.sh
+++ b/ckcp/openshift_dev_setup.sh
@@ -51,6 +51,8 @@ parse_args() {
     -w | --work-dir)
       shift
       WORK_DIR="$1"
+      mkdir -p "$WORK_DIR"
+      WORK_DIR="$(cd "$1" >/dev/null; pwd)"
       ;;
     -d | --debug)
       set -x
@@ -96,7 +98,8 @@ init() {
   KUBECONFIG="$WORK_DIR/credentials/kubeconfig/compute/compute.kubeconfig.base"
   KUBECONFIG_MERGED="merged-config.kubeconfig:$KUBECONFIG:$KUBECONFIG_KCP"
   export KUBECONFIG
-  kcp_workspace="compute"
+  kcp_org="root:default"
+  kcp_workspace="pipelines-service-compute"
   kcp_version="$(yq '.images[] | select(.name == "kcp") | .newTag' "$SCRIPT_DIR/openshift/overlays/dev/kustomization.yaml")"
 }
 
@@ -262,12 +265,11 @@ patches:
   podname="$(kubectl get pods --ignore-not-found -n "$ns" -l=app=kcp-in-a-pod -o jsonpath='{.items[0].metadata.name}')"
   mkdir -p "$(dirname "$KUBECONFIG_KCP")"
   # Wait until admin.kubeconfig file is generated inside ckcp pod
-  while [[ $(kubectl exec -n $APP "$podname" -- ls /etc/kcp/config/admin.kubeconfig >/dev/null 2>&1; echo $?) -ne 0 ]]; do
+  while [[ $(kubectl exec -n $ns "$podname" -- ls /etc/kcp/config/admin.kubeconfig >/dev/null 2>&1; echo $?) -ne 0 ]]; do
     echo -n "."
     sleep 5
   done
-  kubectl cp "$APP/$podname:/etc/kcp/config/admin.kubeconfig" "$KUBECONFIG_KCP" >/dev/null 2>&1
-  KUBECONFIG="$KUBECONFIG_KCP" kubectl config rename-context "default" "workspace.kcp.dev/current" >/dev/null
+  kubectl cp "$ns/$podname:/etc/kcp/config/admin.kubeconfig" "$KUBECONFIG_KCP" >/dev/null 2>&1
   echo "OK"
 
   # Check if external ip is assigned and replace kcp's external IP in the kubeconfig file
@@ -279,16 +281,19 @@ patches:
 
   # Workaround to prevent the creation of the creation of a new workspace until KCP is ready.
   # This fixes `error: creating a workspace under a Universal type workspace is not supported`.
-  while ! KUBECONFIG="$KUBECONFIG_KCP" kubectl kcp workspace create "dummy" --ignore-existing >/dev/null; do
+  ws_name=$(echo "$kcp_org" | cut -d: -f2)
+  while ! KUBECONFIG="$KUBECONFIG_KCP" kubectl kcp workspace create "$ws_name" --type root:organization --ignore-existing >/dev/null; do
     sleep 5
   done
+  KUBECONFIG="$KUBECONFIG_KCP" kubectl kcp workspace use "$ws_name"
 
   echo "  - Setup kcp access:"
   "$SCRIPT_DIR/../images/access-setup/content/bin/setup_kcp.sh" \
     --kubeconfig "$KUBECONFIG_KCP" \
+    --kcp-org "$kcp_org" \
     --kcp-workspace "$kcp_workspace" \
     --work-dir "$WORK_DIR"
-  KUBECONFIG_KCP="$WORK_DIR/credentials/kubeconfig/kcp/ckcp-ckcp.default.compute.kubeconfig"
+  KUBECONFIG_KCP="$WORK_DIR/credentials/kubeconfig/kcp/ckcp-ckcp.${ws_name}.${kcp_workspace}.kubeconfig"
 }
 
 install_compute() {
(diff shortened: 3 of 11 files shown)


fgiloux:
reviewing

.github/workflows/kcp-registrar-image.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
           containerfiles: |
             ./images/kcp-registrar/Dockerfile
           build-args: |
-            KCP_BRANCH=release-0.5
+            KCP_BRANCH=v0.6.0-alpha.0
```
fgiloux, line 39: so far we have been following a release branch, rather than using a tagged version. We may want to change that in the future but I think that we should keep it this way for now, hence: release-0.6

ckcp/openshift/overlays/dev/kustomization.yaml
```diff
@@ -6,6 +6,6 @@ resources:
 images:
 - name: kcp
   newName: ghcr.io/kcp-dev/kcp
-  newTag: release-0.5
+  newTag: v0.6.0-alpha.0
```
fgiloux, line 9: as above

fgiloux:
@Roming22 have you tested it locally?

1. in local/kcp/start.sh
default branch should also be modified:
KCP_BRANCH="${KCP_BRANCH:-release-0.5}"

2. Error from server: error when creating "/tmp/romain/pipelines-service/local/kcp/pipelines-service-org.yaml": failed to convert unstructured to ClusterWorkspace: cannot restore struct from: string
When I had tried it from main I needed to add the path:
apiVersion: tenancy.kcp.dev/v1alpha1
kind: ClusterWorkspace
metadata:
  name: pipelines-service
spec:
  type:
    name: Organization
    path: root

Could you please check the above and let me know when it is addressed? I can then resume with testing

.github/workflows/kcp-registrar-image.yaml
Roming22: The release was just tagged this morning, hence why I was using the alpha version up to this point. Fixed.

ckcp/openshift/overlays/dev/kustomization.yaml
Roming22: Fixed

Roming22:
> @Roming22 have you tested it locally?
> 
> 1. in local/kcp/start.sh
>    default branch should also be modified:
>    KCP_BRANCH="${KCP_BRANCH:-release-0.5}"
> 2. Error from server: error when creating "/tmp/romain/pipelines-service/local/kcp/pipelines-service-org.yaml": failed to convert unstructured to ClusterWorkspace: cannot restore struct from: string
>    When I had tried it from main I needed to add the path:
>    apiVersion: tenancy.kcp.dev/v1alpha1
>    kind: ClusterWorkspace
>    metadata:
>    name: pipelines-service
>    spec:
>    type:
>    name: Organization
>    path: root
> 
> Could you please check the above and let me know when it is addressed? I can then resume with testing

I can't test the content of `local`, I've got issues with my podman configuration that is preventing me from running it. I test all my changes by spawning a cluster through clusterbot and then using the `ckcp/openshift_dev_setup.sh` script, which I was able to do successfully.

1. Forever fixed.
2. Did you upgrade the kcp plugin before running the test?

local/kcp/start.sh
```diff
@@ -40,7 +40,7 @@ kcp-binaries() {
     git clone https://github.com/kcp-dev/kcp.git
     KCP_DIR="${KCP_PARENT_DIR}/kcp"
     pushd kcp
-    KCP_BRANCH="${KCP_BRANCH:-release-0.5}"
+    KCP_BRANCH="${KCP_BRANCH:-$(yq '.images[] | select(.name == "kcp") | .newTag' "$PARENT_PATH/../../ckcp/openshift/overlays/dev/kustomization.yaml")}"
```
fgiloux, line 43: this change reduces readability and creates unnecessary dependencies, hence fragility.
If you don't want to change the default version in multiple places let's have a file with them.

fgiloux:
> 2. Did you upgrade the kcp plugin before running the test?

yes I have but the fact that the apiresourceschemas has changed has little to do with the client:
~~~ 
            type:
              description: "type defines properties of the workspace both on creation
                (e.g. initial resources and initially installed APIs) and during runtime
                (e.g. permissions). If no type is provided, the default type for the
                workspace in which this workspace is nesting will be used. \n The
                type is a reference to a ClusterWorkspaceType in the listed workspace,
                but lower-cased. The ClusterWorkspaceType existence is validated at
                admission during creation. The type is immutable after creation. The
                use of a type is gated via the RBAC clusterworkspacetypes/use resource
                permission."
              properties:
                name:
                  description: name is the name of the ClusterWorkspaceType
                  pattern: ^[A-Z][a-zA-Z0-9]+$
                  type: string
                path:
                  description: path is an absolute reference to the workspace that
                    owns this type, e.g. root:org:ws.
                  pattern: ^root(:[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
                  type: string
              required:
              - name
              - path
              type: object
~~~

Could you please make the change I mentioned in my previous comment?

Roming22:
> Could you please make the change I mentioned in my previous comment?

Done

local/kcp/start.sh
Roming22: I tried to improve readabilty.

I'm not sure I want to start refactoring/implementing a single source of truth for version with this PR, but I agree that it would be nice to have at some point, so I've logged it in PLNSRVCE-498.

local/kcp/start.sh
```diff
@@ -40,7 +40,8 @@ kcp-binaries() {
     git clone https://github.com/kcp-dev/kcp.git
     KCP_DIR="${KCP_PARENT_DIR}/kcp"
     pushd kcp
-    KCP_BRANCH="${KCP_BRANCH:-release-0.5}"
+    kcp_default_version=$(yq '.images[] | select(.name == "kcp") | .newTag' "$PARENT_PATH/../../ckcp/openshift/overlays/dev/kustomization.yaml")
+    KCP_BRANCH="${KCP_BRANCH:-$kcp_default_version}"
     git checkout "${KCP_BRANCH}"
     make build
     popd
```
fgiloux, line 47: under line 92: kubectl --kubeconfig="${KUBECONFIG}" config use-context root
it is now needed to navigate back to the root workspace:
~~~
KUBECONFIG="${KUBECONFIG}" kubectl kcp workspace use root
~~~

fgiloux (changes requested):
These changes allow at least to start kcp locally

fgiloux:
@Roming22 have you tested the registration?

~~~
$ KCP_ORG="root:pipelines-service" KCP_WORKSPACE="compute" KCP_SYNC_TAG="release-0.6" DATA_DIR="/tmp/gitops-ps/gitops/sre" images/kcp-registrar/register.sh
Switching to organization root:pipelines-service
Current workspace is "root:pipelines-service".
Switching to workspace compute
  - Create workspace compute
Current workspace is "root:pipelines-service:compute".
Registering clusters to kcp
Getting the list of registered clusters
error: the server doesn't have a resource type "workloadclusters"
~~~

fgiloux:
I validated that registration is working

images/access-setup/content/bin/setup_kcp.sh
```diff
@@ -140,7 +140,7 @@ generate_kcp_credentials() {
   printf "    - Create workspace:\n"
   kubectl kcp workspace use "$KCP_ORG"
   if ! kubectl kcp workspace use "$KCP_WORKSPACE" >/dev/null 2>&1; then
-    kubectl kcp workspace create "$KCP_WORKSPACE" --enter >/dev/null
+    kubectl kcp workspace create "$KCP_WORKSPACE" --type universal --enter >/dev/null
```
fgiloux, line 143: no change required: --type universal is the default and unnecessary

local/kcp/pipelines-service-org.yaml
```diff
@@ -4,3 +4,4 @@ metadata:
   name: pipelines-service
 spec:
   type: Organization
+  path: root
```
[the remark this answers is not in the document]
Roming22: Sorry about that, it should be OK now.

local/kcp/start.sh
Roming22: Done.

local/kcp/pipelines-service-org.yaml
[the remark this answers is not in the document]
Roming22: I was able to validate that `local/kcp/start.sh` completes locally.

ckcp/openshift_dev_setup.sh
```diff
@@ -267,7 +270,7 @@ patches:
     sleep 5
```
fgiloux, line 270: it does not originate from this PR but there are a couple of occurrences where $APP is used instead of $ns. Both variables contain the same value but ns was defined for the purpose and it is just an arbitrary decision to have the name of the namespace being the same as the application name: ckcp.

ckcp/openshift_dev_setup.sh
```diff
@@ -308,7 +314,7 @@ install_compute() {
     "$GITOPS_DIR/pac/setup.sh"
 
   echo "  - Register compute to KCP"
-  KCP_ORG="root:default" KCP_WORKSPACE="$kcp_workspace" DATA_DIR="$WORK_DIR" KCP_SYNC_TAG="$kcp_version" \
+  KCP_ORG="$kcp_org" KCP_WORKSPACE="$kcp_workspace" DATA_DIR="$WORK_DIR" KCP_SYNC_TAG="$kcp_version" \
```
fgiloux, line 317: probably due to register.sh but I have hundreds of contexts in my home kubeconfig and I get thousand of lines displayed. Also I get two synctargets created, one not be relevant (for a dead cluster)

fgiloux (commented):
Tests were successful, although there are a couple of things that are not clean

ckcp/openshift_dev_setup.sh
```diff
@@ -279,16 +282,19 @@ patches:
 
   # Workaround to prevent the creation of the creation of a new workspace until KCP is ready.
```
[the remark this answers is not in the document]
Roming22: Caught and fixed by #151.

ckcp/openshift_dev_setup.sh
Roming22: Too much output: caught and fixed in #151 (see the added -q to grep on line 160).

Regarding the 2 synctargets, I'll try to see if I can reproduce and fix.

Roming22: Fix for too many targets added to #151.

ckcp/openshift_dev_setup.sh
```diff
@@ -267,7 +270,7 @@ patches:
     sleep 5
   done
   kubectl cp "$APP/$podname:/etc/kcp/config/admin.kubeconfig" "$KUBECONFIG_KCP" >/dev/null 2>&1
-  KUBECONFIG="$KUBECONFIG_KCP" kubectl config rename-context "default" "workspace.kcp.dev/current" >/dev/null
+  KUBECONFIG="$KUBECONFIG_KCP" kubectl config rename-context "root" "workspace.kcp.dev/current" >/dev/null
```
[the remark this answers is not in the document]
Roming22: Deprecated, removed.

bnallapeta (approved):
Tested it locally on ckcp and local. Works for me.

fgiloux (approved):

PR #156, opened 2022-07-19
perform a dry-run create of Certificate for cert-manager check 

xinredhat:
**Changes:**
Added one check that attempts to perform a dry-run create of a cert-manager Certificate resource in order to verify that CRDs are installed and all the required webhooks are reachable by the K8S API server.

### DIFF
--- a/ckcp/openshift_dev_setup.sh
+++ b/ckcp/openshift_dev_setup.sh
@@ -178,9 +178,17 @@ install_cert_manager() {
     argocd app wait "$APP" >/dev/null 2>&1
   fi
 
-  # Wait until cert-manager is ready
+  # Wait until cert-manager pods are ready
   pods_cmd="kubectl -n openshift-cert-manager get pods -o=jsonpath='{range .items[*]}{.metadata.name}{\" \"}{.status.containerStatuses[*].ready}{\"\n\"}{end}'"
   until [[ $(eval $pods_cmd | wc -l) -eq 3 ]]; do
+    echo -n "."
+    sleep 5
+  done
+
+  # perform a dry-run create of a cert-manager
+  # Certificate resource in order to verify that CRDs are installed and all the
+  # required webhooks are reachable by the K8S API server.
+  until kubectl create -f $CKCP_DIR/openshift/base/certs.yaml --dry-run >/dev/null 2>&1; do
     echo -n "."
     sleep 5
   done


ckcp/openshift_dev_setup.sh
```diff
@@ -178,12 +178,33 @@ install_cert_manager() {
...
+  # Wait until cert-manager pods are ready
   pods_cmd="kubectl -n openshift-cert-manager get pods -o=jsonpath='{range .items[*]}{.metadata.name}{\" \"}{.status.containerStatuses[*].ready}{\"\n\"}{end}'"
   until [[ $(eval $pods_cmd | wc -l) -eq 3 ]]; do
     echo -n "."
     sleep 5
   done
+
+  # perform a dry-run create of a cert-manager
+  # Certificate resource in order to verify that CRDs are installed and all the
+  # equired webhooks are reachable by the K8S API server.
```
Roming22, line 190: `equired` -> `requried`

ckcp/openshift_dev_setup.sh
```diff
@@ -178,12 +178,33 @@ install_cert_manager() {
...
+  local tmp_dir=$(mktemp -d)
+  echo "
```
Roming22, line 192: I'm not a big fan of in-lining resources in scripts. Maybe you could create the yaml in `ckcp/openshift-operators/cert-manager/`? That would get rid of the tmp_dir as well.

ckcp/openshift_dev_setup.sh
xinredhat: fixed

ckcp/openshift_dev_setup.sh
xinredhat: The current in-lining resource of  is more like a temporary  resource, it's useless for any applications. It doens't make sense to create the yaml in  ckcp/openshift-operators/cert-manager/

Roming22: Then the tmp_dir should be cleaned up .

xinredhat: @Roming22 I would like to use this file `ckcp/openshift/base/certs.yaml` for executing dry-run check. WDYT? So  we don't need the in-lining resources in scripts

xinredhat: fixed.

bnallapeta:
@xinredhat @Roming22 with k8s 1.24, --dry-run must be used with client or server. It throws this error otherwise:

```
[bnr@bnr pipelines-service]$ kubectl create -f /home/bnr/workspace/pipelines-service/ckcp/openshift/base/certs.yaml --dry-run
error: --dry-run flag without a value was specified. A value must be set: "none", "server", or "client".
```

have raised #161 to fix this.

Issue #99, opened 2022-06-09
Provide base for pipeline configuration


fgiloux:
In our base directory we should provide a base [TektonConfig](https://tekton.dev/docs/operator/tektonconfig/
) with guidance on common overlay configurations:
- pruning
- enabling feature gates like enable-tekton-oci-bundles, which we may want to enable per default

bnallapeta:
@fgiloux TektonConfig is being provided as part of #122 for pruning purpose. Based on the description, you also wanted to provide other options such as feature flags. But that's specific for every customer right? Is the idea to actually enable them or just to document that it is possible to enable/disable?

fgiloux:
@bnallapeta yes pruning will be addressed when #122 merges.
feature gates are a moving target, the normal lifecycle being that features are not activated by default in early stage, then gets activated by default and finally the feature gate is deprecated and removed when the feature is GA.
Regarding enable-tekton-oci-bundles it is [a feature we want to leverage in early stage](https://github.com/redhat-appstudio/infra-deployments/blob/941b81f5723ff1085706d3f759fd2c2e490ddcc1/components/build/openshift-pipelines/openshift-pipelines-feature-flags.yaml#L13). We should cover that in this or the next sprint.
The other aspect is yes we want to document and provide guidance for further customization.

bnallapeta:
@fgiloux Thanks for the explanation. Let's plan for this in the upcoming sprint. Could you please create a Jira to track this?

fgiloux:
closing this issue following #157

PR #157, opened 2022-07-20
PLNSRVCE-478 : Migration: openshift-pipelines

bnallapeta:
- This PR is about the migration of [this](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/build/openshift-pipelines/) as part of kcp transition Epic.
- We are installing and managing OpenShift Pipelines via the operator using GitOps (ArgoCD). Hence, no changes with the installation itself.
- Regarding the config, all the other parameters are either not necessary or available by default. Have added support for oci-bundles in this PR.
- Pipelines-as-Code will be taken up in a separate story/PR.


fgiloux:
fixes  #99

PR #161, opened 2022-07-20
Fixed issue with --dry-run option

bnallapeta:
with k8s 1.24, --dry-run must be used with a value or either client or server. It throws this error otherwise:

```
[bnr@bnr pipelines-service]$ kubectl create -f /home/bnr/workspace/pipelines-service/ckcp/openshift/base/certs.yaml --dry-run
error: --dry-run flag without a value was specified. A value must be set: "none", "server", or "client".
```


xinredhat:
LGTM

Roming22 (approved):

PR #165, opened 2022-07-21
Shellcheck compliancy

Roming22:
This PR solves some (but not all) shellcheck issues.


fgiloux:
@Roming22 have you tested every single changes? I remember that some "recommendations" were just plain breaking the code

xinredhat:
maybe we can specify `--severity=warning`

Roming22:
@fgiloux There was an issue with `local/kcp/start.sh` where I needed to convert a var into an array,  but otherwise I have the same behavior on this branch and main (with `local/kind/setup.sh` throwing `error: availableReplicas is not found` if I do not use `NO_ARGOCD=true`).

fgiloux:
> @fgiloux There was an issue with `local/kcp/start.sh` where I needed to convert a var into an array, but otherwise I have the same behavior on this branch and main (with `local/kind/setup.sh` throwing `error: availableReplicas is not found` if I do not use `NO_ARGOCD=true`).

I tested it successfully. As discussed on chat you may have hit the issue described in the development readme: [too many open files](https://github.com/openshift-pipelines/pipelines-service/blob/main/DEVELOPMENT.md#troubleshooting)

bnallapeta (approved):
Ran ckcp/openshift_dev_setup.sh locally. LGTM.

PR #171, opened 2022-07-27
bugfix: access_setup

Roming22:
* Check that ArgoCD has been installed on the compute cluster
* Do not change workspace if we're already in the right one

### DIFF
--- a/images/access-setup/content/bin/setup_compute.sh
+++ b/images/access-setup/content/bin/setup_compute.sh
@@ -107,6 +107,14 @@ init() {
   mkdir -p "$credentials_dir"
 }
 
+check_prerequisites() {
+  # Check that argocd has been installed
+  if [[ $(kubectl api-resources | grep -c "argoproj.io/") = "0" ]]; then
+    echo "ArgoCD must be deployed on the cluster first" >&2
+    exit 1
+  fi
+}
+
 generate_compute_credentials() {
   current_context=$(kubectl config current-context)
   compute_name="$(yq '.contexts[] | select(.name == "'"$current_context"'") | .context.cluster' < "$KUBECONFIG" | sed 's/:.*//')"
@@ -130,6 +138,7 @@ main() {
   parse_args "$@"
   prechecks
   init
+  check_prerequisites
   generate_compute_credentials
 }
 
--- a/.github/workflows/access-setup-image.yaml
+++ b/.github/workflows/access-setup-image.yaml
@@ -1,4 +1,4 @@
-name: Build and Publish cluster-setup Image
+name: Build and Publish access-setup Image
 
 on:
   push:
@@ -35,6 +35,8 @@ jobs:
           tags: latest ${{ steps.vars.outputs.sha_short }} ${{ github.ref_name }}
           containerfiles: |
             ./images/access-setup/Dockerfile
+          build-args: |
+            KCP_BRANCH=release-0.6
 
       - name: Push to quay.io
         id: push-to-quay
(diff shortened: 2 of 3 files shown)


images/access-setup/content/bin/setup_kcp.sh
```diff
@@ -138,7 +138,9 @@ generate_kcp_credentials() {
   kubeconfig="$credentials_dir/kcp/$kcp_name.kubeconfig"
 
   printf "    - Create workspace:\n"
-  kubectl kcp workspace use "$KCP_ORG"
+  if [[ "$(kubectl kcp workspace current | cut -d\" -f2)" != "$KCP_ORG" ]]; then
```
fgiloux, line 141: I don't see why it is useful, tested with the latest 0.6 release (server and client):
~~~
$ kubectl kcp workspace use root:pipelines-service
Current workspace is "root:pipelines-service".
$ kubectl kcp workspace use root:pipelines-service
Current workspace is "root:pipelines-service".
$ kubectl kcp workspace use root
Current workspace is "root".
$ kubectl kcp workspace use root:pipelines-service
Current workspace is "root:pipelines-service".
~~~

images/access-setup/content/bin/setup_compute.sh
```diff
@@ -107,6 +107,14 @@ init() {
...
 }
 
+check_prerequisites() {
+  # Check that argocd has been installed
+  if ! kubectl get applications.argoproj.io --ignore-not-found; then
+    echo "ArgoCD must be deployed on the cluster first" >&2
+    exit 1
+  fi
+}
+
```
fgiloux, line 117: Having the API does not guarantee that the operator is installed on the cluster. OLM for instance would not remove the CRD when the operator is uninstalled by default.
If I was to check the API I would rather call `kubectl api-resources` as it is to check the APIs available rather than making a get and waiting it to fail.
[Here](https://github.com/openshift-pipelines/pipelines-service/blob/main/images/cluster-setup/install.sh#L125-L147) for instance the availability of the deployment is checked.

images/access-setup/content/bin/setup_kcp.sh
Roming22: Strange, I have a different behavior:
```
> ./images/access-setup/content/bin/setup_kcp.sh -w ~/Code/appStudio/infra/kcp-stable --kubeconfig ~/Code/appStudio/infra/infra/work/credentials/kubeconfig/kcp/kcp.kubeconfig  -d
...
+ kubectl kcp workspace use root:rh-sso-15826143
error: workspaces.tenancy.kcp.dev "rh-sso-15826143" not found
```

I used kcp release-0.6 (commit 789d0ca1f14eebc783558ce7178044e80e890158), and made sure to recompile.

Bharath is creating an image, let's wait to see if I can reproduce using the image.

images/access-setup/content/bin/setup_compute.sh
Roming22: I had an issue with `grep -q` not behaving as expected. I used `grep -c` instead.

images/access-setup/content/bin/setup_kcp.sh
Roming22: The image is missing a few things, which I've corrected in this PR.

I've reverted this specific change, even though I still see the issue using the new image. If the issue is real, I can take care of it in a separate PR if necessary.

fgiloux:
As discussed in the chat channel. I think that the PR brings these two things that are useful: 
- check that Argo CD has been installed
- build the kcp plugin (till the kcp team makes it available with the release)

I am merging it as you have confirmed that you successfully validated it.

PR #173, opened 2022-07-28
PLNSRVCE-530: Make changes in the repo to move images from ghcr.io to quay.io

bnallapeta:
- Created a new organization in quay.io called redhat-pipeline-service
- Created repositories in quay.io in order to upload respective images
- Created a robot account in order to grant access to GitHub actions to upload images to quay.io
- Updated references from ghcr.io to quay.io throughout the repo - both in GitHub actions and PipelineRuns where it was being used
- Added access-setup workflow.
- Tested it on my repo to verify its working fine (similar steps were followed on my quay account and github repo)

### DIFF
--- a/.tekton/shellcheck-yaml-lint-pipeline.yaml
+++ b/.tekton/shellcheck-yaml-lint-pipeline.yaml
@@ -41,7 +41,7 @@ spec:
         taskSpec:
           steps:
             - name: run-shellcheck
-              image: ghcr.io/openshift-pipelines/test/shellcheck:main
+              image: quay.io/redhat-pipeline-service/shellcheck:main
               script: |
                 #!/usr/bin/env bash
                 find $(workspaces.source.path) -name "*.sh" -exec  shellcheck --severity=warning  {} +
--- a/.github/workflows/access-setup-image.yaml
+++ b/.github/workflows/access-setup-image.yaml
@@ -1,13 +1,13 @@
-name: Build and Publish kcp-registrar Image
+name: Build and Publish cluster-setup Image
 
 on:
   push:
     branches:
       - main
     paths:
-      - "images/kcp-registrar/**"
+      - "images/access-setup/**"
   workflow_dispatch:
-  
+
 jobs:
   build-push:
 
@@ -26,17 +26,15 @@ jobs:
         run: echo "::set-output name=sha_short::$(echo ${{ github.sha }} | cut -b -7)"
 
       # Build and push a kcp-registrar image, tagged with latest and the commit SHA.
-      - name: Build kcp-registrar Image
+      - name: Build access-setup Image
         id: build-image
         uses: redhat-actions/buildah-build@v2
         with:
-          image: kcp-registrar
-          context: ./images/kcp-registrar
+          image: access-setup
+          context: ./images/access-setup
           tags: latest ${{ steps.vars.outputs.sha_short }} ${{ github.ref_name }}
           containerfiles: |
-            ./images/kcp-registrar/Dockerfile
-          build-args: |
-            KCP_BRANCH=release-0.6
+            ./images/access-setup/Dockerfile
 
       - name: Push to quay.io
         id: push-to-quay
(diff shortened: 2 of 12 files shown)


.tekton/shellcheck-yaml-lint-pipeline.yaml
```diff
@@ -41,7 +41,7 @@ spec:
         taskSpec:
           steps:
             - name: run-shellcheck
-              image: ghcr.io/openshift-pipelines/test/shellcheck:main
+              image: quay.io/redhat-pipeline-service/shellcheck:latest
```
fgiloux, line 44: why changing from main to latest?
The CI jobs use the branch name as a tag

fgiloux (commented):
Honest question: does it have to be either/or? Couldn't we have both?
From a security point of view it makes sense to own the name in the common public image repositories: quay, docker, ghcr. It helps with preventing hijacking when users are referencing images through short name (what they should avoid). [Here](https://www.redhat.com/sysadmin/container-image-short-names) is some background.

.tekton/shellcheck-yaml-lint-pipeline.yaml
bnallapeta: Yes, 'main' tag exists. But I thought to call using the more standard 'latest' tag. Should not be much of a difference between latest and main right?

bnallapeta: by default, 3 tags are uploaded with every run - latest, main and commit ID.

bnallapeta:
@fgiloux Why would we want to maintain our images in two registries? Do you see any usecase?

.tekton/shellcheck-yaml-lint-pipeline.yaml
Roming22: I'm fine with also having latest pointing to main until we have a v1 release, at which point will need to make latest point to the latest stable release.

Roming22:
> @fgiloux Why would we want to maintain our images in two registries? Do you see any usecase?

That's related to the doc he linked.
TL;DR: If we don't claim `redhat-pipeline-service` on ghcr.io and docker, someone could claim it, push tainted images there, and an unsuspecting user doing a `pull docker redhat-pipeline-service/setup_compute:latest` could open themselves to harm.

.tekton/shellcheck-yaml-lint-pipeline.yaml
fgiloux: Currently we neither create release branches nor tags. It has been fine so far but at some point, which is coming close now we will need to do it. Often when dependencies change like the kcp version we need to amend our scripts or manifests.
We are now using kcp 0.6 but we should be able to do work to prepare 0.7 and 0.8 without breaking redeployments based on 0.6.  For each release of Pipeline-Service we should have a clear set of dependency versions: kcp, pac, argo cd, etc
Specifically in regard to main vs latest. At least main says what is inside the tin: the latest build of the main branch. Latest can be anything. We don't have multiple branches and tags currently but when we do latest can be a build from any branch, which may be months behind the main branch (if we happen to support and fix a release branch for months).

fgiloux:
> @fgiloux Why would we want to maintain our images in two registries? Do you see any usecase?

As per the above: From a security point of view it makes sense to own the name in the common public image repositories: quay, docker, ghcr. It helps with preventing hijacking when users are referencing images through short name (what they should avoid). [Here](https://www.redhat.com/sysadmin/container-image-short-names) is some background.

.tekton/shellcheck-yaml-lint-pipeline.yaml
bnallapeta: > For each release of Pipeline-Service we should have a clear set of dependency versions: kcp, pac, argo cd, etc
Yes, agreed. This sort of a dependency matrix could be started once we have the first ever release.

Reg latest vs main : Yes, in the future when we have releases and multiple branches, latest may not necessarily mean latest. But for now, it is. Anyways, let me revert to 'main' to avoid any confusion.

bnallapeta:
Agreed. As discussed, have created #174 to track this. We can revisit this around the time of our release.

fgiloux (approved):
/lgtm

PR #175, opened 2022-07-29
Fix minor issues in local install scripts

Mo3m3n:
- Add kubectl and jq in prechecks
- Add information about exit status: normal exit(including ctrl-C) vs failure
- Fix wait_command: now program exits if command is not ready and
  timeout expires
- Fix kubectl-kcp plugin path

### DIFF
--- a/local/kcp/start.sh
+++ b/local/kcp/start.sh
@@ -53,14 +53,13 @@ kcp-start() {
   (cd "${TMP_DIR}" && exec "${KCP_DIR}/bin/kcp" start "${PARAMS[@]}") &> "${TMP_DIR}/kcp.log" &
   KCP_PID=$!
   KCP_PIDS+=(${KCP_PID})
+  wait_command "ls ${KUBECONFIG}" 30
   printf "KCP server started: %s\n" $KCP_PID
-
   touch "${TMP_DIR}/kcp-started"
 
-  wait_command "ls ${KUBECONFIG}" 30
   printf "Waiting for KCP to be ready ...\n"
   wait_command "kubectl --kubeconfig=${KUBECONFIG} get --raw /readyz" 30
-  printf "KCP ready: %s\n" $?
+  printf "KCP ready\n"
 }
 
 ingress-ctrler-start() {
@@ -90,7 +89,7 @@ envoy-start() {
 create-org() {
   printf "Creating organization\n"
   kubectl --kubeconfig="${KUBECONFIG}" config use-context root
-  KUBECONFIG="${KUBECONFIG}" kubectl kcp workspace use root
+  KUBECONFIG="${KUBECONFIG}" ${KCP_DIR}/bin/kubectl-kcp workspace use root
   kubectl --kubeconfig="${KUBECONFIG}" create -f "${PARENT_PATH}/pipelines-service-org.yaml"
 }
 
@@ -103,7 +102,6 @@ TMP_DIR="$(mktemp -d -t kcp-pipelines-service.XXXXXXXXX)"
 printf "Temporary directory created: %s\n" "${TMP_DIR}"
 
 KUBECONFIG="${TMP_DIR}/.kcp/admin.kubeconfig"
-printf "KUBECONFIG=%s\n" "${KUBECONFIG}"
 
 source "${PARENT_PATH}/../utils.sh"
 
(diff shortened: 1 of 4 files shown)


local/utils.sh
```diff
@@ -19,9 +19,17 @@ wait_command() {
   local wait_seconds="${1:-40}"; shift # 40 seconds as default timeout
 
   until [[ $((wait_seconds--)) -eq 0 ]] || eval "$command &> /dev/null" ; do sleep 1; done
+
+  ((++wait_seconds))
```
ramessesii2, line 23: I'm not sure if adding this line is helpful.

ramessesii2 (approved):
Thanks @Mo3m3n. Tested it locally, works fine for me.

local/utils.sh
fgiloux: +1

local/kcp/start.sh
```diff
@@ -123,6 +121,8 @@ if [[ -z "${PARAMS}" ]]; then
 fi
 
 kcp-start
+printf "KUBECONFIG=%s\n" "${KUBECONFIG}"
+printf "kubectl kcp plugin: %s\n" "${KCP_DIR}/bin/kubectl-kcp"
```
fgiloux, line 125: The standard way of using a kubectl plugin is to have it copied to the same location as the kubectl binary and then calling the kubectl binary rather than directly the plugin.
I think that we should instruct the user to copy the plugin binary to the same location as the kubectl binary.

fgiloux (commented):
Hi @Mo3m3n. Thanks a lot for this PR and helping with making the local experience better. We have proposed a few minor changes but the PR should be good to merge by the next iteration.

local/utils.sh
Mo3m3n: Without this line `wait_command` will always succeed even if the command fails after the timeout.
Which means for example the script [continue here](https://github.com/openshift-pipelines/pipelines-service/blob/43cdfd1c247b5538b68ab828a8695a6d13081eb6/local/kcp/start.sh#L61) even if kcp startup failed.
There are different ways to fix this, I have used the one from the [original kcp script](https://github.com/kcp-dev/kcp/blob/ac174c63a28feec6eaaa6ace6d63528b2333ad50/contrib/demos-unmaintained/demo/.startUtils#L9). Where if the counter `wait_seconds` is non-zero after the loop (command succeeded), the return status will be 0 (success). Otherwise the return status will be 1 (error).

fgiloux: but it should be ((--wait_seconds)), shouldn't it?
I did a quick test
~~~
set -o errexit
set -o nounset

wait_command() {
  local command="$1"; shift
  local wait_seconds="${1:-40}"; shift # 40 seconds as default timeout

  until [[ $((wait_seconds--)) -eq 0 ]] || eval "$command &> /dev/null" ; do sleep 1; done

  echo $wait_seconds
  ((--wait_seconds))
  ## does not work
  ## ((++wait_seconds))
}

echo "step 1"
wait_command "sleep 1s" 30
echo $?
echo "step 2"
wait_command "sleep 5s" 2
echo $?
echo "finished"
~~~

Mo3m3n: according to this loop:
```
  until [[ $((wait_seconds--)) -eq 0 ]] || eval "$command &> /dev/null" ; do sleep 1; done
```
We exit the loop when:
-  the command succeeds 
- or the timeout has expired and in this case we will always have **wait_seconds==-1**
So we need to `((++wait_seconds))` and send the exist status (0 failure, success otherwise).

If we `((--wait_seconds))` then this will always succeed even if the command is failing. For example you can try
```
wait_command "ls /non-existing-file" 3
echo finished
```

In your example:  the 2 timeout in `wait_command "sleep 5s" 2` is a particular case where it will always fail, you can try:
```
wait_command "ls /tmp" 3
echo finished
```
**NB**: `wait_seconds` is kind of a biased name, cause the wait is not in seconds in case the command takes more than a second to return a result and the `sleep` is a good example. So the param name makes sense only with commands taking less than a second to return a result

fgiloux: Thanks for the explanation. I did miss the point. Replacing step 2 with the following gives the expected result with ((++wait_seconds)): `wait_command "/tmp/test5/this_file_does_not_exist" 10`
The construct is a bit fragile if the command hangs like with sleep. Unlikely with `ls` a bit more with kubectl. I guess it is similar to watch in this respect.

local/kcp/start.sh
Mo3m3n: I have [updated](https://github.com/openshift-pipelines/pipelines-service/compare/43cdfd1c247b5538b68ab828a8695a6d13081eb6..774612422644696194afaa69b56b04cd43997ce8) the output message to reflect that.
Let me know if that's ok or we need to provide more information

PR #178, opened 2022-08-01
Specify minimum yq version

avinal:
- yq version must be >= 4.18.1, refer https://mikefarah.gitbook.io/yq/#notice-for-v4.x-versions-prior-to-4.18.1


fgiloux:
Hi @avinal thank you for opening this PR. I am not sure whether we want to specify the version in the readme. We don't have it for the other components: kcp-plugin, argo cli, jq. The link to the oc client points to an old version. I am concerned that we may not keep this information updated. Misleading information may be worth than not providing the information at all.
On the other hand we do want to have the versions for all the components documented, see Jira PLNSRVCE-461

avinal:
Hi Frederic, since this is a major failing point, we should mention this IMO. Also, it is a minimum version requirement so it won't need to be changed anytime soon, WDYT? I have already updated the dependencies doc in PLNSRVCE-461.

ramessesii2:
Do we still require this, now that #185 is merged? cc @avinal

avinal:
@ramessesii2 No, closing...

PR #179, opened 2022-08-03
removed precheck for kubectl cert-manager plugin

xinredhat:
Removed precheck for kubectl cert-manager plugin because the script doesn't rely on  kubectl cert-manager plugin


bnallapeta (approved):
We are not running any commands at the client end that requires the use of cert-manager plugin. Hence this PR to remove the
 check.

Issue #118, opened 2022-06-18
Improve UX of installation/setup scripts


avinal:
Currently, there are many tradeoffs and missing scenarios in the installation and setup scripts. This issue will track the overall progress of improving the user experience. Please update the list as needed.

- [ ] Clean the containers in case of failure
- [ ] Stop running kcp processes in case of failure
- [ ] Add parameter/environment variable to enable compilation of kcp binaries when not found

fgiloux:
Note: This is for local development environment.

adambkaplan:
Some other notes from my "local development" experiences:

1. kcp in a pod requests a full CPU. This often means that kcp can't be scheduled given that we install a lot of other components that are resource-heavy.
2. Our script simply stops when kcp fails to deploy, with no error message.

adambkaplan:
More notes:

- After overcoming `ckcp` install issues, I struggle with the "what's next" step. Our current README docs aren't exactly clear on this.
- The setup scripts default to laying out credentials for KCP and the compute cluster in a temporary directory. Don't developers need these to do any real work? Should we change the location of this data to be somewhere that is more local to the script?

Roming22:
@avinal A lot changed since you opened this issue, and many improvement were made. I'll mark this issue as stale.
Please put a comment if you still have pain points that need to be looked after.

avinal:
Sure @Roming22 I don't think this issue is relevant now. We should close it.

adambkaplan:
/close

Per @avinal's comment

PR #184, opened 2022-08-05
ckcp: Optimize resource requests and limits

adambkaplan:
kcp-in-container (ckcp) currently has CPU and memory requests that make
it difficult to schedule on a standard IPI cluster. In testing, the
following behavior was observed:

- Initial memory usage hits 177Mi, is never lower.
- kcp memory increases to ~500Mi once syncer is deployed.
- CPU usage rarely exceeds 0.1 CPU per minute. 100m CPU request is
typical for a system workload.

Partially addresses #118

### DIFF
--- a/ckcp/openshift/base/deployment.yaml
+++ b/ckcp/openshift/base/deployment.yaml
@@ -40,8 +40,8 @@ spec:
           limits:
             memory: 2Gi
           requests:
-            cpu: '1'
-            memory: 1Gi
+            cpu: 100m
+            memory: 178Mi
         volumeMounts:
         - name: kcp-certs
           mountPath: /etc/kcp/tls/server


adambkaplan:
Note - with this configuration update, ckcp can be deployed on a minimal AWS IPI-installed cluster (3 control plane nodes, 3 m6i.large worker nodes with 2 CPU, ~8Gi RAM)

ckcp/openshift/base/deployment.yaml
```diff
@@ -38,10 +38,11 @@ spec:
           value: ckcp-ckcp.apps.domain.org
         resources:
           limits:
+            cpu: 1000m
```
fgiloux, line 41: What is the purpose of the cpu limit?
Limit does not play any role in scheduling or resource reservation.
It could protect against a bad behavior of kcp but if kcp is not responding there is nothing left to protect.

fgiloux (commented):
Thanks Adam for looking at that. This PR will definitively make the resources for running the ckcp based environment more affordable.

ckcp/openshift/base/deployment.yaml
adambkaplan: I'll remove this. Since CPU is compressible, limits only throttle the process. This doesn't really matter for development.

fgiloux: Thanks!

PR #186, opened 2022-08-09
Fix tektonconfig patch

avinal:
- add job to patch the tektonconfig


avinal:
The URL is still pointed to my branch for testing purposes, once it is tested to work I will revert them.

gitops/argocd/tektoncd/patch-tekton-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+  name: patch-tekton-config-parameters
+  namespace: openshift-pipelines
+  annotations:
+    argocd.argoproj.io/hook: PostSync
+    argocd.argoproj.io/hook-delete-policy: HookSucceeded
+spec:
+  template:
+    spec:
+      containers:
+        - image: registry.redhat.io/openshift4/ose-cli:v4.9
```
fgiloux, line 13: The image should be 4.10

gitops/argocd/tektoncd/patch-tekton-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+            - -c
+            - |
+              echo "Waiting for TektonConfig config to be present"
+              until oc get TektonConfig config -n openshift-pipelines
+              do
+                sleep $SLEEP;
+              done
+
+              echo "Patching TektonConfig config patameters"
+              oc patch TektonConfig config --type='json' -p='[{"op":"remove","path":"/spec/pruner/keep"},{"op":"add","path":"/spec/pruner/keep-since","value":2880},{"op":"replace","path":"/spec/pruner/resources/0","value":"taskrun"},{"op":"replace","path":"/spec/pipeline/enable-tekton-oci-bundles","value":true}]'
```
fgiloux, line 25: This is imperative. We don't need Argo CD if we patch in an imperative way one of the few config files we have

fgiloux (changes requested):
I am concerned with the approach.
If we cannot have a change to a simple resource applied after it has been created we can as well get rid of Argo CD.

fgiloux:
@avinal As a side note: It should be possible to have a GitOps repository with overlays for testing that overwrite the required URLs so that they are pointing to your repository where required. In this way you can test your change without the need to change the URL in the resources. The scripts accept parameters.
If it does not work let us know as it would be something that we need to fix.

avinal:
Closing in favor of #229

PR #192, opened 2022-08-10
Upgrade kcp to release-0.7

Roming22:


Roming22:
I've tested it using ckcp, and used this version to deploy the 4.11 cluster.

I did not test the new version with local/kind.

bnallapeta:
@Roming22 I tried to test it locally.

4.10 OpenShift cluster
1. With a minor modification [here](https://github.com/Roming22/pipelines-service/blob/kcp-0.7/gitops/sre/environment/kcp/kustomization.yaml) from ref=docs to ref=release-0.7, I was able to get ckcp working
2. But unable to do any operation within my newly created workspace. No api-resources being listed too.

4.11 OpenShift cluster (I tested it on 4.11 because there was a comment that this was used to deploy in the 4.11 cluster)
Script fails at adding cluster to ArgoCD because in 4.11 which is k8s 1.24, secrets are not automatically created by service account. We need to do it manually. 
```
+ argocd cluster add default/api-ci-ln-cbydm6k-72292-origin-ci-int-gce-dev-rhcloud-com:6443/kube:admin --name=plnsvc --upsert --yes
INFO[0000] ServiceAccount "argocd-manager" already exists in namespace "kube-system" 
INFO[0001] ClusterRole "argocd-manager-role" updated    
INFO[0002] ClusterRoleBinding "argocd-manager-role-binding" updated 
FATA[0032] Failed to wait for service account secret: timed out waiting for the condition
```

fgiloux:
@Roming22 Mind that we now also have a reference in [the access-setup image](https://github.com/openshift-pipelines/pipeline-service/blob/main/.github/workflows/access-setup-image.yaml#L39)

bnallapeta:
@Roming22 I just realized that you don't have the latest copy. Kindly rebase and you should find DEPENDENCIES.md. We would need to update release-0.7 there as well.

Roming22:
@bnallapeta 
1. That seems to have been fixed by another PR in the meantime. There should not have been a reference to my repo.
2. I cannot reproduce with the latest commit. After the script has run, I can run `KUBECONFIG=./work/credentials/kubeconfig/kcp/ckcp-ckcp.default.pipeline-service-compute.kubeconfig kubectl api-resources`, and `kubectl apply -f https://raw.githubusercontent.com/tektoncd/pipeline/release-v0.35.x/examples/v1beta1/pipelineruns/pipelinerun-with-pipelinespec.yaml` will run the pipeline successfully.
3. the KCP pod takes a bit of time to boot up and be fully operational. The creation of the org workspace `root:default` is there to make sure that kcp is operational. The name has been chosen arbitrarily and is generic. We then register our compute in `root:default:pipeline-service-compute`. There's no specific reason for that and if you'd prefer it to be `root:pipeline-service-compute`, that would be fine with me. I'm not sure I understand why it matters though.

My comment about 4.11 was wrong, it was in fact deployed on 4.10, and we'll need to work on 4.11 compatibility at some point.

I've rebased, and all reference to 0.6 should now be gone.

DEPENDENCIES.md
```diff
@@ -34,7 +34,7 @@
 |--------------------|-------------------|-----------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | oc (OpenShift CLI) | 4.10              | To interact with the cluster                                                | Follows OpenShift version                                                                                                                                        |
 | kubectl            | 1.23.x            | To interact with the cluster                                                | Follows kubernetes version which follows OpenShift version. We only need either oc or kubectl                                                                    |
```
fgiloux, line 36: kubectl needs to be moved to 1.24 as kcp 0.7 is based on 1.24. OpenShift 4.10 is based on 1.23 but it is best to have the version mismatch (n-2 compatibility is guaranteed IIRC) in this direction not to miss any new feature.

fgiloux (changes requested):

DEPENDENCIES.md
fgiloux: see also: https://github.com/openshift-pipelines/pipeline-service/pull/193#discussion_r946322220

fgiloux (approved):
/lgtm although I have not tested it, especially the image builds. I will leave that and the merge to @bnallapeta

bnallapeta (approved):
- Tested image builds too. It contains the kcp-plugin 0.7.4 as expected.
- Had some issues with the 0.7.0 kcp plugin. Upgraded to 0.7.4 and that seems to have cleared the issue. So, it won't work with generic release-0.7 plugin.
- Considering the above point, I think we should update the DEPENDENCIES.md to reflect 0.7.4 for kcp plugin.
- Apart from this small point, the PR looks good.

PR #193, opened 2022-08-11
add ci for testing local dev environment

xinredhat:
**Changes**
Added a CI for testing local dev environment

### DIFF
--- a/.github/workflows/local-dev-ci.yaml
+++ b/.github/workflows/local-dev-ci.yaml
@@ -1,0 +1,58 @@
+name: Execute local test
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+    paths-ignore:
+      - "docs/**"
+      - "**/*.md"
+
+jobs:
+  local-dev-test:
+    name: local-dev-test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: v1.17
+
+      # Install kind with a local registry
+      - uses: container-tools/kind-action@v1
+        name: Kubernetes kind Cluster w/local registry
+        with:
+          kubectl_version: v1.23.3
+
+      - name: Checkout kcp repository
+        uses: actions/checkout@v3
+        with:
+          repository: kcp-dev/kcp
+          ref: release-0.6
+      - name: Build kcp binaries
+        run: |-
+          make install
+          kubectl-kcp --version
+
+      - name: Checkout pipeline-service repository
+        uses: actions/checkout@v3
+      - name: Set up Kubernetes workload clusters
+        run: |-
+          ./local/kind/setup.sh
+
+      - name: Set up a kcp cluster.
+        run: |-
+          nohup ./local/kcp/start.sh > nohup.out &
+          i=0
+          while ! grep "ctrl-C" nohup.out >/dev/null; do
+            i=$((i+1))
+            if [ $i -gt 70 ]; then
+              printf "Failed to set up a kcp cluster\n"
+              cat nohup.out
+              exit 1
+            fi
+            sleep 5
+          done
+          cat nohup.out


.github/workflows/local-dev-ci.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+      - uses: azure/setup-kubectl@v3
+        with:
+          version: v1.23.3
+        
+      - name: Checkout kcp repository
+        uses: actions/checkout@v3
+        with:
+          repository: kcp-dev/kcp
+          ref: release-0.6
```
Roming22, line 31: You should probably bump that version to 0.7

.github/workflows/local-dev-ci.yaml
```diff
@@ -0,0 +1,60 @@
...
+jobs:
+  local-dev-test:
+    name: local-dev-test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: v1.17
+
+      - uses: azure/setup-kubectl@v3
```
fgiloux, line 23: out of curiosity why setup-kubectl rather than kubectl-simple or kube-tools?

.github/workflows/local-dev-ci.yaml
```diff
@@ -0,0 +1,60 @@
...
+          version: v1.23.3
+        
+      - name: Checkout kcp repository
+        uses: actions/checkout@v3
+        with:
+          repository: kcp-dev/kcp
+          ref: release-0.6
+      - name: Build KCP plugin
+        run: |-
+          go mod vendor
```
fgiloux, line 34: What is the benefit of running this command here?

.github/workflows/local-dev-ci.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - uses: azure/setup-kubectl@v3
+        with:
+          version: v1.23.3
+        
+      - name: Checkout kcp repository
+        uses: actions/checkout@v3
+        with:
+          repository: kcp-dev/kcp
+          ref: release-0.6
+      - name: Build KCP plugin
```
fgiloux, line 32: It is building all the kcp binaries not only the kcp plugin: kcp needs to be written with lowercase.

.github/workflows/local-dev-ci.yaml
fgiloux: also looking further down, isn't [kind-action already bringing kubectl](https://github.com/container-tools/kind-action/blob/main/kind.sh#L149-L154)?

.github/workflows/local-dev-ci.yaml
xinredhat: When PR #192 is merged, I will change that version to 0.7

.github/workflows/local-dev-ci.yaml
xinredhat: you are right, kind-action already bring kubectl. fixed

.github/workflows/local-dev-ci.yaml
xinredhat: fixed it by removing it

.github/workflows/local-dev-ci.yaml
xinredhat: fixed it. it was change to `name: Build kcp binaries`

.github/workflows/local-dev-ci.yaml
fgiloux: Note: this is not only kcp version that will need to be updated but also kubectl 1.24.x and go v1.18 versions

fgiloux (approved):

.github/workflows/local-dev-ci.yaml
```diff
@@ -0,0 +1,58 @@
...
+          while ! grep "ctrl-C" nohup.out >/dev/null; do
+            sleep 5
+            i=$((i+1))
+            if [ $i -gt 70 ]; then
+              printf "Failed to set up a kcp cluster\n"
+              cat nohup.out
+              exit 1
+            fi
+          done
+          cat nohup.out
```
avinal, line 58:
```suggestion
          cat nohup.out
          
```
Please add a terminating newline.

.github/workflows/local-dev-ci.yaml
```diff
@@ -0,0 +1,58 @@
...
+    name: local-dev-test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: v1.17
+
+      # Install kind with a local registry
+      - uses: container-tools/kind-action@v1
+        name: Kubernetes KinD Cluster w/local registry
```
avinal, line 25:
```suggestion
        name: Kubernetes kind Cluster w/local registry
```
Nothing much important, but if there is no specific meaning of `KinD` then please change.

.github/workflows/local-dev-ci.yaml
xinredhat: fixed.

.github/workflows/local-dev-ci.yaml
xinredhat: fxied.

.github/workflows/local-dev-ci.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Set up Kubernetes workload clusters
+        run: |-
+          ./local/kind/setup.sh
+
+      - name: Set up a kcp cluster.
+        run: |-
+          nohup ./local/kcp/start.sh > nohup.out &
+          i=0
+          while ! grep "ctrl-C" nohup.out >/dev/null; do
+            sleep 5
```
fgiloux, line 50: ideally the sleep 5 would be at the end of the loop. In the last run there is no point of waiting for 5s more if it will fail any way.

xinredhat: done

PR #196, opened 2022-08-15
fixed issue of  extracting SA token

xinredhat:
**Issue**
The secrets for the pac-manager SA tokens are created asynchronously. For now we need to wait for their creation before trying to extract the token form them.

**Changes**
Wait until pac-manager SA token is created before trying to extract the token form them.

### DIFF
--- a/images/access-setup/content/bin/common.sh
+++ b/images/access-setup/content/bin/common.sh
@@ -37,8 +37,21 @@ get_context() {
     printf "[ERROR] Install jq\n" >&2
     exit 1
   fi
-  token_secret="$(KUBECONFIG="$target" kubectl get sa "$sa" -n "$namespace" -o json |
-    jq -r '.secrets[].name | select(. | test(".*token.*"))')"
+
+  for i in {1..5}
+  do
+    token_secret="$(KUBECONFIG="$target" kubectl get sa "$sa" -n "$namespace" -o json |
+      jq -r '.secrets[]?.name | select(. | test(".*token.*"))')"
+    if [ -n "$token_secret" ]; then
+      break
+    fi
+    sleep 5
+  done
+  if [ -z "$token_secret" ]; then
+      printf "Failed to extract secret token\n"
+      exit 1
+  fi
+
   current_cluster="$(KUBECONFIG="$target" kubectl config view \
     -o jsonpath="{.contexts[?(@.name==\"$current_context\")].context.cluster}")"
 


images/access-setup/content/bin/common.sh
```diff
@@ -37,8 +37,16 @@ get_context() {
...
     exit 1
   fi
-  token_secret="$(KUBECONFIG="$target" kubectl get sa "$sa" -n "$namespace" -o json |
-    jq -r '.secrets[].name | select(. | test(".*token.*"))')"
+
+  token_secret=""
+  while [ -z "$token_secret" ]; do
+    token_secret="$(KUBECONFIG="$target" kubectl get sa "$sa" -n "$namespace" -o json |
+    jq -r '.secrets[]?.name | select(. | test(".*token.*"))')"
+    if [ -n "$token_secret" ]; then
```
Roming22, line 45: I believe that bit should not be necessary, since on the next while loop the end condition should be met.

fgiloux (approved):
Apart from @Roming22's comment this looks good to me

images/access-setup/content/bin/common.sh
xinredhat: so we just use `sleep 5` to wait a while, right?

fgiloux: "sleep 1" or "sleep 2" could be an enhancement but it is unrelated to Romain's point.
Romain's point was to remove:
~~~
    if [ -n "$token_secret" ]; then
      break
    fi
~~~
as with each iteration `[ -z "$token_secret" ]`is evaluated (right after the above code) and if the variable has been populated the loop is stopped

xinredhat: ah, thank you for the explanation . fixed.

images/access-setup/content/bin/common.sh
```diff
@@ -37,8 +37,13 @@ get_context() {
...
     exit 1
   fi
-  token_secret="$(KUBECONFIG="$target" kubectl get sa "$sa" -n "$namespace" -o json |
-    jq -r '.secrets[].name | select(. | test(".*token.*"))')"
+
+  token_secret=""
+  while [ -z "$token_secret" ]; do
+    token_secret="$(KUBECONFIG="$target" kubectl get sa "$sa" -n "$namespace" -o json |
+    jq -r '.secrets[]?.name | select(. | test(".*token.*"))')"
+  done
```
ramessesii2, line 45: Shouldn't we have a timeout as well? In case, `token_secret` never gets populated.

fgiloux: yes good point

xinredhat: fixed

fgiloux: there should be a sleep 1 between the trials. It is in my opinion better than just relying on the latency of the API server.

images/access-setup/content/bin/common.sh
```diff
@@ -37,8 +37,19 @@ get_context() {
...
   fi
-  token_secret="$(KUBECONFIG="$target" kubectl get sa "$sa" -n "$namespace" -o json |
-    jq -r '.secrets[].name | select(. | test(".*token.*"))')"
+
+  token_secret=""
+  i=0
+  while [ -z "$token_secret" ]; do
+  	i=$((i+1))
+  	if [ $i -gt 10 ]; then
+      printf "Failed to extract secret token\n"
```
ramessesii2, line 46: Would it be a good idea to use this function [exit_error()](https://github.com/openshift-pipelines/pipeline-service/blob/main/images/access-setup/content/bin/common.sh#L17)

xinredhat: > Would it be a good idea to use this function [exit_error()](https://github.com/openshift-pipelines/pipeline-service/blob/main/images/access-setup/content/bin/common.sh#L17)

@ramessesii2 the function [exit_error()](https://github.com/openshift-pipelines/pipeline-service/blob/main/images/access-setup/content/bin/common.sh#L17)  has a function `usage` inside it, maybe is not good to use it.

ramessesii2: ohh, missed that. thanks

ramessesii2 (approved):

images/access-setup/content/bin/common.sh
xinredhat: discussed with @fgiloux ,  it should not take more than a couple of seconds and we should not let wait the user for more than 1 minute. we decided to be sleep 5 and loop 5 times. 
Fixed

fgiloux: sorry Xin I was not really clear. The sleep 5s should only happen when it is needed. If the first call is successful we should not make the process wait.
I would propose to do a bit of refactoring, something like:
~~~
for i in {1..5}
do
    token_secret="$(KUBECONFIG="$target" kubectl get sa "$sa" -n "$namespace" -o json |
      jq -r '.secrets[]?.name | select(. | test(".*token.*"))')"
    if [ -n "$token_secret" ]; then
       break
    fi
    sleep 5
done
if [ -z "$token_secret" ]; then
    printf "Failed to extract secret token\n"
    exit 1
fi
~~~

xinredhat: thank you for your suggestion, your logic look more clear. I directly copied your codes. Fixed.

fgiloux: old logic needs to be trimmed

xinredhat: done.

PR #198, opened 2022-08-16
add owners file for openshift-ci to manage merge process

xinredhat:
**Changes:**

Added `OWNERS` file to the repo so that openshift-ci can manage merge process for us, i.e. merging automatically after review&approval

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,13 @@
+reviewers:
+  - Roming22 
+  - fgiloux
+  - ramessesii2
+  - bnallapeta
+  - Mo3m3n
+  - xinredhat
+
+approvers:
+  - Roming22
+  - fgiloux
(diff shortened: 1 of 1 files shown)


OWNERS
```diff
@@ -0,0 +1,11 @@
+approvers:
+- Roming22 
+- fgiloux
+- ramessesii2
+- bnallapeta
+
+approvers:
```
fgiloux, line 7: Why two lists of approvers? Shouldn't the second list be reviewers? In that case the list could be extended to Xin and Moemen

xinredhat: fixed. And updated the lists of approvers and reviewers according to @bnallapeta 's suggestion

bnallapeta (approved):

Issue #174, opened 2022-07-28
Upload pipeline-service images to multiple repos


bnallapeta:
- Currently, we are storing our images in quay.io.
- Short names for container images comes with the potential risk of an outsider uploading an image with the same name in another image registry which could be a potential security threat at the user's end.
- If we don't claim redhat-pipeline-service on ghcr.io and docker (other image registry?), someone could claim it, push tainted images there, and an unsuspecting user doing a pull docker redhat-pipeline-service/setup_compute:latest could open themselves to harm.
- In order to avoid this, we need to upload our images to multiple image registries.
- https://www.redhat.com/sysadmin/container-image-short-names
- https://www.redhat.com/en/blog/be-careful-when-pulling-images-short-name

Roming22:
@ramessesii2 We've decided against uploading our images to other registries, but we've secured access to some organizations on various image registries to prevent that type of squatting.

Do you think that's good enough to close this issue?

bnallapeta:
@Roming22 @ramessesii2 Yes, we have created orgs in other registries and that is good enough for now. Good to close this one.

PR #202, opened 2022-08-18
Push image to ghcr.io

Roming22:
Images are still being pushed to quay.io

### DIFF
--- a/.github/workflows/shellcheck-image.yaml
+++ b/.github/workflows/shellcheck-image.yaml
@@ -25,7 +25,7 @@ jobs:
         id: build-image
         uses: redhat-actions/buildah-build@v2
         with:
-          image: shellcheck
+          image: test-shellcheck
           context: ./test/images/shellcheck
           tags: latest ${{ steps.vars.outputs.sha_short }} ${{ github.ref_name }}
           containerfiles: |
@@ -39,5 +39,14 @@ jobs:
           registry: quay.io/redhat-pipeline-service
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_TOKEN }}
+      - name: Push to ghcr.io
+        id: push-to-ghcr
+        uses: redhat-actions/push-to-registry@v2
+        with:
+          image: ${{ steps.build-image.outputs.image }}
+          tags: ${{ steps.build-image.outputs.tags }} ${{ github.ref_name }}
+          registry: ghcr.io/${{ github.repository_owner }}
+          username: ${{ github.actor }}
+          password: ${{ github.token }}
       - name: Print image url
         run: echo "Image pushed to ${{ steps.push-to-quay.outputs.registry-paths }}"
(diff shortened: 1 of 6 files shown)


.github/workflows/shellcheck-image.yaml
```diff
@@ -39,5 +39,14 @@ jobs:
           registry: quay.io/redhat-pipeline-service
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_TOKEN }}
+      - name: Push to ghcr.io
+        id: push-to-ghcr
+        uses: redhat-actions/push-to-registry@v2
+        with:
+          image: ${{ steps.build-image.outputs.image }}
+          tags: ${{ steps.build-image.outputs.tags }} ${{ github.ref_name }}
+          registry: ghcr.io/${{ github.repository_owner }}
```
xinredhat, line 48: shellcheck image is for testing, shall we expect those images to be under something like test/images?

Roming22: ghcr is an image registry so `images` seems a bit redundant, but I've added `test` to the path.

.github/workflows/shellcheck-image.yaml
```diff
@@ -39,5 +39,14 @@ jobs:
           registry: quay.io/redhat-pipeline-service
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_TOKEN }}
+      - name: Push to ghcr.io
+        id: push-to-ghcr
+        uses: redhat-actions/push-to-registry@v2
+        with:
+          image: ${{ steps.build-image.outputs.image }}
+          tags: ${{ steps.build-image.outputs.tags }} ${{ github.ref_name }}
+          registry: ghcr.io/${{ github.repository_owner }}/test
```
fgiloux, line 48: Having  the images in different paths in different registries, defeat some of the purpose of having them in multiple registries. In the past using sub-paths for storing images was not supported. I am not sure if that is still the case with some of the registries. If it is I see two options:
- we say that we don't ship test images and as such we store them only in ghcr.io
- we create a second organization: ${{ github.repository_owner }}-test or something similar

In this respect having different names for the organizations in quay.io and ghcr.io defeats the purpose of having the images in multiple registries.

fgiloux (changes requested):

.github/workflows/shellcheck-image.yaml
Roming22: How about I prefix test images with `test-` and leave them at the root of the registry?

Roming22: I implemented the change, let me know if you're not happy with the solution.

fgiloux: I understand that it is not directly related to this PR but I am still concerned with 
ghcr.io/openshift-pipelines vs quay.io/redhat-pipeline-service
If users do `docker pull openshift-pipelines/cluster-setup:latest` and they don't have ghcr.io configured in their registry list best case they don't get the image, worst case they get someone's else image, possibly from an attacker.

fgiloux: I thought that it is what we wanted to address by publishing to multiple registries.

bnallapeta: +1 to @fgiloux 's comment. I created this #174 a while back to track this.

Roming22: I also agree on the need to push to docker. That would be a separate PR though.

PR #204, opened 2022-08-23
PLNSRVCE-522 : Containers - Clair - Vulnerability Scanner

bnallapeta:
- Added the image (Dockerfile + script) for future consumption in Pipeline Runs.
- For now, created a Github Action to use the same script to fetch the results to check for vulnerabilities in our images on quay.
- Have also setup Authorization token via Github secrets.

### DIFF
--- a/.github/workflows/individual-image-scanner-quay.yaml
+++ b/.github/workflows/individual-image-scanner-quay.yaml
@@ -1,0 +1,186 @@
+---
+name: Fetch security scan results from quay
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  scans:
+    runs-on: ubuntu-latest
+    outputs:
+      access-setup-output: ${{ steps.access-setup-scan.outputs.VULNERABILITIES_EXIST }}
+      cluster-setup-output: ${{ steps.cluster-setup-scan.outputs.VULNERABILITIES_EXIST }}
+      gateway-deployment-output: ${{ steps.gateway-deployment-scan.outputs.VULNERABILITIES_EXIST }}
+      kcp-registrar-output: ${{ steps.kcp-registrar-scan.outputs.VULNERABILITIES_EXIST }}
+      shellcheck-output: ${{ steps.shellcheck-scan.outputs.VULNERABILITIES_EXIST }}
+      vulnerability-scan-output: ${{ steps.vulnerability-scan.outputs.VULNERABILITIES_EXIST }}
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            access-setup:
+            - 'images/access-setup/**'
+            cluster-setup:
+              - 'images/cluster-setup/**'
+            gateway-deployment:
+              - 'images/gateway-deployment/**'
+            kcp-registrar:
+              - 'images/kcp-registrar/**'
+            shellcheck:
+              - 'test/images/shellcheck/**'
+            vulnerability:
+              - 'test/images/vulnerability-scan/**'
+
+      - name: access-setup scan
+        continue-on-error: true
+        id: access-setup-scan
+        if: steps.filter.outputs.access-setup == 'true'
+        run: |
+          ./test/images/clair-scan/scan-image.sh | tee /tmp/clair-scan.log
+          echo "::set-output name=VULNERABILITIES_EXIST::$(tail -1 /tmp/clair-scan.log)"
+        env:
+          AUTH_BEARER_TOKEN: ${{ secrets.AUTH_BEARER_TOKEN }}
+          IMAGE_NAME: access-setup
+
+      - name: cluster-setup scan
+        continue-on-error: true
+        id: cluster-setup-scan
+        if: steps.filter.outputs.cluster-setup == 'true'
+        run: |
+          ./test/images/clair-scan/scan-image.sh | tee /tmp/clair-scan.log
+          echo "::set-output name=VULNERABILITIES_EXIST::$(tail -1 /tmp/clair-scan.log)"
+        env:
+          AUTH_BEARER_TOKEN: ${{ secrets.AUTH_BEARER_TOKEN }}
+          IMAGE_NAME: cluster-setup
+
+      - name: gateway-deployment scan
+        continue-on-error: true
+        id: gateway-deployment-scan
+        if: steps.filter.outputs.gateway-deployment == 'true'
+        run: |
+          ./test/images/clair-scan/scan-image.sh | tee /tmp/clair-scan.log
+          echo "::set-output name=VULNERABILITIES_EXIST::$(tail -1 /tmp/clair-scan.log)"
+        env:
+          AUTH_BEARER_TOKEN: ${{ secrets.AUTH_BEARER_TOKEN }}
+          IMAGE_NAME: gateway-deployment
+
+      - name: kcp-registrar scan
+        continue-on-error: true
+        id: kcp-registrar-scan
+        if: steps.filter.outputs.kcp-registrar == 'true'
+        run: |
+          ./test/images/clair-scan/scan-image.sh | tee /tmp/clair-scan.log
+          echo "::set-output name=VULNERABILITIES_EXIST::$(tail -1 /tmp/clair-scan.log)"
+        env:
+          AUTH_BEARER_TOKEN: ${{ secrets.AUTH_BEARER_TOKEN }}
+          IMAGE_NAME: kcp-registrar
+
+      - name: shellcheck scan
+        continue-on-error: true
+        id: shellcheck-scan
+        if: steps.filter.outputs.shellcheck == 'true'
+        run: |
+          ./test/images/clair-scan/scan-image.sh | tee /tmp/clair-scan.log
+          echo "::set-output name=VULNERABILITIES_EXIST::$(tail -1 /tmp/clair-scan.log)"
+        env:
+          AUTH_BEARER_TOKEN: ${{ secrets.AUTH_BEARER_TOKEN }}
+          IMAGE_NAME: shellcheck
+
+      - name: vulnerability scan
+        continue-on-error: true
+        id: vulnerability-scan
+        if: steps.filter.outputs.shellcheck == 'true'
+        run: |
+          ./test/images/clair-scan/scan-image.sh | tee /tmp/clair-scan.log
+          echo "::set-output name=VULNERABILITIES_EXIST::$(tail -1 /tmp/clair-scan.log)"
+        env:
+          AUTH_BEARER_TOKEN: ${{ secrets.AUTH_BEARER_TOKEN }}
+          IMAGE_NAME: vulnerability-scan
+
+  check-results:
+    runs-on: ubuntu-latest
+    needs: scans
+    if: always()
+    steps:
+      - name: Check access setup results
+        id: check-access-setup-results
+        if: always()
+        run: |
+          res=${{ needs.scans.outputs.access-setup-output }}
+          res=${res:=0}
+          if [[ $res != 0 ]]; then
+            echo "Vulnerabilities found with access-setup image. Please check scans job for more details."
+            exit 1
+          else
+            echo "No vulnerabilities found"
+          fi
+
+      - name: Check cluster setup results
+        id: check-cluster-setup-results
+        if: always()
+        run: |
+          res=${{ needs.scans.outputs.cluster-setup-output }}
+          res=${res:=0}
+          if [[ $res != 0 ]]; then
+            echo "Vulnerabilities found with cluster-setup image. Please check scans job for more details."
+            exit 1
+          else
+            echo "No vulnerabilities found"
+          fi
+
+      - name: Check gateway deployment results
+        id: check-gateway-deployment-results
+        if: always()
+        run: |
+          res=${{ needs.scans.outputs.gateway-deployment-output }}
+          res=${res:=0}
+          if [[ $res != 0 ]]; then
+            echo "Vulnerabilities found with gateway-deployment image. Please check scans job for more details."
+            exit 1
+          else
+            echo "No vulnerabilities found"
+          fi
+
+      - name: Check kcp registrar results
+        id: check-kcp-registrar-results
+        if: always()
+        run: |
+          res=${{ needs.scans.outputs.kcp-registrar-output }}
+          res=${res:=0}
+          if [[ $res != 0 ]]; then
+            echo "Vulnerabilities found with kcp-registrar image. Please check scans job for more details."
+            exit 1
+          else
+            echo "No vulnerabilities found"
+          fi
+
+      - name: Check shellcheck results
+        id: check-shellcheck-results
+        if: always()
+        run: |
+          res=${{ needs.scans.outputs.shellcheck-output }}
+          res=${res:=0}
+          if [[ $res != 0 ]]; then
+            echo "Vulnerabilities found with shellcheck image. Please check scans job for more details."
+            exit 1
+          else
+            echo "No vulnerabilities found"
+          fi
+
+      - name: Check vulnerability-scan results
+        id: check-vulnerability-scan-results
+        if: always()
+        run: |
+          res=${{ needs.scans.outputs.vulnerability-scan-output }}
+          res=${res:=0}
+          if [[ $res != 0 ]]; then
+            echo "Vulnerabilities found with vulnerability-scan image. Please check scans job for more details."
+            exit 1
+          else
+            echo "No vulnerabilities found"
+          fi
--- a/test/images/vulnerability-scan/Dockerfile
+++ b/test/images/vulnerability-scan/Dockerfile
@@ -1,0 +1,27 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
+LABEL build-date= \
+      com.redhat.build-host= \
+      description="This image provides a script to scan Clair for security vulnerabilities on container images via Quay." \
+      distribution-scope="public" \
+      io.k8s.description="This image provides a script to scan Clair for security vulnerabilities on container images via Quay." \
+      io.k8s.display-name="clair scan" \
+      maintainer="Pipeline Service" \
+      name="clair-scan" \
+      release="0.1" \
+      summary="Provides the latest results of clair security scan." \
+      url="https://github.com/openshift-pipelines/pipeline-service/tree/main/test/images/vulnerability-scan" \
+      vcs-ref=  \
+      vcs-type="git" \
+      vendor="Pipeline Service" \
+      version="0.1"
+WORKDIR /
+RUN mkdir /workspace && chmod 777 /workspace && chown 65532:65532 /workspace
+ENV HOME /tmp/home
+RUN mkdir $HOME && chmod 777 $HOME && chown 65532:65532 $HOME
+RUN microdnf install curl
+COPY ./scan.sh /usr/local/bin/scan.sh
+RUN chmod 755 /usr/local/bin/scan.sh
+USER 65532:65532
+VOLUME /workspace
+WORKDIR /workspace
+ENTRYPOINT ["/usr/local/bin/scan.sh"]
--- a/test/images/vulnerability-scan/scan-image.sh
+++ b/test/images/vulnerability-scan/scan-image.sh
@@ -1,0 +1,85 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 The Pipeline Service Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+
+VULNERABILITIES_EXIST=1
+
+get_digest() {
+  #number of retries=3
+  for i in {1..3}
+  do
+    #Get the digest of the image
+    digest_http_code=$(curl -s -w '%{http_code}' -o /tmp/digest.json -H "Accept: application/vnd.quay.distribution.manifest.v2+json Content-type: application/json Authorization: Bearer $AUTH_BEARER_TOKEN" https://quay.io/api/v1/repository/redhat-pipeline-service/"$IMAGE_NAME")
+
+    if [[ "$digest_http_code" == "200" ]]; then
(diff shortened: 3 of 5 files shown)


.github/workflows/vulnerabilites-scanner-quay.yaml
```diff
@@ -0,0 +1,33 @@
...
+          echo "::set-output name=VULNERABILITIES_EXIST::$VULNERABILITIES_EXIST"
+        env:
+          AUTH_BEARER_TOKEN: ${{ secrets.AUTH_BEARER_TOKEN }}
+
+      - name: Check results
+        id: check-results
+        if: ${{ steps.fetch-results.outputs.VULNERABILITIES_EXIST }} == "true"
+        run: |
+          echo "Vulnerabilities found"
+          exit 1
```
avinal, line 33:
```suggestion
          exit 1

```

test/images/vulnerability-scan/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN mkdir /workspace && chmod 777 /workspace && chown 65532:65532 /workspace
+ENV HOME /tmp/home
+RUN mkdir $HOME && chmod 777 $HOME && chown 65532:65532 $HOME
+RUN microdnf install curl
+COPY ./scan.sh /usr/local/bin/scan.sh
+RUN chmod 755 /usr/local/bin/scan.sh
+USER 65532:65532
+VOLUME /workspace
+WORKDIR /workspace
+ENTRYPOINT ["/usr/local/bin/scan.sh"]
```
avinal, line 27:
```suggestion
ENTRYPOINT ["/usr/local/bin/scan.sh"]

```

test/images/vulnerability-scan/scan.sh
```diff
@@ -0,0 +1,82 @@
...
+
+main() {
+  for img  in "${images[@]}"; do
+      get_digest
+  done
+}
+
+if [ "${BASH_SOURCE[0]}" == "$0" ]; then
+  main "$@"
+fi
```
avinal, line 82:
```suggestion
fi

```

avinal (changes requested):
Whitespace after # in Bash/YAML comments is considered a good practice.

test/images/vulnerability-scan/scan.sh
```diff
@@ -0,0 +1,82 @@
...
+ "kcp-registrar"
+ "shellcheck"
+)
+
+get_digest() {
+  #number of retries=3
+  for _ in {1..3}
+  do
+    #Get the digest of the image
+    digest_http_code=$(curl -s -w '%{http_code}' -o /tmp/digest.json -H "Accept: application/vnd.quay.distribution.manifest.v2+json Content-type: application/json Authorization: Bearer           $AUTH_BEARER_TOKEN " https://quay.io/api/v1/repository/redhat-pipeline-service/"$img")
```
ramessesii2, line 34:
```suggestion
    digest_http_code=$(curl -s -w '%{http_code}' -o /tmp/digest.json -H "Accept: application/vnd.quay.distribution.manifest.v2+json Content-type: application/json Authorization: Bearer $AUTH_BEARER_TOKEN " https://quay.io/api/v1/repository/redhat-pipeline-service/"$img")
```
I'm not sure if those spaces were intended.

ramessesii2 (commented):
It looks good to me.

test/images/vulnerability-scan/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+      com.redhat.build-host= \
+      description="This image provides a script to scan Clair for security vulnerabilities on container images via Quay." \
+      distribution-scope="public" \
+      io.k8s.description="This image provides a script to scan Clair for security vulnerabilities on container images via Quay." \
+      io.k8s.display-name="clair scan" \
+      maintainer="Pipeline Service" \
+      name="clair-scan" \
+      release="0.1" \
+      summary="Provides the latest results of clair security scan." \
+      url="https://github.com/openshift-pipelines/pipeline-service/tree/main/test/images/clair-scan" \
```
fgiloux, line 12: Shouldn't the URL be ending with test/images/vulnerability-scan?

test/images/vulnerability-scan/scan.sh
```diff
@@ -0,0 +1,82 @@
...
+    rm -f /tmp/vulnerability.json
+    return
+  else
+    printf "Error while fetching vulnerability report from Quay. Status code: %s\n" "${results_http_code}"
+  fi
+}
+
+main() {
+  for img  in "${images[@]}"; do
+      get_digest
```
fgiloux, line 76: it may be a bit nicer not to rely on global variables in the function, passing arguments:
call: get_digest $img
retrieving it in the function with $1

test/images/vulnerability-scan/scan.sh
```diff
@@ -0,0 +1,82 @@
...
+
+images=(
+ "access-setup"
+ "argocd-registrar"
+ "cluster-setup"
+ "gateway-deployment"
+ "kcp-registrar"
+ "shellcheck"
+)
+
```
fgiloux, line 28: This means that we fetch the scan of every images when a PR modifies any of them.
Technically an image scan may return new vulnerabilities when the image itself has not been modified (due to new CVEs being added to Clair's vulnerability database) but I feel that it is not helpful in case of a pull request.
I would propose the following:
- to have a GitHub action that retrieves image scans only for images that have been modified by the PR, so that PR authors don't get asked to fix a vulnerability in an image they have not touched
- to have a periodic GitHub action that once a day retrieves the scan results for all the images.

.github/workflows/vulnerabilites-scanner-quay.yaml
```diff
@@ -0,0 +1,34 @@
...
+      packages: write
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Fetch results
+        id: fetch-results
+        run: |
+          ./test/images/clair-scan/scan.sh
+          echo "::set-output name=VULNERABILITIES_EXIST::$VULNERABILITIES_EXIST"
```
Roming22, line 25: Because you're executing scan.sh, I don't expect `$VULNERABILITIES_EXIST` to be set once you exit the script. I would change the script to exit with a non-0 code when something went wrong and use that instead.

.github/workflows/vulnerabilites-scanner-quay.yaml
```diff
@@ -0,0 +1,34 @@
...
+        env:
+          AUTH_BEARER_TOKEN: ${{ secrets.AUTH_BEARER_TOKEN }}
+
+      - name: Check results
+        id: check-results
+        if: ${{ steps.fetch-results.outputs.VULNERABILITIES_EXIST }} == "true"
```
Roming22, line 31: The check would be more robust if the check was `!= false`, because it would fail by default.

test/images/vulnerability-scan/scan.sh
```diff
@@ -0,0 +1,82 @@
...
+    #Get the digest of the image
+    digest_http_code=$(curl -s -w '%{http_code}' -o /tmp/digest.json -H "Accept: application/vnd.quay.distribution.manifest.v2+json Content-type: application/json Authorization: Bearer $AUTH_BEARER_TOKEN" https://quay.io/api/v1/repository/redhat-pipeline-service/"$img")
+
+    if [[ "$digest_http_code" == "200" ]]; then
+      digest=$(jq '.tags[] | select(.name == "main").manifest_digest ' /tmp/digest.json | tr -d '"')
+      rm -f /tmp/digest.json
+      get_vulnerabilities
+      break
+    else
+      printf "Error while fetching digests from Quay. Status code: %s\n" "${digest_http_code}"
```
Roming22, line 42: Related to the other comment about failing by default. If we can't get any digest, the var would not be set, and the check would go through even though the images were not scanned.

I believe we should fail if any image can't be scanned.

test/images/vulnerability-scan/scan.sh
```diff
@@ -0,0 +1,82 @@
...
+    printf " \t %s results " "$img"
+    printf "\n *********************************** \n"
+
+    #Filtering results with vulnerabilities only.
+    results=$(jq -r '.data.Layer.Features[].Vulnerabilities[]' /tmp/vulnerability.json)
+
+    if [[ -n $results ]]; then
+      printf "%s" "$results" | jq
+      printf "\n Check the full report here: https://quay.io/repository/redhat-pipeline-service/%s/manifest/%s?tab=vulnerabilities \n" "$img" "$digest"
+      export VULNERABILITIES_EXIST=true
```
Roming22, line 62: As hinted earlier, I'd use a `STATUS_CODE` set to 1 by default, that would be changed to 0 only when no vulnerability has been found and all images have been processed and exit with that status code.

test/images/vulnerability-scan/scan.sh
bnallapeta: I have developed two GitHub actions now.
periodic-scanner-quay.yaml - Runs everyday at 00.00.00
individual-image-scanner-quay.yaml - Runs with every PR. But only scans for those images that are modified in the PR.

bnallapeta:
@fgiloux @Roming22 I have processed all the comments. PTAL.

.github/workflows/individual-image-scanner-quay.yaml
```diff
@@ -0,0 +1,159 @@
+---
+name: Fetch security scan results from quay
+on:
+  push:
+    branches:
+      - clair
+  workflow_dispatch:
```
avinal, line 7: There must be a `pull_request` trigger here according to [this comment](https://github.com/openshift-pipelines/pipeline-service/pull/204#r954583381). We will be changing the branches to `main` right?

bnallapeta: ah, clair was a typo. Was testing on that and forgot to update to main. Done.
The link is not redirecting to any comment. But, I think we want to run the vulnerability check once the PR is merged. Not for a Pull Request. Happy to hear other's thoughts on this as well.

avinal: I think that too, once PR is merged then only check should start.

fgiloux: Images are built on push not on PR. I am open to have this behavior changed but that would be a prerequisite. Bharath, I haven't looked at the details but how do you ensure that you are retrieving the scan result of the image that has been built after merge of the PR and not the previous one, i.e that the actions are sequential or that you wait till the new result is available?

avinal: I think adding `concurrency` to the GitHub Actions might help. See https://docs.github.com/en/actions/using-jobs/using-concurrency

fgiloux: Concurrency is one thing but I am not too worried with having multiple PRs getting merged at the same time. We don't have that many volumes. The thing is that jjobs need to happen sequentially. [workflow_run](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) may help but may still not be enough. IIRC clair scanning is done async.

bnallapeta: For now, we don't have that level of granular control. We could use workflow_run or workflow_dispatch but I would prefer to do this once we unify 6 different GitHub actions we currently have for building images into one using paths_filter that I have used here. 
That way, we could trigger scan action once build action is completed. WDYT?

p.s.: I think we might be able to control this better when we have PipelineRuns instead of actions.

fgiloux: Yes I am fine either ways. Until we move to PaC what is important to me is that we lower the noise ratio:
- when a PR is created we get the CVEs for +/- the changes that have been introduced by it
- we don't need a subsequent PR to figure out CVEs introduced by a previous PR

bnallapeta: Ack. If there are no other changes, then how about merging this PR?
Next actions:
1. Merge individual actions into one single action for building and uploading images.
2. Introduce workflow_run/workflow_dispatch to ensure sync between image uploads and scan results.

test/images/vulnerability-scan/scan-image.sh
```diff
@@ -0,0 +1,74 @@
...
+    digest_http_code=$(curl -s -w '%{http_code}' -o /tmp/digest.json -H "Accept: application/vnd.quay.distribution.manifest.v2+json Content-type: application/json Authorization: Bearer $AUTH_BEARER_TOKEN" https://quay.io/api/v1/repository/redhat-pipeline-service/"$IMAGE_NAME")
+
+    if [[ "$digest_http_code" == "200" ]]; then
+      digest=$(jq '.tags[] | select(.name == "main").manifest_digest ' /tmp/digest.json | tr -d '"')
+      rm -f /tmp/digest.json
+      get_vulnerabilities
+      break
+    else
+      printf "Error while fetching digests from Quay. Status code: %s\n" "${digest_http_code}"
+    fi
```
fgiloux, line 36: What do you think about adding a sleep before retry?

test/images/vulnerability-scan/scan-image.sh
```diff
@@ -0,0 +1,74 @@
...
+  #number of retries=3
+  for _ in {1..3}
+  do
+    #Get the digest of the image
+    digest_http_code=$(curl -s -w '%{http_code}' -o /tmp/digest.json -H "Accept: application/vnd.quay.distribution.manifest.v2+json Content-type: application/json Authorization: Bearer $AUTH_BEARER_TOKEN" https://quay.io/api/v1/repository/redhat-pipeline-service/"$IMAGE_NAME")
+
+    if [[ "$digest_http_code" == "200" ]]; then
+      digest=$(jq '.tags[] | select(.name == "main").manifest_digest ' /tmp/digest.json | tr -d '"')
+      rm -f /tmp/digest.json
+      get_vulnerabilities
```
fgiloux, line 32: From a code structure point of view wouldn't it make more sense that the orchestration is done from main?
meaning that the code could exit if it was not possible to get the digest and get_vulnerabilities would then get called from main.

test/images/vulnerability-scan/scan-image.sh
```diff
@@ -0,0 +1,74 @@
...
+
+    if [[ -n $results ]]; then
+      printf "%s" "$results" | jq
+      printf "\n Check the full report here: https://quay.io/repository/redhat-pipeline-service/%s/manifest/%s?tab=vulnerabilities \n" "$IMAGE_NAME" "$digest"
+    else
+      printf "No vulnerabilities found \n"
+      VULNERABILITIES_EXIST=0
+    fi
+
+    rm -f /tmp/vulnerability.json
```
fgiloux, line 60: Just wondering whether we need that. I don't see where it would play a role later on in the script and my understanding is that there is no persistent storage mounted, meaning that it would get deleted when the container exits anyway.

test/images/vulnerability-scan/scan.sh
```diff
@@ -0,0 +1,81 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+
+images=( "access-setup" "argocd-registrar" "cluster-setup" "gateway-deployment" "kcp-registrar" "shellcheck" )
+images_vul=("${images[@]/*/1}")
+
+get_digest() {
```
fgiloux, line 23: Couldn't we just share the same function with what is used for single image scan?

test/images/vulnerability-scan/scan.sh
```diff
@@ -0,0 +1,81 @@
...
+      rm -f /tmp/digest.json
+      get_vulnerabilities "$img" "$img_vul"
+      break
+    else
+      printf "Error while fetching digests from Quay. Status code: %s\n" "${digest_http_code}"
+    fi
+  done
+}
+
+get_vulnerabilities() {
```
fgiloux, line 43: same as above

test/images/vulnerability-scan/scan-image.sh
fgiloux: It may be different if the cod is shared with what is used for multiple scans.

test/images/vulnerability-scan/scan.sh
bnallapeta: Hmm. It would require a lot of modification but yes, it could be made possible. 
But, even in the future when we write PipelineRuns for this, I was thinking we would have two different images under test /images - one for individual scan and one for periodic scan - meaning two Dockerfiles and respective scripts.

test/images/vulnerability-scan/scan-image.sh
bnallapeta: Sure, I just added that to keep it clean. I don't think it matters either way because like you said, the container will die once its done. Deleted the line.

test/images/vulnerability-scan/scan-image.sh
bnallapeta: GitHub actions doesn't play nice with exit codes. Would lead to increase in complexity between jobs and steps.

Initially I had written the logic in main but introduced a function to keep it modular. And we don't want to call get_vulnerabilities unless we have a 200 from digest.

test/images/vulnerability-scan/scan-image.sh
bnallapeta: I thought the default time curl takes and retires to try and fetch was enough. But yeah, don't mind adding more time to reduce the chance of failure. 
Added sleep 30.

test/images/vulnerability-scan/scan-image.sh
fgiloux: exit code would have made the handling less verbose but if it is not good with github action it does not prevent you to use return and break the loop making main aware that the normal flow should be interrupted and the error flow should be executed.

test/images/vulnerability-scan/scan-image.sh
fgiloux: there could be some improvement here to avoid unnecessary waits, something like:
~~~
for i in {1..3}
...
else
  printf "Error while fetching digests from Quay. Status code: %s\n" "${digest_http_code}"
  if i < 3
    sleep 30
  fi
~~~

test/images/vulnerability-scan/scan.sh
fgiloux: Fine you can have multiple images that import some common scripts or functions, possibly in a separate file.

.github/workflows/individual-image-scanner-quay.yaml
```diff
@@ -0,0 +1,159 @@
...
+            shellcheck:
+              - 'test/images/shellcheck/**'
+
+      - name: access-setup scans
+        continue-on-error: true
+        id: access-setup-scan
+        if: steps.filter.outputs.access-setup == 'true'
+        run: |
+          ./test/images/clair-scan/scan-image.sh
+          echo "::set-output name=VULNERABILITIES_EXIST::$(./test/images/clair-scan/scan-image.sh | tail -1)"
```
Roming22, line 42: Aren't you running the scan-image.sh script twice? 
If that's the case, I suggest:
```
./test/images/clair-scan/scan-image.sh | tee /tmp/clair-scan.log
echo "::set-output name=VULNERABILITIES_EXIST::$(tail -1 /tmp/clair-scan.log)"
```
That way you still have the output in the job for the user, and can access them in the job without needing to re-run the script.

.github/workflows/individual-image-scanner-quay.yaml
```diff
@@ -0,0 +1,159 @@
...
+            cluster-setup:
+              - 'images/cluster-setup/**'
+            gateway-deployment:
+              - 'images/gateway-deployment/**'
+            kcp-registrar:
+              - 'images/kcp-registrar/**'
+            shellcheck:
+              - 'test/images/shellcheck/**'
+
+      - name: access-setup scans
```
Roming22, line 36: `scans` -> `clair-scan` or `scan`. I don't think we're doing multiple scans.

.github/workflows/individual-image-scanner-quay.yaml
```diff
@@ -0,0 +1,159 @@
...
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  scans:
+    runs-on: ubuntu-latest
+    outputs:
+      as-output: ${{ steps.access-setup-scan.outputs.VULNERABILITIES_EXIST }}
```
Roming22, line 13: Do not use abreviations. It will quickly become confusing.

.github/workflows/individual-image-scanner-quay.yaml
```diff
@@ -0,0 +1,159 @@
...
+        id: access-setup-scan
+        if: steps.filter.outputs.access-setup == 'true'
+        run: |
+          ./test/images/clair-scan/scan-image.sh
+          echo "::set-output name=VULNERABILITIES_EXIST::$(./test/images/clair-scan/scan-image.sh | tail -1)"
+        env:
+          AUTH_BEARER_TOKEN: ${{ secrets.AUTH_BEARER_TOKEN }}
+          IMAGE_NAME: access-setup
+
+      - name: cluster-setup scans
```
Roming22, line 47: The same step(s) repeats over and over. I would think there's a way to create a generic step. I'm fine if you tackle this as a latter improvement.

.github/workflows/individual-image-scanner-quay.yaml
```diff
@@ -0,0 +1,159 @@
...
+            access-setup:
+            - 'images/access-setup/**'
+            cluster-setup:
+              - 'images/cluster-setup/**'
+            gateway-deployment:
+              - 'images/gateway-deployment/**'
+            kcp-registrar:
+              - 'images/kcp-registrar/**'
+            shellcheck:
+              - 'test/images/shellcheck/**'
```
Roming22, line 34: Seems like you forgot to scan the vulnerability-scan image itself.

test/images/vulnerability-scan/scan-image.sh
bnallapeta: Done. Improved the log message as well.

test/images/vulnerability-scan/scan-image.sh
bnallapeta: Fixed.

test/images/vulnerability-scan/scan.sh
bnallapeta: Ack. Will work on this in a subsequent PR.

.github/workflows/individual-image-scanner-quay.yaml
bnallapeta: Fixed.

.github/workflows/individual-image-scanner-quay.yaml
bnallapeta: Fixed.

.github/workflows/individual-image-scanner-quay.yaml
bnallapeta: Yes, I understand there is repetition, but we need to create a new step for every image as we are going to run that step only if the files in the respective image's path is modified. And also, I am passing unique IMAGE_NAME env variable for each step which is translated into a container.

.github/workflows/individual-image-scanner-quay.yaml
bnallapeta: yes, I missed that. Added it now.

Roming22:
@fgiloux Please merge when you/re ready to approve the PR.

avinal (approved):

PR #207, opened 2022-08-25
added  Hadolint configuration file

xinredhat:
**Changes:**

Added Hadolint configuration file for configuring rules. The CI job is implemented with Openshift CI, this is the [PR](https://github.com/openshift/release/pull/31602)

### DIFF
--- a/test/config/hadolint.yaml
+++ b/test/config/hadolint.yaml
@@ -1,0 +1,14 @@
+---
+# Hadolint configuration file
+
+# configure ignore rules
+# see https://github.com/hadolint/hadolint#rules for a list of available rules.
+
+ignored: []
+failure-threshold: info
+format: tty
+no-color: false
+no-fail: false
+trustedRegistries:
+  - quay.io
+  - registry.access.redhat.com


test/config/hadolint.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+# configure ignore rules
+# see https://github.com/hadolint/hadolint#rules for a list of available rules.
+
+ignored: []
+failure-threshold: info
+format: tty
+no-color: false
+no-fail: false
+trustedRegistries:
```
bnallapeta, line 12: Currently, we are using just quay.io. Until we work on creating 'redhat-pipeline-service' org on other registries and upload our images there, I'm not confident to trust other registries. If registry.redhat.io is being used in the Openshift CI platform, then we could keep that as well.

bnallapeta:
@xinredhat Also, I see just the config file. What about the GitHub actions for hadolint? Or have you set this up in openshift-ci? If so, please update that link in this PR's description.

xinredhat:
> Or have you set this up in

@bnallapeta yes, as they helped fixed the openshift CI issue. I will continue using Openshift CI for Hadolint check. This is the CI PR implemented with Openshift CI.
https://github.com/openshift/release/pull/31602

test/config/hadolint.yaml
xinredhat: ok, I will delete other trusted registries.

xinredhat: @bnallapeta because our Dockerfile files are using `registry.access.redhat.com/ubi8/ubi-minimal:8.6` as base image. I also need to add it as one of trusted registries.

xinredhat: fixed

bnallapeta (approved):
/lgtm

PR #210, opened 2022-08-26
added config file for SAST tools Checkov

xinredhat:
**Changes:**

Added config file for SAST tools Checkov. The Checkov [CI PR](https://github.com/openshift/release/pull/31716) relies on this PR.

You could reference [CLI Command Reference](https://www.checkov.io/2.Basics/CLI%20Command%20Reference.html) for the the properties in config file.

**Verify:**
You can execute the command with its Docker image to see the result.
```
$ docker run -it --rm --volume <the path of pipelines-service directory>:/tf --workdir /tf bridgecrew/checkov  --directory /tf --config-file /tf/test/config/checkov.yaml
```

### DIFF
--- a/test/config/checkov.yaml
+++ b/test/config/checkov.yaml
@@ -1,0 +1,15 @@
+---
+check: []
+compact: false
+evaluate-variables: true
+framework:
+  - kubernetes
+  - dockerfile
+  - github_actions
+  - kustomize
+skip-download: false
+output: cli
+quiet: true
+skip-check: []
+skip-fixes: true
+soft-fail: true


Roming22 (approved):

test/config/checkov.yaml
```diff
@@ -0,0 +1,20 @@
...
+compact: false
+evaluate-variables: true
+external-checks-dir:
+  - docs
+  - exploration/argocd-external/docs
+framework:
+  - kubernetes
+  - dockerfile
+  - github_actions
+  - gitlab_ci
```
fgiloux, line 12: Why do we need gitlab_ci?

test/config/checkov.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - kustomize
+  - argo_workflows
```
fgiloux, line 14: I don't think we use argo workflows, do we?

xinredhat: you are right.  We don't have the below  argo workflows.  Fixed.

```
apiVersion: argoproj.io/v1alpha1
kind: Workflow
```

test/config/checkov.yaml
xinredhat: fixed.

bnallapeta (approved):
LGTM. @fgiloux if your earlier comments are addressed, please go ahead and merge the PR.

PR #211, opened 2022-08-26
bugfix: ckcp/test.sh

Roming22:
A change in kcp broke the function retrieving the namespace
on the compute cluster. This change fixes the issue and
adds a test to ensure that the scripts fails nicely if the
issue was to happen again.

### DIFF
--- a/ckcp/test.sh
+++ b/ckcp/test.sh
@@ -1,28 +1,37 @@
 #!/usr/bin/env bash
 
 #quit if exit status of any cmd is a non-zero value
-set -exuo pipefail
+set -euo pipefail
+
+# Uncomment the below line to enable debugging
+# set -x
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" >/dev/null ; pwd)"
 KUBECONFIG="${KUBECONFIG:-$HOME/.kube/config}"
 export KUBECONFIG=$KUBECONFIG
-KUBECONFIG_KCP="${KUBECONFIG_KCP:-$SCRIPT_DIR/work/kubeconfig/admin.kubeconfig}"
+KUBECONFIG_KCP="${KUBECONFIG_KCP:-$SCRIPT_DIR/work/credentials/kubeconfig/kcp/ckcp-ckcp.default.pipeline-service-compute.kubeconfig}"
 
 get_namespace() {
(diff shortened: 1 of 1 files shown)


ckcp/test.sh
```diff
@@ -11,18 +12,24 @@ KUBECONFIG_KCP="${KUBECONFIG_KCP:-$SCRIPT_DIR/work/kubeconfig/admin.kubeconfig}"
 get_namespace() {
   # Retrieve the KCP namespace id
   local ns_locator
-  ns_locator="{\"logical-cluster\":\"$(
+  ns_locator="\"workspace\":\"$(
     KUBECONFIG="$KUBECONFIG_KCP" kubectl kcp workspace current | cut -d\" -f2
```
ramessesii2, line 16: > KUBECONFIG_KCP="${KUBECONFIG_KCP:-$SCRIPT_DIR/work/kubeconfig/admin.kubeconfig}"

I think we'll need to change `$KUBECONFIG_KCP` to point to the kcp kubeconfig inside `$WORK_DIR/credentials/kubeconfig/kcp/-`, now generated every-time `openshift_dev_setup.sh` is run.

Roming22: There should be a way to pass the work dir as an argument when running the script.
I think that the whole script should be refactored to have a similar structure to the other scripts. That's something for another PR.

ckcp/test.sh
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 
 #quit if exit status of any cmd is a non-zero value
-set -exuo pipefail
+set -euo pipefail
+# set -x
```
bnallapeta, line 5: Suggestion. Adding a comment above this would be helpful.
#Uncomment the below line to enable debugging.

bnallapeta (approved):

ckcp/test.sh
Roming22: I think I kinda missed your point. I've fixed the default path.

ckcp/test.sh
Roming22: Done

Roming22:
@bnallapeta Tests finally passed, ready to merge if you are OK with the changes.

PR #220, opened 2022-09-01
Check for local binaries

drpaneas:
Make sure to check for binaries used by the script, beforehand, if it's not present fail as early as possible. For example in our case we didn't know `yq` is supposed to be part of the container image:

e.g. https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/31857/rehearse-31857-pull-ci-redhat-appstudio-managed-gitops-main-managed-gitops-e2e-tests-on-kcp/1565006131774885888#1:build-log.txt%3A37

### DIFF
--- a/ckcp/openshift_dev_setup.sh
+++ b/ckcp/openshift_dev_setup.sh
@@ -81,8 +81,6 @@ precheck_binary() {
 }
 
 init() {
-  precheck_binary "kubectl" "yq" "curl"
-
   APP_LIST=(
             "openshift-gitops"
             "cert-manager"
@@ -384,6 +382,7 @@ check_cr_sync() {
 
 main() {
   parse_args "$@"
+  precheck_binary "kubectl" "yq" "curl" "argocd"
   init
   precheck
   check_cluster_role
--- a/ckcp/openshift_dev_setup.sh
+++ b/ckcp/openshift_dev_setup.sh
@@ -70,7 +70,19 @@ parse_args() {
   done
 }
 
+# Checks if a binary is present on the local system
+precheck_binary() {
+  for binary in "$@"; do
+    command -v "$binary" >/dev/null 2>&1 || {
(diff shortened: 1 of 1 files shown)


ckcp/openshift_dev_setup.sh
```diff
@@ -70,7 +70,19 @@ parse_args() {
...
+  for binary in "$@"; do
+    command -v "$binary" >/dev/null 2>&1 || {
+      echo >&2 "openshift_dev_setup.sh requires '$binary' command-line utility to be installed on your local machine. Aborting..."
+      exit 1
+    }
+  done
+}
+
 init() {
+  precheck_binary "kubectl" "yq" "curl"
```
ramessesii2, line 84: I would suggest to add `argodcd CLI` as well to this list - https://argo-cd.readthedocs.io/en/stable/cli_installation/. We're using it in this script for e.g.- to register clusters to Argo CD.

ramessesii2: Also, would it be a good idea to move this function call to `main()`  since it's checking for binaries? init() is initializing different configs and variables for setting up ckcp and so `precheck_binary` could be called before init(). wdyt?

ramessesii2 (commented):
Thank you @drpaneas for adding this check. Overall, looks good to me, just few comments.

ckcp/openshift_dev_setup.sh
drpaneas: yes makes sense, thx for the feedback :D :D

drpaneas: I've just pushed again, PTAL and let me know if it's better now ;)

ramessesii2 (approved):
LGTM

ckcp/openshift_dev_setup.sh
ramessesii2: Thanks

Roming22 (approved):

PR #223, opened 2022-09-02
add debug switch for debugging ckcp error

xinredhat:
**Changes:**

Recently we often saw ckcp failed at waiting ckcp up.  So added a debug switch to print more information for debugging the failure when it fails.

### DIFF
--- a/ckcp/openshift_dev_setup.sh
+++ b/ckcp/openshift_dev_setup.sh
@@ -275,9 +275,20 @@ patches:
         value: $ckcp_route " >>"$ckcp_temp_dir/kustomization.yaml"
 
   echo -n "  - kcp $kcp_version: "
-  kubectl apply -k "$ckcp_temp_dir" >/dev/null 2>&1
+  kubectl apply -k "$ckcp_temp_dir" >/dev/null
   # Check if ckcp pod status is Ready
-  kubectl wait --for=condition=Ready -n $ns pod -l=app=kcp-in-a-pod --timeout=90s >/dev/null
+  local error="false"
+  kubectl wait --for=condition=Ready -n "$ns" pod -l=app=kcp-in-a-pod --timeout=120s >/dev/null 2>&1 || error="true"
+  if [ "$error" == "true" ]; then
+    printf "Debug: print out pod state\n"
+    kubectl -n "$ns" describe pods
+
+    printf "Debug: print out pod logs\n"
+    kubectl -n "$ns" logs "$(kubectl -n "$ns" get pods -l=app=kcp-in-a-pod -o name)"
+
+    exit 1
+  fi
+
   # Clean up kustomize temp dir
   rm -rf "$ckcp_temp_dir"
 


ckcp/openshift_dev_setup.sh
```diff
@@ -275,9 +275,17 @@ patches:
...
 
   echo -n "  - kcp $kcp_version: "
-  kubectl apply -k "$ckcp_temp_dir" >/dev/null 2>&1
+  kubectl apply -k "$ckcp_temp_dir" >/dev/null
   # Check if ckcp pod status is Ready
-  kubectl wait --for=condition=Ready -n $ns pod -l=app=kcp-in-a-pod --timeout=90s >/dev/null
+  kubectl wait --for=condition=Ready -n $ns pod -l=app=kcp-in-a-pod --timeout=90s >/dev/null 2>&1 || debug="true"
+  if [ "$debug" == "true" ]; then
+    printf "Debug: print out pod state\n"
+    kubectl -n $ns get pod
```
fgiloux, line 283: please use describe to get pods related status and events
`kubectl -n $ns describe pods

ckcp/openshift_dev_setup.sh
```diff
@@ -275,9 +275,17 @@ patches:
...
+
+    printf "Debug: print out pod logs\n"
+    kubectl -n $ns logs "$(kubectl -n $ns get pod -l=app=kcp-in-a-pod -o name)"
+  fi
```
fgiloux, line 287: if debug="true" this is an error state and you should exit the script with a non-zero value.
you may want to rename debug to something more accurate like `error`

fgiloux (changes requested):

ckcp/openshift_dev_setup.sh
xinredhat: fixed

ckcp/openshift_dev_setup.sh
xinredhat: Fixed.

fgiloux (approved):

ckcp/openshift_dev_setup.sh
```diff
@@ -275,9 +275,19 @@ patches:
...
 
   echo -n "  - kcp $kcp_version: "
-  kubectl apply -k "$ckcp_temp_dir" >/dev/null 2>&1
+  kubectl apply -k "$ckcp_temp_dir" >/dev/null
   # Check if ckcp pod status is Ready
-  kubectl wait --for=condition=Ready -n $ns pod -l=app=kcp-in-a-pod --timeout=90s >/dev/null
+  kubectl wait --for=condition=Ready -n $ns pod -l=app=kcp-in-a-pod --timeout=90s >/dev/null 2>&1 || error="true"
+  if [ "$error" == "true" ]; then
+    printf "Debug: print out pod state\n"
+    kubectl -n $ns describe pods
```
Roming22, line 283: I'm surprised shellcheck did not raise an error about missing `"`. Can you please have a look?

xinredhat: I double checked that shellcheck did not raise an error about missing `"`

ramessesii2, line 283: Hey @xinredhat, is this intentional to try to describe all the pods in the `$ns` NS?

xinredhat: yes

xinredhat: Done.

bnallapeta: It is better if we still mention the name of the pod instead of describing all pods in that namespace (at the moment, there's only one pod but may change in the future).
we could use a local variable to get the name of the pod and then substitute it on line 281, 284 and 287

PR #224, opened 2022-09-02
Combine 6 different actions into one for building and pushing images

bnallapeta:
The PR does the following:
- Combines 6 different actions into 1 based Github Action.
- Makes use of dorny/paths-filter@v2 action.
- Triggers a build and push only if the files in the path of an image is modified.
- Deleted old Github actions.
- Will have a separate PR (if required) for argocd-registrar-image as it is not included in security scans actions too. Needs discussion.


avinal (commented):
I think it would be nice to put different tasks as different `jobs`, it would make it easier to go through the log if something breaks also doing this will enable running the workflows in parallel which is for now sequential.

Roming22:
I'm not sure I understand the reason for merging all the files into a single one.

In any case it will be made obsolete once we start dogfooding, which should start happening in the next couple months.

bnallapeta:
> I'm not sure I understand the reason for merging all the files into a single one.
> 
> In any case it will be made obsolete once we start dogfooding, which should start happening in the next couple months.

Please see this comment chain for the reason. It basically allows us to trigger scan only when build is done. At the moment, since we have 6 actions for each image build, it will be messy to maintain the chain of events. 
Yes, once we start using our own service, we can simply delete all actions :)

bnallapeta:
> I think it would be nice to put different tasks as different `jobs`, it would make it easier to go through the log if something breaks also doing this will enable running the workflows in parallel which is for now sequential.

We can still see the logs of individual steps in actions. See this - https://github.com/bnallapeta/pipeline-service/runs/8156337803?check_suite_focus=true
The dorny action I am using to implement path based filtering supports it at the step level.

PR #225, opened 2022-09-05
Improve ckcp deployment to avoid cert-manger error

xinredhat:
**Purpose**

When you first install cert-manager, it will take a few seconds before the cert-manager API is usable. This change is to improve ckcp deployment to avoid the issue that the cert-manager API is usable while deploying ckcp.

**Changes**
1. Added a retry loop to ensure ckcp resources are deployed to OCP successfully
2. Minor changes to print out error messages when it fails
3. Update `check_cert_manager` function to reuse `check_deployments` function for validating cert manger deployment

### DIFF
--- a/ckcp/openshift_dev_setup.sh
+++ b/ckcp/openshift_dev_setup.sh
@@ -11,6 +11,9 @@ SCRIPT_DIR="$(
 # shellcheck source=ckcp/hack/util/update-git-reference.sh
 source "$SCRIPT_DIR/hack/util/update-git-reference.sh"
 
+# shellcheck source=images/cluster-setup/bin/utils.sh
+source "$SCRIPT_DIR/../images/cluster-setup/bin/utils.sh"
+
 GITOPS_DIR="$(dirname "$SCRIPT_DIR")/gitops"
 CKCP_DIR="$(dirname "$SCRIPT_DIR")/ckcp"
 CONFIG="$CKCP_DIR/config.yaml"
@@ -195,35 +198,23 @@ install_openshift_gitops() {
 
   # Register the host cluster as pipeline-cluster
   local cluster_name="plnsvc"
-  echo -n "  - Register host cluster to Argo CD as '$cluster_name': "
+  echo "  - Register host cluster to ArgoCD as '$cluster_name': "
   if ! KUBECONFIG="$KUBECONFIG_MERGED" argocd cluster get "$cluster_name" >/dev/null 2>&1; then
     argocd cluster add "$(yq e ".current-context" <"$KUBECONFIG")" --name="$cluster_name" --upsert --yes >/dev/null
   fi
-  echo "OK"
+  echo "    OK"
 }
 
 install_cert_manager(){
   APP="cert-manager-operator"
-  echo -n "  - OpenShift-Cert-Manager: "
+  echo "  - OpenShift-Cert-Manager: "
   kubectl apply -f "$GITOPS_DIR/argocd/argo-apps/$APP.yaml" >/dev/null
   check_cert_manager
 }
 
 check_cert_manager() {
-  # perform a dry-run create of a cert-manager
-  # Certificate resource in order to verify that CRDs are installed and all the
-  # required webhooks are reachable by the K8S API server.
-  local i=0
-  until kubectl create -f "$CKCP_DIR/openshift/base/certs.yaml" --dry-run=client >/dev/null 2>&1; do
-    echo -n "."
-    sleep 5
-    i=$((i+1))
-    if [ "$i" == 12 ]; then
-      echo "[ERROR] Failed to deploy OpenShift Cert-Manager operator on the host cluster."
-      exit 1
-    fi
-  done
-  echo "OK"
+  certManagerDeployments=("cert-manager" "cert-manager-cainjector" "cert-manager-webhook")
+  check_deployments "openshift-cert-manager" "${certManagerDeployments[@]}"
 }
 
 install_ckcp() {
@@ -275,20 +266,18 @@ patches:
         value: $ckcp_route " >>"$ckcp_temp_dir/kustomization.yaml"
 
   echo -n "  - kcp $kcp_version: "
-  kubectl apply -k "$ckcp_temp_dir" >/dev/null
+  # Deploy ckcp until all resources are successfully appied to OCP cluster 
+  local i=0
+  while ! error_msg=$(kubectl apply -k "$ckcp_temp_dir" 2>&1 1>/dev/null); do
+    sleep 10
+    i=$((i+1))
+    if [ $i -gt 12 ]; then
+      printf "\n Failed to deploy ckcp \n"
+      exit_error "$error_msg"
+    fi
+  done
   # Check if ckcp pod status is Ready
-  local error="false"
-  kubectl wait --for=condition=Ready -n "$ns" pod -l=app=kcp-in-a-pod --timeout=120s >/dev/null 2>&1 || error="true"
-  if [ "$error" == "true" ]; then
-    printf "Debug: print out pod state\n"
-    kubectl -n "$ns" describe pods
-
-    printf "Debug: print out pod logs\n"
-    kubectl -n "$ns" logs "$(kubectl -n "$ns" get pods -l=app=kcp-in-a-pod -o name)"
-
-    exit 1
-  fi
-
+  kubectl wait --for=condition=Ready -n $ns pod -l=app=kcp-in-a-pod --timeout=90s >/dev/null
   # Clean up kustomize temp dir
   rm -rf "$ckcp_temp_dir"
 
@@ -341,7 +330,7 @@ install_pipeline_service() {
     --git-remote-ref "$GIT_REF"
 
   echo "  - Deploy compute:"
-  "$SCRIPT_DIR/../images/cluster-setup/install.sh" --workspace-dir "$WORK_DIR"
+  "$SCRIPT_DIR/../images/cluster-setup/bin/install.sh" --workspace-dir "$WORK_DIR"
 
   echo "  - Install Pipelines as Code:"
   # Passing dummy values to the parameters of the pac/setup.sh script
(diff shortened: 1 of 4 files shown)


ckcp/openshift_dev_setup.sh
```diff
@@ -211,18 +211,24 @@ install_cert_manager(){
...
     if [ "$i" == 12 ]; then
-      echo "[ERROR] Failed to deploy OpenShift Cert-Manager operator on the host cluster."
+      printf "[ERROR] The cert-manager CRDs are not yet installed on the Kubernetes API server.\n"
       exit 1
     fi
   done
+
+  local ns="openshift-cert-manager"
+  #a loop to check if the cert-manager deployments are Available and Ready
+  if ! kubectl wait --for=condition=Available=true --all deployments -n "$ns" --timeout=100s >/dev/null 2>&1; then
```
fgiloux, line 228: `>/dev/null 2>&1`
as discussed before this is not a good practice

ckcp/openshift_dev_setup.sh
```diff
@@ -211,18 +211,24 @@ install_cert_manager(){
...
+  # Certificate resource in order to verify that the cert-manager CRDs are installed on the Kubernetes API server
   local i=0
   until kubectl create -f "$CKCP_DIR/openshift/base/certs.yaml" --dry-run=client >/dev/null 2>&1; do
-    echo -n "."
+    printf "."
     sleep 5
     i=$((i+1))
     if [ "$i" == 12 ]; then
-      echo "[ERROR] Failed to deploy OpenShift Cert-Manager operator on the host cluster."
+      printf "[ERROR] The cert-manager CRDs are not yet installed on the Kubernetes API server.\n"
```
fgiloux, line 221: The previous message was more clearly stating that it is a failure.
Proposing: "Failed to deploy the cert-manager operator. CRDs are not available.\n"

ckcp/openshift_dev_setup.sh
```diff
@@ -211,18 +211,24 @@ install_cert_manager(){
 
 check_cert_manager() {
   # perform a dry-run create of a cert-manager
-  # Certificate resource in order to verify that CRDs are installed and all the
-  # required webhooks are reachable by the K8S API server.
+  # Certificate resource in order to verify that the cert-manager CRDs are installed on the Kubernetes API server
   local i=0
   until kubectl create -f "$CKCP_DIR/openshift/base/certs.yaml" --dry-run=client >/dev/null 2>&1; do
```
fgiloux, line 216: I thought it had been fixed, maybe in another inflight PR. `>/dev/null 2>&1;`is not a good practice

fgiloux: Also I find the check a bit awkward. Why not simply asking the API server whether the crd is available rather than trying to create a resource and seeing the creation failing?

fgiloux (changes requested):

ckcp/openshift_dev_setup.sh
xinredhat: 1. I find the check logic from [here](https://github.com/cert-manager/cert-manager/blob/release-1.9/pkg/util/cmapichecker/cmapichecker.go#L87) 
2. I wanna ensure all the resources defined in `$CKCP_DIR/openshift/base/certs.yaml` are ready, so I directly apply the `certs.yaml` file. I think all of the resources need time to be installed in the API server. 
3.with `--dry-run=client`, it doesn't need to a cleanup 

please correct me if my understanding is not correct.

fgiloux: 1. and 2. still awkward
3. if you ask the API whether the resource exists it does not need any cleanup either

Really it increases readability and really checks what is described in the comment.
There are potentially other reasons for the creation of a resource not being successful besides the availability of the API, e.g. the successful execution of admission webhooks and other policies (I don't know whether they are in use with cert-manager). I am ok for you to test them (if they are used) but then the comment needs to be changed.
Otherwise kubectl get crd and potentially checking that versions match seem a more straightforward approach.

ckcp/openshift_dev_setup.sh
```diff
@@ -275,20 +267,12 @@ patches:
...
-    printf "Debug: print out pod state\n"
-    kubectl -n "$ns" describe pods
-
-    printf "Debug: print out pod logs\n"
-    kubectl -n "$ns" logs "$(kubectl -n "$ns" get pods -l=app=kcp-in-a-pod -o name)"
-
-    exit 1
+  # Deploy ckcp until all resources are successfully appied to OCP cluster 
+  if ! timeout 120s bash -c "while ! kubectl apply -k $ckcp_temp_dir >/dev/null 2>&1; do printf '.'; sleep 10; done"; then
+    exit_error "ckcp deployment failed (timeout)"
```
fgiloux, line 272: Why do you think that it is here acceptable to  erase error messages?
I understand that there is a retry loop but I would like at least to see what resource creation failed and for what reason during the last iteration.
""ckcp deployment failed (timeout)" is not useful for troubleshooting

images/cluster-setup/Dockerfile
```diff
@@ -22,7 +22,8 @@ RUN KUBE_VERSION=v1.24.0 && \
     curl -L -o /usr/local/bin/kubectl "https://dl.k8s.io/release/$KUBE_VERSION/bin/linux/amd64/kubectl" && \
     chmod 755 /usr/local/bin/kubectl
 RUN microdnf install -y findutils-4.6.0 git-2.31.1 && microdnf clean all
-COPY ./install.sh /usr/local/bin/install.sh
+COPY bin /usr/local/bin
+RUN chmod 755 /usr/local/bin/*.sh
```
fgiloux, line 26: Could you please be specific and apply the chmod to the files you are adding?
I don't think there is any shell script in the base image but if one gets added I would not like this Dockerfile to change its access rights unknowingly.

images/cluster-setup/bin/utils.sh
```diff
@@ -0,0 +1,45 @@
...
+  local ns="$1"
+  shift
+  local deployments=("$@")
+
+  for deploy in "${deployments[@]}"; do
+    printf "    - %s: " "$deploy"
+
+    #a loop to check if the deployment exists
+    if ! timeout 300s bash -c "while ! kubectl get deployment/$deploy -n $ns >/dev/null 2>/dev/null; do printf '.'; sleep 10; done"; then
+      exit_error "$deploy not found (timeout)"
```
fgiloux, line 33: Why do you think that it is here acceptable to erase error messages?
I understand that there is a retry loop but I would like at least to see what is the error when the deployment is not found. Is it due to a networking issue, is it due to the availability of the  API or is the resource not available on the cluster?
"$deploy not found (timeout)" does not carry this information

images/cluster-setup/bin/utils.sh
```diff
@@ -0,0 +1,45 @@
...
+    else
+      printf "Exists"
+    fi
+
+    #a loop to check if the deployment is Available and Ready
+    if kubectl wait --for=condition=Available=true "deployment/$deploy" -n "$ns" --timeout=100s >/dev/null 2>/dev/null; then
```
fgiloux, line 39: Why do you think that it is here acceptable to erase error messages?

bnallapeta:
@xinredhat Apart from the comments mentioned from @fgiloux, the errors being redirected to /dev/null is something that has to change. So please remove ">/dev/null 2>&1" wherever applicable. Because, if we have this, then we won't have any messages printed onto the terminal for troubleshooting.

ckcp/openshift_dev_setup.sh
xinredhat: @fgiloux when the retry loop fails, we execute the command again, do you agree?
```
 if ! timeout 120s bash -c "while ! kubectl apply -k $ckcp_temp_dir >/dev/null 2>&1; do printf '.'; sleep 10; done"; then
    kubectl apply -k $ckcp_temp_dir >/dev/null
 fi
```

fgiloux: Yes that would be a good solution. You would still keep the exit_error if the command is not successful. A comment stating "final trial with error display this time" or something similar would also be welcome

images/cluster-setup/bin/utils.sh
xinredhat: fixed. Please review

ckcp/openshift_dev_setup.sh
xinredhat: when timeout is triggered, maybe there is a short period of time between the last failure and the timeout. So I didn't use the logic that I mentioned `when the retry loop fails, we execute the command again`. Probably it will get different result from the latest failure. I modified the logic which just prints the error message of the last failure.

images/cluster-setup/Dockerfile
xinredhat: fixed

images/cluster-setup/bin/utils.sh
xinredhat: fixed.

images/cluster-setup/bin/utils.sh
fgiloux: Wrong place. There is no loop here for a single deployment. For the line above you can just remove   >/dev/null 2>/dev/null;
The comment is also misleading it is wait not a loop

ckcp/openshift_dev_setup.sh
```diff
@@ -275,20 +266,18 @@ patches:
...
 
   echo -n "  - kcp $kcp_version: "
-  kubectl apply -k "$ckcp_temp_dir" >/dev/null
+  # Deploy ckcp until all resources are successfully appied to OCP cluster 
+  local i=0
+  while ! error_msg=$(kubectl apply -k "$ckcp_temp_dir" 2>&1 1>/dev/null); do
+    sleep 10
+    i=$((i+1))
+    if [ $i -gt 12 ]; then
+      printf "\n Failed to deploy ckcp \n"
```
fgiloux, line 275: Here there is a loop and you need to do the apply one more time without redirecting to /dev/null
(See my  comment below). Only comment and exit if the latest apply fails

fgiloux (changes requested):

ckcp/openshift_dev_setup.sh
fgiloux: sorry I missed that you were storing the output in error_msg. Ignore my previous comment.

xinredhat: discussed with @fgiloux , we can keep it this way

images/cluster-setup/bin/utils.sh
fgiloux: ok I was confused (possibly by the comment). The reason for the "if" and redirection to /dev/null is run the describe command, which is fine. Didn't we also ask for the logs previously?

xinredhat: fixed. It just redirect stdout to /dev/null. when it fail, it prints error message directly and could also get the logs in the else clause

fgiloux (approved):

PR #227, opened 2022-09-06
PLNSRVCE-420 : Implement Tekton results for openshift development

CathalOConnorRH:
Implement Tekton Results and Postgresql DB for openshift development

Signed-off-by: Cathal O'Connor [cathal@redhat.com](mailto:cathal@redhat.com)

### DIFF
--- a/ckcp/openshift_dev_setup.sh
+++ b/ckcp/openshift_dev_setup.sh
@@ -34,6 +34,14 @@ Optional arguments:
         Directory in which to create the gitops file structure.
         If the directory already exists, all content will be removed.
         By default a temporary directory will be created in $TMPDIR.
+    --tekton-results-database-user TEKTON_RESULTS_DATABASE_USER
+        Username for tekton results database.
+        Can be read from \$TEKTON_RESULTS_DATABASE_USER
+        Default: %s
+    --tekton-results-database-password TEKTON_RESULTS_DATABASE_PASSWORD
+        Password for tekton results database.
+        Can be read from \$TEKTON_RESULTS_DATABASE_PASSWORD
+        Default: %s
     -d, --debug
         Activate tracing/debug mode.
     -h, --help
@@ -52,6 +60,14 @@ parse_args() {
       WORK_DIR="$1"
       mkdir -p "$WORK_DIR"
       WORK_DIR="$(cd "$1" >/dev/null; pwd)"
+      ;;
+    --tekton-results-database-user)
+      shift
+      TEKTON_RESULTS_DATABASE_USER="$1"
+      ;;
+    --tekton-results-database-password)
+      shift
+      TEKTON_RESULTS_DATABASE_PASSWORD="$1"
       ;;
     -d | --debug)
       set -x
@@ -339,6 +355,13 @@ patches:
 }
 
 install_pipeline_service() {
+
+  TEKTON_RESULTS_DATABASE_USER="$(yq '.tekton_results_db.user' "$CONFIG")"
+  TEKTON_RESULTS_DATABASE_PASSWORD="$(yq '.tekton_results_db.password' "$CONFIG")"
+
+  TEKTON_RESULTS_DATABASE_USER=${TEKTON_RESULTS_DATABASE_USER:="tekton"}
+  TEKTON_RESULTS_DATABASE_PASSWORD=${TEKTON_RESULTS_DATABASE_PASSWORD:=$(openssl rand -base64 20)}
+
   echo "- Setup compute access:"
   "$PROJECT_DIR/images/access-setup/content/bin/setup_compute.sh" \
     ${DEBUG:+"$DEBUG"} \
@@ -346,7 +369,9 @@ install_pipeline_service() {
     --work-dir "$WORK_DIR" \
     --kustomization "$GIT_URL/gitops/compute/pac-manager?ref=$GIT_REF" \
     --git-remote-url "$GIT_URL" \
-    --git-remote-ref "$GIT_REF" 2>&1 |
+    --git-remote-ref "$GIT_REF" \
+    --tekton-results-database-user "$TEKTON_RESULTS_DATABASE_USER" \
+    --tekton-results-database-password "$TEKTON_RESULTS_DATABASE_PASSWORD" 2>&1 |
     indent 2
 
   echo "- Deploy compute:"
@@ -360,6 +385,10 @@ install_pipeline_service() {
   GITOPS_REPO="https://example.git.com/my/repo" GIT_TOKEN="placeholder_token" \
     WEBHOOK_SECRET="placeholder_webhook" \
     "$GITOPS_DIR/pac/setup.sh" | indent 4
+
+  echo -n "- Install tekton-results DB: "
+  kubectl apply -k "$CKCP_DIR/manifests/tekton-results-db/openshift" 2>&1 |
+  indent 4
 
 }
 
--- a/ckcp/test.sh
+++ b/ckcp/test.sh
@@ -147,7 +147,8 @@ test_pipelines() {
   BASE_URL="https://raw.githubusercontent.com/tektoncd/pipeline/v0.32.0"
   manifest="pipelineruns/using_context_variables.yaml"
   # change ubuntu image to ubi to avoid dockerhub registry pull limit
-  curl --fail --silent "$BASE_URL/examples/v1beta1/$manifest" | sed 's|ubuntu|registry.access.redhat.com/ubi8/ubi-minimal:latest|' | sed '/serviceAccountName/d' | KUBECONFIG="$KUBECONFIG_KCP" kubectl create -f -
+  PIPELINE_RUN=$(curl --fail --silent "$BASE_URL/examples/v1beta1/$manifest" | sed 's|ubuntu|registry.access.redhat.com/ubi8/ubi-minimal:latest|' | sed '/serviceAccountName/d' | KUBECONFIG="$KUBECONFIG_KCP" kubectl create -f -)
+  echo "$PIPELINE_RUN"
   
   KUBECONFIG="$KUBECONFIG_KCP" kubectl wait --for=condition=Succeeded  PipelineRun --all --timeout=60s >/dev/null
   echo "Print pipelines custom resources inside kcp"
@@ -156,6 +157,9 @@ test_pipelines() {
   
   KCP_NS_NAME="$(get_namespace)"
   kubectl get pods -n "$KCP_NS_NAME"
+
+  test_results
+
   echo
 }
 
@@ -178,6 +182,51 @@ test_triggers() {
   kill $SVC_FORWARD_PID
   sleep 20
   KUBECONFIG="$KUBECONFIG_KCP" kubectl get pipelineruns
+  echo
+}
+
+test_results() {
+  KCP_NS_NAME="$(get_namespace)"
+  if [[ $PIPELINE_RUN == *"created"* ]]; then
+    PIPELINE_RUN=$(echo "$PIPELINE_RUN" | grep -o -P '(?<=/).*(?= created)')
+  fi
+  echo "[verify_results]"
+  echo "Verify tekton-results has stored the results in the database"
+
+  # Prepare a custom Service Account that will be used for debugging purposes
+  if ! KUBECONFIG="$KUBECONFIG" kubectl get serviceaccount tekton-results-debug -n openshift-pipelines >/dev/null 2>&1; then
+    KUBECONFIG="$KUBECONFIG" kubectl create serviceaccount tekton-results-debug -n openshift-pipelines
+  fi
+  # Grant required privileges to the Service Account
+  if ! KUBECONFIG="$KUBECONFIG" kubectl get clusterrolebinding tekton-results-debug -n openshift-pipelines >/dev/null 2>&1; then
+    KUBECONFIG="$KUBECONFIG" kubectl create clusterrolebinding tekton-results-debug --clusterrole=tekton-results-readonly --serviceaccount=openshift-pipelines:tekton-results-debug
+  fi
+
+  # Proxies the remote Service to localhost.
+  KUBECONFIG="$KUBECONFIG" kubectl port-forward -n openshift-pipelines service/tekton-results-api-service 50051 >/dev/null & 
+  PORTFORWARD_PID=$!
+  echo $PORTFORWARD_PID
+  # download the API Server certificate locally and configure gRPC.
+  KUBECONFIG="$KUBECONFIG" kubectl get secrets tekton-results-tls -n openshift-pipelines --template='{{index .data "tls.crt"}}' | base64 -d > /tmp/results.crt
+  export GRPC_DEFAULT_SSL_ROOTS_FILE_PATH=/tmp/results.crt
+  
+  RESULT_UID=$(kubectl get pipelinerun "$PIPELINE_RUN" -n "$KCP_NS_NAME" -o yaml | yq .metadata.uid)
+  
+  # This is required to pass shellcheck due to the single quotes in the GetResult name parameter.
+  QUERY=("name: \"$KCP_NS_NAME/results/$RESULT_UID\"")
+  RECORD_CMD=("grpc_cli call --channel_creds_type=ssl --ssl_target=tekton-results-api-service.openshift-pipelines.svc.cluster.local --call_creds=access_token=$(kubectl get secrets -n openshift-pipelines -o jsonpath="{.items[?(@.metadata.annotations['kubernetes\.io/service-account\.name']=='tekton-results-debug')].data.token}"| cut -d ' ' -f 2 | base64 --decode) localhost:50051 tekton.results.v1alpha2.Results.GetResult '${QUERY[@]}'")
+  RECORD_RESULT=$(eval "${RECORD_CMD[@]}")
+
+  # kill backgrounded port forwarding process as it is no longer required. 
+  kill "$PORTFORWARD_PID"
+
+  if [[ $RECORD_RESULT == *$RESULT_UID* ]]; then
+    echo "OK"
+  else
+    echo "Failed"
+    echo "[ERROR] Unable to retrieve record $RESULT_UID from pipeline run $PIPELINE_RUN" >&2
+    exit 1
+  fi
   echo
 }
 
--- a/gitops/argocd/tekton-results/README.md
+++ b/gitops/argocd/tekton-results/README.md
@@ -1,0 +1,10 @@
+# Tekton Results
+
+## ***Description*** 
+This will install tekton results on cluster to gather Tekton Pipelinerun results for long term storage.
+
+Installation is based on the [installation instructions](https://github.com/tektoncd/results/blob/main/docs/install.md) from upstream tekton results
+
+In a ckcp install, a postgresql database is installed so that results can be stored on cluster otherwise tekton results will use a configurable external database such as Amazon's RDS. 
+
+More information can be found [here](https://github.com/tektoncd/results#readme)
--- a/images/access-setup/content/bin/setup_compute.sh
+++ b/images/access-setup/content/bin/setup_compute.sh
@@ -51,6 +51,14 @@ Optional arguments:
         Git repo's ref to be referenced to apply various customizations.
         Can be read from \$GIT_REF.
         Default: %s
+    --tekton-results-database-user TEKTON_RESULTS_DATABASE_USER
+        Username for tekton results database.
+        Can be read from \$TEKTON_RESULTS_DATABASE_USER
+        Default: %s
+    --tekton-results-database-password TEKTON_RESULTS_DATABASE_PASSWORD
+        Password for tekton results database.
+        Can be read from \$TEKTON_RESULTS_DATABASE_PASSWORD
+        Default: %s
     -w, --work-dir WORK_DIR
         Directory into which the credentials folder will be created.
     -d, --debug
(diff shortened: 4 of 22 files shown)


ckcp/tekton-results/openshift/configmap.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
Roming22, line 1: All yaml must start with `---`, otherwise yamllint will complain in the CI. Occurs multiple times in the PR.

ckcp/tekton-results/openshift/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
Roming22, line 1: Please move `ckcp/tekton-results` to `ckcp/manifests/tekton-results`

images/access-setup/content/bin/setup_compute.sh
```diff
@@ -176,6 +178,28 @@ tekton_chains_manifest(){
   printf "OK\n"
 }
 
+tekton_results_manifest(){
+  manifest="$manifests_dir/compute/tekton-results/tekton-results-secret.yaml"
+  if [ ! -e "$manifest" ]; then
+    manifests_tmp_dir="$(dirname "$manifest")/tmp"
+    mkdir -p "$manifests_tmp_dir"
+    read -p "Please enter database username: " database_user
+    read -p "Please enter database password: " database_password
```
Roming22, line 187: Add `-s` for secret. You probably need `-r` as well so that shellcheck does not complain.

ckcp/openshift_dev_setup.sh
```diff
@@ -378,6 +379,18 @@ register_compute() {
...
 }
 
+install_tekton_results(){
+  APP="tekton-results"
+  echo -n "  - tekton-results application: "
+
+  kubectl apply -k "tekton-results/openshift" >/dev/null 2>&1
+
+  # Generate postgres secret
+  kubectl create secret generic tekton-results-database --namespace="openshift-pipelines" --from-literal=DATABASE_USER=postgres --from-literal=DATABASE_PASSWORD=$(openssl rand -base64 20) --dry-run=client -o yaml | kubectl apply -f - >/dev/null 2>&1
```
Roming22, line 389: I would expect `"$WORK_DIR/manifests/compute/tekton-results/tekton-results-secret.yaml"` to be created for the user at that point.
It would then be applied by `setup_compute.sh`.

images/access-setup/content/bin/setup_compute.sh
```diff
@@ -142,6 +142,8 @@ generate_shared_manifests(){
   printf -- "- Generating shared manifests:\n"
   printf -- "  - tekton-chains manifest:\n"
   tekton_chains_manifest 2>&1 | indent 4
+  printf -- "  - tekton-results manifest:\n"
+  tekton_results_manifest #2>&1 | indent 4
```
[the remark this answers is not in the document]
CathalOConnorRH: Commented for testing, I've removed commented code now as it was blocking the request for username/password from displaying

ckcp/tekton-results/openshift/configmap.yaml
CathalOConnorRH: Updated yaml

ckcp/tekton-results/openshift/kustomization.yaml
CathalOConnorRH: Moved.

images/access-setup/content/bin/setup_compute.sh
CathalOConnorRH: added both options

ckcp/openshift_dev_setup.sh
CathalOConnorRH: removed create secret as it shouldn't be needed.

images/access-setup/content/bin/setup_compute.sh
```diff
@@ -176,6 +178,28 @@ tekton_chains_manifest(){
   printf "OK\n"
 }
 
+tekton_results_manifest(){
+  manifest="$manifests_dir/compute/tekton-results/tekton-results-secret.yaml"
+  if [ ! -e "$manifest" ]; then
+    manifests_tmp_dir="$(dirname "$manifest")/tmp"
```
Roming22, line 184: This var seems to be unused and could be removed (L184,L185,L194).

images/access-setup/content/bin/setup_compute.sh
```diff
@@ -176,6 +178,28 @@ tekton_chains_manifest(){
   printf "OK\n"
 }
 
+tekton_results_manifest(){
+  manifest="$manifests_dir/compute/tekton-results/tekton-results-secret.yaml"
```
Roming22, line 182: I would expect that manifest to be applied in https://github.com/openshift-pipelines/pipeline-service/blob/main/images/cluster-setup/bin/install.sh#L133

CathalOConnorRH: added now.

images/access-setup/content/bin/setup_compute.sh
CathalOConnorRH: var removed

Roming22:
@bnallapeta @Mo3m3n @ramessesii2 The PR is quite large, so I'd like another pair of eyes to go through it before it's merged.

CathalOConnorRH:
@Mo3m3n  I've added a README to the tekton-results section in the gitops directory.

ckcp/openshift_dev_setup.sh
```diff
@@ -89,6 +89,7 @@ init() {
   APP_LIST=(
             "openshift-gitops"
             "cert-manager"
+            "tekton-results"
```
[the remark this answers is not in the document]
CathalOConnorRH: This PR covers two different install modes where the requirements are different. In a ckcp install we need to configure a postgres db instance where as on workload clusters we will be using an external database.

images/access-setup/content/bin/setup_compute.sh
```diff
@@ -142,6 +142,8 @@ generate_shared_manifests(){
   printf -- "- Generating shared manifests:\n"
   printf -- "  - tekton-chains manifest:\n"
   tekton_chains_manifest 2>&1 | indent 4
+  printf -- "  - tekton-results manifest:\n"
+  tekton_results_manifest
```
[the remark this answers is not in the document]
CathalOConnorRH: This had to be removed so that the user could be asked for credentials for configuring the database. If this is left in, the script hangs with no output until the credentials are entered and then the output is displayed which would be too late.

ckcp/openshift_dev_setup.sh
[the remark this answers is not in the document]
Mo3m3n: Understood, so you're right this part is only relevant to local dev environment.
In such case, what about moving this task to `install_pipeline_service()`. This is where we handle deploying pipeline service in a local ckcp cluster and includes all the necessary bricks. IMO since the only purpose of this db is to be used with tekton_results then we can apply its manifets in `install_pipeline_service()`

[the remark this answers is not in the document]
CathalOConnorRH: sure, updated now.

images/access-setup/content/bin/setup_compute.sh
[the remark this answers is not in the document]
Mo3m3n: Ok I would have preferred to keep the possibility to run ckcp script in a non interactive way so we can just run it and forget about it. But we can handle this in a later task.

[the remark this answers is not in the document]
Roming22: There should be no user interaction when using the script (both for UX, and because it's used by the CI bot). Please find a way to inject the credentials.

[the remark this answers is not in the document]
Mo3m3n: Actually I am wondering if this is going to work with github test
@Roming22 @xinredhat if you confirm that github-ci is running a fresh ckcp cluster then the script is going to hang for user input.

Roming22:
@CathalOConnorRH Please update `ckcp/test.sh` to add a `results` test that retrieve something from the DB (can be based on the pipelines test). There's currently no way to validate that the service was setup properly.

gitops/argocd/tekton-results/README.md
```diff
@@ -0,0 +1,10 @@
+# Tekton Results
+
+## ***Description*** 
+This will install tekton results on cluster to gather Tekton Pipelinerun results for long term storage.
+
+Installation is based on the [installation instructions](https://github.com/tektoncd/results/blob/main/docs/install.md) from upstream tekton results
```
avinal, line 6: Should we rather point to our fork?

images/access-setup/content/bin/setup_compute.sh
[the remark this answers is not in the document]
avinal: I don't know if this makes sense, but can we provide credentials using config.yaml?

images/access-setup/content/bin/setup_compute.sh
```diff
@@ -176,6 +178,28 @@ tekton_chains_manifest(){
   printf "OK\n"
 }
 
+tekton_results_manifest(){
+  manifest="$manifests_dir/compute/tekton-results/tekton-results-secret.yaml"
+  if [ ! -e "$manifest" ]; then
+    manifests_tmp_dir="$(dirname "$manifest")/tmp"
+    mkdir -p "$manifests_tmp_dir"
+    read -p "Please enter database username: " -r database_user
+    read -p "Please enter database password: " -s -r database_password
```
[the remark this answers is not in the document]
avinal: yes this is possible, even in the installation instructions we have a db with random password: 

https://github.com/openshift-pipelines/tektoncd-results/blob/172d1b55bf7cfe62bfe1c7a0071cf9820a2dadb1/docs/install.md?plain=1#L24

images/access-setup/content/bin/setup_compute.sh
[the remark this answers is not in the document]
CathalOConnorRH: I've updated to now pass tekton details as arguments or generate if not passed into the scripts

images/access-setup/content/bin/setup_compute.sh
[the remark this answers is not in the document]
CathalOConnorRH: updated now

gitops/argocd/tekton-results/README.md
CathalOConnorRH: I think it'll be cleaner to confirm this PR works as expected and then update what's needed for our fork and the included changes we may need once this is merged.

ckcp/openshift_dev_setup.sh
```diff
@@ -332,20 +348,25 @@ patches:
     --kcp-org "$kcp_org" \
     --kcp-workspace "$kcp_workspace" \
     --work-dir "$WORK_DIR" \
-    --kustomization "$GIT_URL/gitops/kcp/pac-manager?ref=$GIT_REF" |
+    --kustomization "$GIT_URL/gitops/kcp/pac-manager?ref=$GIT_REF" 2>&1|
```
Roming22, line 351: With the latest changes, my understanding is that this change can be reverted.

ckcp/openshift_dev_setup.sh
```diff
@@ -360,6 +381,9 @@ install_pipeline_service() {
     WEBHOOK_SECRET="placeholder_webhook" \
     "$GITOPS_DIR/pac/setup.sh" | indent 4
 
+  echo -n "- Install tekton-results DB: "
+  kubectl apply -k "manifests/tekton-results/openshift" >/dev/null 2>&1
```
Roming22, line 385: Do not redirect the output, instead use `indent` to format it properly.

ckcp/test.sh
```diff
@@ -181,6 +185,50 @@ test_triggers() {
...
+  echo $PORTFORWARD_PID
+  # download the API Server certificate locally and configure gRPC.
+  KUBECONFIG="$KUBECONFIG" kubectl get secrets tekton-results-tls -n openshift-pipelines --template='{{index .data "tls.crt"}}' | base64 -d > /tmp/results.crt
+  export GRPC_DEFAULT_SSL_ROOTS_FILE_PATH=/tmp/results.crt
+  
+  RESULT_UID=$(kubectl get pipelinerun "$PIPELINE_RUN" -n "$KCP_NS_NAME" -o yaml | yq .metadata.uid)
+  
+  # This is required to pass shellcheck due to the single quotes in the GetResult name parameter.
+  RECORD_CMD="grpc_cli call --channel_creds_type=ssl --ssl_target=tekton-results-api-service.openshift-pipelines.svc.cluster.local --call_creds=access_token=$(kubectl get secrets -n openshift-pipelines -o jsonpath="{.items[?(@.metadata.annotations['kubernetes\.io/service-account\.name']=='tekton-results-debug')].data.token}"| cut -d ' ' -f 2 | base64 --decode) localhost:50051 tekton.results.v1alpha2.Results.GetResult 'name: $KCP_NS_NAME/results/$RESULT_UID')" >/dev/null
+  RECORD="${RECORD_CMD}"
```
Roming22, line 217: That seems like a simple `a=b`. Do you expect `RECORD_CMD` to be executed?

ckcp/test.sh
```diff
@@ -181,6 +185,50 @@ test_triggers() {
...
+
+  # kill backgrounded port forwarding process as it is no longer required. 
+  kill $PORTFORWARD_PID
```
Roming22, line 220: Shellcheck is likely to report an error about missing '"' here. Please double check, not sure I understand why an error was not reported.

ckcp/openshift_dev_setup.sh
CathalOConnorRH: Reverted now.

ckcp/openshift_dev_setup.sh
CathalOConnorRH: replaced with Indent

ckcp/test.sh
CathalOConnorRH: Good spot. I've updated now and confirmed the record is returned correctly

ckcp/test.sh
CathalOConnorRH: There is no error with or without the quote, I think shellcheck can allow some commands without quotes. I've added them for completeness and to ensure no future issues if they are detected at a later stage.

images/access-setup/content/bin/setup_compute.sh
```diff
@@ -59,13 +67,16 @@ Optional arguments:
...
     %s -d -k /path/to/compute.kubeconfig
-" "${0##*/}" "$KUSTOMIZATION" "$GIT_URL" "$GIT_REF" "${0##*/}" >&2
+" "${0##*/}" "$KUSTOMIZATION" "$GIT_URL" "$GIT_REF" "$TEKTON_RESULTS_DATABASE_USER" "$TEKTON_RESULTS_DATABASE_PASSWORD" "${0##*/}" >&2
 }
 
 parse_args() {
   KUSTOMIZATION=${KUSTOMIZATION:-github.com/openshift-pipelines/pipeline-service/gitops/compute/pac-manager?ref=main}
   GIT_URL=${GIT_URL:-"https://github.com/openshift-pipelines/pipeline-service.git"}
   GIT_REF=${GIT_REF:="main"}
+  TEKTON_RESULTS_DATABASE_USER=${TEKTON_RESULTS_DATABASE_USER:="tekton"}
```
Roming22, line 77: I'm not sure those params should have default values.

images/access-setup/content/bin/setup_compute.sh
```diff
@@ -176,6 +197,26 @@ tekton_chains_manifest(){
   printf "OK\n"
 }
 
+tekton_results_manifest(){
+  manifest="$manifests_dir/compute/tekton-results/tekton-results-secret.yaml"
+  if [ ! -e "$manifest" ]; then
+    manifests_tmp_dir="$(dirname "$manifest")/tmp"
+    mkdir -p "$manifests_tmp_dir"
+    {
```
Roming22, line 205: The TEKTON_RESULTS_* vars should probably be checked before generating the secret, and an error returned if unset.

images/access-setup/content/bin/setup_compute.sh
CathalOConnorRH: defaults removed

images/access-setup/content/bin/setup_compute.sh
CathalOConnorRH: added check and error output

adambkaplan:
Current failure:

```
ekton-chains signing-secrets --from-file=/tmp/tmp.YXhWwhhvek/credentials/manifests/compute/tekton-chains/tmp --dry-run=client -o yaml
        + yq '. += {"immutable" :true}'
        + yq 'sort_keys(.)'
        + rm -rf /tmp/tmp.YXhWwhhvek/credentials/manifests/compute/tekton-chains/tmp
        ++ yq .data
        ++ grep -cE '^cosign.key:|^cosign.password:|^cosign.pub:'
        + '[' 3 '!=' 3 ']'
        + printf 'OK\n'
        OK
    + printf -- '  - tekton-results manifest:\n'
      - tekton-results manifest:
    + tekton_results_manifest
    + indent 4
    ++ printf %4s
    + sed 's/^/    /'
        + manifest=/tmp/tmp.YXhWwhhvek/credentials/manifests/compute/tekton-results/tekton-results-secret.yaml
        + '[' '!' -e /tmp/tmp.YXhWwhhvek/credentials/manifests/compute/tekton-results/tekton-results-secret.yaml ']'
        + [[ -z tekton ]]
        + [[ -z i5pjaEMiv5fXAfmzay6mnI7Oorg= ]]
        //images/access-setup/content/bin/setup_compute.sh: line 200: /tmp/tmp.YXhWwhhvek/credentials/manifests/compute/tekton-results/tekton-results-secret.yaml: No such file or directory
        ++ yq .data
        //images/access-setup/content/bin/setup_compute.sh: line 216: /tmp/tmp.YXhWwhhvek/credentials/manifests/compute/tekton-results/tekton-results-secret.yaml: No such file or directory
        ++ grep -cE '^DATABASE_USER:|^DATABASE_PASSWORD:'
        + '[' 0 '!=' 2 ']'
        + printf '[ERROR] Invalid manifest: '\''%s'\''' /tmp/tmp.YXhWwhhvek/credentials/manifests/compute/tekton-results/tekton-results-secret.yaml
        [ERROR] Invalid manifest: '/tmp/tmp.YXhWwhhvek/credentials/manifests/compute/tekton-results/tekton-results-secret.yaml'+ exit 1
```

Roming22:
@CathalOConnorRH You need to change the regex to "^ *DATABASE_" as the value is indented and we don't care by how much.

ckcp/openshift_dev_setup.sh
```diff
@@ -34,6 +34,14 @@ Optional arguments:
         Directory in which to create the gitops file structure.
         If the directory already exists, all content will be removed.
         By default a temporary directory will be created in $TMPDIR.
+    --tekton-results-database-user TEKTON_RESULTS_DATABASE_USER
```
Mo3m3n, line 37: @CathalOConnorRH Can you please move the database params in `config/config.yaml`.
In "openshift_dev_setup.sh" we only deal with global params, but for specific app configs we centralise that in "config/config.yaml" otherwise we may end up with a myriad of script args for the different apps to configure.
Smth like the following would work:
```yaml
tekton_results_db:
  user:
  password:
```
You can find in the ckcp script how we parse that file.

Mo3m3n:
> /lgtm @Mo3m3n Please check if your comments have been answered. If you don't see any other issues, please merge.

We're almost good, one last review I have submitted then will try this locally and merge it if the test is successful

ckcp/openshift_dev_setup.sh
```diff
@@ -361,6 +386,10 @@ install_pipeline_service() {
     WEBHOOK_SECRET="placeholder_webhook" \
     "$GITOPS_DIR/pac/setup.sh" | indent 4
 
+  echo -n "- Install tekton-results DB: "
+  kubectl apply -k "manifests/tekton-results-db/openshift" 2>&1 |
```
Mo3m3n, line 390: You will need to prefix `$CKCP_DIR` otherwise this is not going to work if run outside of ckcp folder

images/access-setup/content/bin/setup_compute.sh
```diff
@@ -176,6 +197,32 @@ tekton_chains_manifest(){
...
+      exit 1
+    fi
+
+    {
+      echo "---"
+      kubectl create namespace openshift-pipelines --dry-run=client -o yaml
+      echo "---"
+      kubectl create secret generic -n openshift-pipelines tekton-results-database --from-literal=DATABASE_USER="$TEKTON_RESULTS_DATABASE_USER" --from-literal=DATABASE_PASSWORD="$TEKTON_RESULTS_DATABASE_PASSWORD" --dry-run=client -o yaml
+    } > "$manifest"
+    if [ "$(yq ".data" < "$manifest" | grep -cE "^*DATABASE_USER:|^*DATABASE_PASSWORD:")" != "2" ]; then
```
Mo3m3n, line 218: no need for wildcards (they are producing warnings) just use `grep -cE "DATABASE_USER|DATABASE_PASSWORD"` it should work fine

Mo3m3n (changes requested):
currently commit message is `Merge branch 'main' into PLNSRVCE-420`
You will need to change it to relate to what the PR is about

images/access-setup/content/bin/setup_compute.sh
CathalOConnorRH: updated now.

ckcp/openshift_dev_setup.sh
CathalOConnorRH: updated now

Mo3m3n (approved):

PR #228, opened 2022-09-07
reverting PR #165 as parameters don't get properly processed

fgiloux:

### DIFF
--- a/local/kcp/start.sh
+++ b/local/kcp/start.sh
@@ -23,7 +23,7 @@ printf "KCP_DIR: a directory with kcp source code, default to a git clone of kcp
 printf "KCP_BRANCH: the kcp branch to use. Mind that the script will do a git checkout, to a default release if the branch is not specified\n"
 printf "KCP_RUNTIME_DIR: the location of the kcp runtime files, default to a temporary directory\n"
 printf "PARAMS: the parameters to start kcp with. They need to be provided as a list, e.g. \n"
-printf "        \"(--token-auth-file ./local/kcp/kcp-tokens --profiler-address localhost:6060 -v 6)\"\n\n"
+printf "        \"--token-auth-file ./local/kcp/kcp-tokens --profiler-address localhost:6060 -v 6\"\n\n"
 
 precheck() {
   if ! command -v "$1" &> /dev/null
@@ -52,7 +52,7 @@ kcp-binaries() {
 
 kcp-start() {
   printf "Starting KCP server ...\n"
-  (cd "${KCP_RUNTIME_DIR}" && exec "${KCP_DIR}/bin/kcp" start "${PARAMS[@]}") &> "${KCP_RUNTIME_DIR}/kcp.log" &
+  (cd "${KCP_RUNTIME_DIR}" && exec "${KCP_DIR}/bin/kcp" start ${PARAMS}) &> "${KCP_RUNTIME_DIR}/kcp.log" &
   KCP_PID=$!
   KCP_PIDS+=("${KCP_PID}")
   wait_command "ls ${KUBECONFIG}" 30
@@ -118,11 +118,9 @@ KCP_CIDS=()
 
 PARAMS="${PARAMS:-}"
 if [[ -z "${PARAMS}" ]]; then
-  PARAMS=(
-    --token-auth-file "${PARENT_PATH}/kcp-tokens"
-    --profiler-address localhost:6060
-    -v 2
-  )
+  PARAMS="--token-auth-file ${PARENT_PATH}/kcp-tokens \
+    --profiler-address localhost:6060 \
+    -v 2"
 fi
 
 kcp-start


Roming22:
Can you please fix the shellcheck issue(s) that's preventing the merge?

fgiloux:
Can you please deactivate the shellcheck rule that's preventing the merge?

local/kcp/start.sh
```diff
@@ -52,7 +52,7 @@ kcp-binaries() {
 
 kcp-start() {
   printf "Starting KCP server ...\n"
-  (cd "${KCP_RUNTIME_DIR}" && exec "${KCP_DIR}/bin/kcp" start "${PARAMS[@]}") &> "${KCP_RUNTIME_DIR}/kcp.log" &
+  (cd "${KCP_RUNTIME_DIR}" && exec "${KCP_DIR}/bin/kcp" start ${PARAMS}) &> "${KCP_RUNTIME_DIR}/kcp.log" &
```
bnallapeta, line 55: I think ${PARAMS} still can be quoted which is what the shellcheck is warning us about - **"${PARAMS}"**
(without the [@])

fgiloux: my experience shows otherwise. I have you tested it?

bnallapeta: sorry for the delay. I tried running it locally. The script is actually running with the older code ("${PARAMS[@]}"). But I'm not sure on how to verify that its reading and processing those params.

./local/kcp/start.sh -v 6 --profiler-address localhost:6060 --discovery-poll-interval 3s

Snippet from logs:
```
Starting KCP server ...
KCP server started: 209332
Waiting for KCP to be ready ...
KCP ready
KUBECONFIG=/tmp/kcp-pipeline-service.fE347styk/.kcp/admin.kubeconfig
```

Roming22: `.start.sh` without any additional configuration works OK for me on both `main` and `pr-228`.

`PARAMS="--token-auth-file ../kcp-tokens --profiler-address localhost:6060 -v 6" ./start.sh` fails on both `main` and `pr-228`.

@fgiloux Could you provide us with the command that fails on `main` and succeeds in `pr-228`?

fgiloux:
~~~
$ KCP_RUNTIME_DIR=/tmp/kcp-runtime KCP_DIR=/home/fgiloux/development/kcp-dev/kcp PARAMS="--feature-gates=KCPSyncerTunnel=true --token-auth-file /tmp/kcp-tokens --profiler-address localhost:6060 -v 6" ./local/kcp/start.sh
~~~
works with PR-228, does not with main:
~~~
Error: invalid argument "KCPSyncerTunnel=true --token-auth-file /tmp/kcp-tokens --profiler-address localhost:6060 -v 6" for "--feature-gates" flag: invalid value of KCPSyncerTunnel=true --token-auth-file /tmp/kcp-tokens --profiler-address localhost:6060 -v 6, err: strconv.ParseBool: parsing "true --token-auth-file /tmp/kcp-tokens --profiler-address localhost:6060 -v 6": invalid syntax
~~~

It is not just that the command needs to succeed but the parameters need to be processed as well. You can check that with the token file for instance.
I thought that using brackets would do the trick with main and the server starts but the later parameters like the auth file are not taken in consideration.
~~~
$ KCP_RUNTIME_DIR=/tmp/kcp-runtime KCP_DIR=/home/fgiloux/development/kcp-dev/kcp PARAMS="(--feature-gates=KCPSyncerTunnel=true --token-auth-file /tmp/kcp-tokens --profiler-address localhost:6060 -v 6)" ./local/kcp/start.sh
~~~

The command you shared works for me with PR-228.

Roming22: If you use brackets, I'd think line 119 might need to be changed to `PARAMS="${PARAMS[@]:-}"` to retain all elements.

Roming22: @fgiloux I believe this is how to solve the issue without angering shellcheck:
``` 
PARAMS="${PARAMS:-"--token-auth-file ${PWD}/kcp-tokens --profiler-address localhost:6060 -v 2"}"
mapfile -d ' ' PARAMS < <(echo "$PARAMS")
```
First line sets the default value, the second one converts it into an array.

I use the following script to validate the behavior:
```
test() {
  # Set default value
  PARAMS="${PARAMS:-"123 abc"}"
  # Transform params into an array
  mapfile -d ' ' PARAMS < <(echo "$PARAMS")
  # Each word should be on its own line
  python -c "
import sys
for a in sys.argv[1:]:
  print(a)
" "${PARAMS[@]}"
}
test
PARAMS="$RANDOM foo bar" test
```

fgiloux: @Roming22 you should try to run: `kcp start options`

Roming22: I've submitted a PR with the fix: see #235 .

Roming22:
Superseded by #235

PR #231, opened 2022-09-08
PLNSRVCE-697: Upgrade kcp to v0.8.0

bnallapeta:
- In this PR, replaced the use of release versions of kcp to tags.
- Added a timeout for a loop in ckcp script.
- Removed KCP_BRANCH arg in access-setup Github Actions because that var is not used anywhere in access setup image.
- In a subsequent PR, will work on automating the upgrade process and using digests.


local/kcp/start.sh
```diff
@@ -42,8 +42,8 @@ kcp-binaries() {
     git clone https://github.com/kcp-dev/kcp.git
     KCP_DIR="${KCP_PARENT_DIR}/kcp"
     pushd kcp
-    KCP_BRANCH="${KCP_BRANCH:-release-0.7}"
-    git checkout "${KCP_BRANCH}"
+    KCP_BRANCH="${KCP_BRANCH:-v0.8.0}"
+    git checkout tags/"${KCP_BRANCH}" -b "${KCP_BRANCH}"
```
Roming22, line 46: We should probably rename KCP_BRANCH to KCP_TAG. That could be done in another PR.

xinredhat:
Apart from Romain's comment of renaming KCP_BRANCH to KCP_TAG, LGTM

DEPENDENCIES.md
```diff
@@ -13,7 +13,7 @@
...
-| kcp                           | release-0.7                                                                                                                                | Platform - acts as a control plane for Pipeline Service                      | Upgrades follow kcp-stable releases                                                                                                                                       |
+| kcp                           | v0.8.0                                                                                                                                | Platform - acts as a control plane for Pipeline Service                      | Upgrades follow kcp-stable releases                                                                                                                                       |
```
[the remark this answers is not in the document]
bnallapeta: Updated.

.github/workflows/build-push-images.yaml
```diff
@@ -124,7 +122,7 @@ jobs:
           containerfiles: |
             ./images/kcp-registrar/Dockerfile
           build-args: |
-            KCP_BRANCH=release-0.7
+            KCP_BRANCH=v0.8.0
```
fgiloux, line 125: latest is [v0.8.2](https://github.com/kcp-dev/kcp/releases/tag/v0.8.2)
v0.8.0 did not reach stable because of issues. We should start with a later version.

ckcp/openshift_dev_setup.sh
```diff
@@ -305,9 +305,13 @@ patches:
   # Workaround to prevent the creation of a new workspace until KCP is ready.
   # This fixes `error: creating a workspace under a Universal type workspace is not supported`.
   ws_name=$(echo "$kcp_org" | cut -d: -f2)
-  while ! KUBECONFIG="$KUBECONFIG_KCP" kubectl kcp workspace create "$ws_name" --type root:organization --ignore-existing >/dev/null; do
-    sleep 5
-  done
+  local sec=0
+  if  [ "$sec" -lt 100 ]; then
```
fgiloux, line 309: It looks like it would always be true :-)

ckcp/openshift_dev_setup.sh
```diff
@@ -305,9 +305,13 @@ patches:
...
+    while ! KUBECONFIG="$KUBECONFIG_KCP" kubectl kcp workspace create "$ws_name" --type root:organization --ignore-existing >/dev/null; do
+      sleep 5
```
fgiloux, line 311: I guess you meant to have the if inside of the loop with a break when it is true

.github/workflows/build-push-images.yaml
Roming22, line 125: Shouldn't this also become `KCP_TAG`?

local/kcp/start.sh
```diff
@@ -42,8 +42,8 @@ kcp-binaries() {
     git clone https://github.com/kcp-dev/kcp.git
     KCP_DIR="${KCP_PARENT_DIR}/kcp"
     pushd kcp
-    KCP_BRANCH="${KCP_BRANCH:-release-0.7}"
-    git checkout "${KCP_BRANCH}"
+    KCP_TAG="${KCP_BRANCH:-v0.8.0}"
```
fgiloux, line 45: It is nice not to break the user interface as you did but it is getting a bit confusing. What do you think of renaming the script parameter? The script is only for local dev, isn't it? Is this script called somewhere, in CI maybe?
I am fine either way.

fgiloux (changes requested):

local/kcp/start.sh
bnallapeta: Sure, renamed the param to KCP_TAG as well.

ckcp/openshift_dev_setup.sh
bnallapeta: Fixed.

.github/workflows/build-push-images.yaml
bnallapeta: Changed it to KCP_TAG.

bnallapeta: As discussed, next PR will do automatic upgrade to v0.8.2 (or whatever is the latest one available at the time)

local/kcp/start.sh
```diff
@@ -20,7 +20,7 @@ set -o nounset
 # Parameters
 printf "The following optional parameters can be passed to the script:\n"
 printf "KCP_DIR: a directory with kcp source code, default to a git clone of kcp in the system temp directory\n"
-printf "KCP_BRANCH: the kcp branch to use. Mind that the script will do a git checkout, to a default release if the branch is not specified\n"
+printf "KCP_TAG: the kcp branch to use. Mind that the script will do a git checkout, to a default release if the branch is not specified\n"
```
fgiloux, line 23: KCP_TAG: the kcp branch to use => KCP_TAG: the kcp tag to use
 if the branch is not specified =>  if the tag is not specified

bnallapeta: Fixed.

PR #233, opened 2022-09-09
fixed issue of testing pipelinerun

xinredhat:
**Changes**

Fixed issue by specifying a WORK_DIR as parameter. so it could find the correct kcp and compute's kubeconfig files. For now it only support testing pipelines.

After you execute the `ckcp/openshift_dev_setup.sh` command, you could use `ckcp/test.sh` to verify if pipeline-service works. The command example as below
```
WORK_DIR=./ckcp/workdir CASES=pipelines ./ckcp/test.sh
```

### DIFF
--- a/ckcp/test.sh
+++ b/ckcp/test.sh
@@ -6,10 +6,23 @@ set -euo pipefail
 # Uncomment the below line to enable debugging
 # set -x
 
-SCRIPT_DIR="$(cd "$(dirname "$0")" >/dev/null ; pwd)"
-KUBECONFIG="${KUBECONFIG:-$HOME/.kube/config}"
-export KUBECONFIG=$KUBECONFIG
-KUBECONFIG_KCP="${KUBECONFIG_KCP:-$SCRIPT_DIR/work/credentials/kubeconfig/kcp/ckcp-ckcp.default.pipeline-service-compute.kubeconfig}"
+usage() {
+
+    printf "Usage: WORK_DIR=/workspace CASES=pipelines ./test.sh\n\n"
+
+    # Parameters
+    printf "WORK_DIR: the location of the gitops files\n"
+    printf "CASES: comma separated list of test cases. Test cases must be any of 'pipelines' or 'triggers'. Pipelines test cases are run by default.\n"
+}
+
+prechecks() {
+    WORK_DIR="${WORK_DIR:-}"
+    if [[ -z "${WORK_DIR}" ]]; then
+        printf "WORK_DIR not set\n\n"
+        usage
+        exit 1
+    fi
+}
 
 get_namespace() {
   # Retrieve the KCP namespace id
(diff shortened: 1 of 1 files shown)


ckcp/test.sh
```diff
@@ -6,10 +6,30 @@ set -euo pipefail
...
+
+prechecks() {
+    WORK_DIR="${WORK_DIR:-}"
+    if [[ -z "${WORK_DIR}" ]]; then
+        printf "WORK_DIR not set\n\n"
+        usage
+        exit 1
+    fi
+
+    CASES="${CASES:-}"
```
Roming22, line 26: Why not run all cases by default?

ckcp/test.sh
```diff
@@ -6,10 +6,30 @@ set -euo pipefail
...
-KUBECONFIG="${KUBECONFIG:-$HOME/.kube/config}"
-export KUBECONFIG=$KUBECONFIG
-KUBECONFIG_KCP="${KUBECONFIG_KCP:-$SCRIPT_DIR/work/credentials/kubeconfig/kcp/ckcp-ckcp.default.pipeline-service-compute.kubeconfig}"
+usage() {
+
+    printf "Usage: WORK_DIR=/workspace CASES=pipelines ./test.sh\n\n"
+
+    # Parameters
+    printf "WORK_DIR: the location of the gitops files\n"
+    printf "CASES: the name of test casesm, the allowed cases are 'pipelines' or 'triggers' or 'pipelines,triggers'\n"
```
Roming22, line 15: `CASES: comma separated list of test cases. Test cases must be any of 'pipelines' or 'triggers'. All test cases are run by default.`

ckcp/test.sh
```diff
@@ -35,66 +55,71 @@ get_namespace() {
...
 
-      sleep 20
-      KUBECONFIG="$KUBECONFIG_KCP" kubectl get pipelineruns
+main() {
+  prechecks
+  init
+  IFS="," read -r -a cases <<< "$CASES"
+  for case in "${cases[@]}"
+  do
+    if [ "$case" == "pipelines" ]; then
```
Roming22, line 113: that should probably be in the form of

```case $case in
pipelines|triggers)
  test_$case
  ;;
*)
  echo "Incorrect case name '[$case]'"
  usage
  exit 1
  ;;
esac
```

ckcp/test.sh
xinredhat: fixed.

ckcp/test.sh
xinredhat: fixed

ckcp/test.sh
xinredhat: fixed

ckcp/test.sh
```diff
@@ -35,66 +48,76 @@ get_namespace() {
   echo "$KCP_NS_NAME"
 }
 
+init() {
+  KUBECONFIG_KCP="$WORK_DIR/credentials/kubeconfig/kcp/admin.kubeconfig.base"
```
Roming22, line 52: Do not use the *.base files, they're used just a couple of times during setup, and are kept around just for debug purposes. Use the *.kubeconfig files instead.

ckcp/test.sh
```diff
@@ -35,66 +48,86 @@ get_namespace() {
...
+compute_kubeconfig() {
+    mapfile -t files < <(find "$WORK_DIR/credentials/kubeconfig/compute/" -name \*.kubeconfig)
+    echo "${files[0]}"
+}
+
+init() {
+  KUBECONFIG_KCP="$(kcp_kubeconfig)"
+  KUBECONFIG="$(compute_kubeconfig)"
+  export KUBECONFIG
+  CASES="${CASES:-"pipelines,triggers"}"
```
bnallapeta, line 65: currently we don't have triggers working right? 
If we default CASES to pipelines and triggers, then the script would fail in case user does not provide any values for CASES. Please default it to pipelines. 
@Roming22 can take a call on chains in his PR.

bnallapeta (commented):
lgtm apart from a minor comment.

ckcp/test.sh
xinredhat: fixed

ckcp/test.sh
xinredhat: > currently we don't have triggers working right?

right. fixed.

bnallapeta: Did you still not push the commit? I still see both pipeline,triggers on line 65.

ckcp/test.sh
```diff
@@ -6,10 +6,23 @@ set -euo pipefail
...
-KUBECONFIG="${KUBECONFIG:-$HOME/.kube/config}"
-export KUBECONFIG=$KUBECONFIG
-KUBECONFIG_KCP="${KUBECONFIG_KCP:-$SCRIPT_DIR/work/credentials/kubeconfig/kcp/ckcp-ckcp.default.pipeline-service-compute.kubeconfig}"
+usage() {
+
+    printf "Usage: WORK_DIR=/workspace CASES=pipelines ./test.sh\n\n"
+
+    # Parameters
+    printf "WORK_DIR: the location of the gitops files\n"
+    printf "CASES: comma separated list of test cases. Test cases must be any of 'pipelines' or 'triggers'. All test cases are run by default.\n"
```
bnallapeta, line 15: "Pipelines test cases are run by default."

xinredhat: @bnallapeta do I need to mention it doesn't `triggers` now?

ckcp/test.sh
xinredhat: sorry for my bad. Done.

ckcp/test.sh
bnallapeta: Last part of the line has to be updated from "All test cases are run by default." to "Pipelines test cases are run by default."

xinredhat: it's already fixed in the last commit :)

Roming22 (approved):
@bnallapeta Please merge if the PR looks good to you to.

PR #239, opened 2022-09-12
Check if the installed kcp plugin is latest

bnallapeta:


ckcp/openshift_dev_setup.sh
```diff
@@ -149,6 +149,12 @@ precheck() {
     printf "kcp plugin could not be found\n"
     exit 1
   fi
+
+  mapfile -t kubectl_kcp_version < <(kubectl-kcp --version | yq -r 'split("-")')
+  if [ "${kubectl_kcp_version[-1]}" -eq "$kcp_version" ]; then
+      printf "kcp plugin version is different from %s. Please install %s\n" "$kcp_version" "$kcp_version"
```
Roming22, line 155: I think it might be better to split the error and the proposed action for the user. The error should got to stderr as well.

```
printf "[ERROR] Bad kcp plugin version. Expected '%s', got '%s'.\n" "$kcp_version" "$kubectl_kcp_version" >&2
printf "Please install the kcp plugin '%s'\n" "$kcp_version"
```

Roming22 (approved):

PR #241, opened 2022-09-14
Migrate to openshift 4.11

Mo3m3n:
Migrating Openshift 4.11 requires:
- [Manually creating a service account token secret](https://github.com/openshift-pipelines/pipeline-service/commit/1945d7fb9a38905555fe32ba4281ff3a14d4bc20)
- [Restricting security context of kcp pods](https://github.com/openshift-pipelines/pipeline-service/commit/8338afe4c01a0caab8030372a2a0d711afe35fc1) which introduces a breaking change with 4.10


bnallapeta:
@Mo3m3n Tried running it locally on a 4.11 cluster. Fails with the below error:

```
[bnr@bnr pipelines-service]$ ./ckcp/openshift_dev_setup.sh 
Working directory: /tmp/tmp.ZlBOATYExn
[openshift-gitops]
  - OpenShift-GitOps: OK
  - Argo CD dashboard: .............................OK
  - Argo CD URL: https://openshift-gitops-server-openshift-gitops.apps.ci-ln-6jrgsck-72292.origin-ci-int-gce.dev.rhcloud.com
  - Argo CD Login: OK
  - Register host cluster to ArgoCD as 'plnsvc': 
INFO[0000] ServiceAccount "argocd-manager" created in namespace "kube-system" 
INFO[0001] ClusterRole "argocd-manager-role" created    
INFO[0001] ClusterRoleBinding "argocd-manager-role-binding" created 
FATA[0032] Failed to wait for service account secret: timed out waiting for the condition 

```

Mo3m3n:
@bnallapeta 
I can't reproduce your error on a fresh 4.11 cluster:
```
➜ oc version
Client Version: 4.10.22
Server Version: 4.11.0-0.nightly-2022-09-14-233224
Kubernetes Version: v1.24.0+3882f8f
```

```
➜ ./ckcp/openshift_dev_setup.sh
Working directory: /tmp/tmp.ADFrj86aYY
[openshift-gitops]                                                  
  - OpenShift-GitOps: OK                                            
  - Argo CD dashboard: ..........................OK                 
  - Argo CD URL: https://openshift-gitops-server-openshift-gitops.apps.ci-ln-v4n4sdb-72292.origin-ci-int-gce.dev.rhcloud.com
  - Argo CD Login: OK                                               
  - Register host cluster to ArgoCD as 'plnsvc':                                                                                        
INFO[0000] ServiceAccount "argocd-manager" created in namespace "kube-system"                                                           
INFO[0000] ClusterRole "argocd-manager-role" created    
INFO[0001] ClusterRoleBinding "argocd-manager-role-binding" created                                                                     
INFO[0001] Created bearer token secret for ServiceAccount "argocd-manager"                                                              
    OK                                                                                            
```

I will see if I can ask for more specific debug output from `openshift-gitops`

Mo3m3n:
Can you provide the following:
`oc version`
`oc describe subscriptions  -n openshift-operators openshift-gitops-operator | grep CSV`

Mo3m3n:
@bnallapeta It turns out this requires argocd CLI v2.3.7 or above in order to be able to [create serviceaccount token for v1.24 clusters](https://github.com/argoproj/argo-cd/pull/9546).
I have updated [DPENDENCIES](https://github.com/openshift-pipelines/pipeline-service/pull/241/commits/6000a229db02611b0908d28c7a1ad5d5e4cfc6fb) to reflect that

DEPENDENCIES.md
```diff
@@ -6,7 +6,7 @@
...
-| OpenShift                     | 4.10                                                                                                                                  | Platform                                                                     | Upgrades to next versions 4.11 need to be tested and approved                                                                                                             |
+| OpenShift                     | 4.11                                                                                                                                  | Platform                                                                     | Upgrades to next versions 4.11 need to be tested and approved                                                                                                             |
```
bnallapeta, line 9: One thought that comes to my mind is, from the user's perspective, does this document sound like a minimum requirement doc? If so, we may need to make it clear. Eg: if we mention 4.11 here, does that translate to - "I need 4.11 cluster. It may not work on 4.10 or below." ?

Would like to hear others' thoughts on this. @Roming22 @adambkaplan @fgiloux

DEPENDENCIES.md
```diff
@@ -32,11 +32,11 @@
...
 | kcp plugin         | 0.8.0             | To interact with the kcp workspace                                          | Follows kcp api/server version                                                                                                                                   |
 | tkn pac plugin     | 0.10.0            | To set up PaC                                                               | Optional plugin for customers during the cluster setup phase. Follows PaC Version                                                                                |
-| Argo CD (client)   | 2.3.5             | To run Argo CD related commands                                              | Not recommended for users to use Argo CD to talk to the Argo server instances. Argo is simply used as a model for management of resources using GitOps principles |
+| Argo CD (client)   | 2.3.7             | To run Argo CD related commands                                              | Not recommended for users to use Argo CD to talk to the Argo server instances. Argo is simply used as a model for management of resources using GitOps principles |
```
bnallapeta, line 39: - Hmm, the comment on this is not quite right. ckcp script does require the correct argocd client to be available. Please update the comment.
- Alos, change 2.3.7 to >=2.3.7

images/kcp-registrar/syncer-patch
```diff
@@ -0,0 +1,8 @@
+        securityContext:
```
bnallapeta, line 1: Shouldn't this have a .yaml extension?

DEPENDENCIES.md
Mo3m3n: >This document lists all the various products/tools used in Pipeline Service along with their current versions.

So it does not sound like a minimum but this is rather the versions we are **currently** using which guarantee that Pipeline Service should work as expected.
If we want to make this file a listing of the minimum requirements then I guess we need to start working on different branches: 4.10, 4.11, etc to be able to provide the right dependencies per openshift version.

images/kcp-registrar/syncer-patch
Mo3m3n: It is not required. 
This is just the input for sed command as you can see [here](https://github.com/openshift-pipelines/pipeline-service/commit/8868e632e8c9180e0adb286d1f93017dbdad1a22#diff-010d75c07c9fda576ad68d1bf93347cf3ded384b058db9926fcd2e52ceafb308R218)
Unfortunately I did not find a better way to patch the syncer output  (deployment), not sure if kustomize(as a cleaner alternative) would work here

bnallapeta:
@Mo3m3n As discussed, with OpenShift 4.11, it looks like scc: restricted-v2 is automatically applied to deployments which brings the capabilities mentioned below as per [this](https://cloud.redhat.com/blog/pod-security-admission-in-openshift-4.11) blog that you had linked in the other commit.

- dropping ALL capabilities from containers
they still allow explicitly adding the NET_BIND_SERVICE capability
- defaulting seccompProfile to “runtime/default”
- requiring allowPrivilegeEscalation to be unset or set to false in security contexts


Do verify this on a new cluster post which we won't have to manually patch things.

Mo3m3n:
@bnallapeta 
Actually even though restricted-v2 scc is applied in Openshift 4.11, this does not prevent the Pod Security admission controller from complaining:
> Warning: would violate PodSecurity "restricted:v1.24": allowPrivilegeEscalation != false (container "kcp-syncer" must set securityContex
t.allowPrivilegeEscalation=false), unrestricted capabilities (container "kcp-syncer" must set securityContext.capabilities.drop=["ALL"])
, runAsNonRoot != true (pod or container "kcp-syncer" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "kcp
-syncer" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")    

Users will get that for the kcp syncer  deployment and you can also see the warning for the kcp pod deployment by updating the followin:
```--- a/ckcp/openshift_dev_setup.sh
+++ b/ckcp/openshift_dev_setup.sh
@@ -275,7 +275,7 @@ patches:
   # Deploy ckcp until all resources are successfully appied to OCP cluster 
   local i=0
-  while ! error_msg=$(kubectl apply -k "$ckcp_temp_dir" 2>&1 1>/dev/null); do
+  while ! error_msg=$(kubectl apply -k "$ckcp_temp_dir" ); do
```

For now it is a warning and my understanding is that in the future it will be denied (unless we update the security policy
> [Link](https://cloud.redhat.com/blog/pod-security-admission-in-openshift-4.11)
In a subsequent release, we intend to move the global configuration to enforce the “restricted” pod security profile globally

We can ignore this for now but my understanding is that we need to patch the deployments sooner or later.

DEPENDENCIES.md
Mo3m3n: In the end I have updated argocd CLI version to follow argocd engine version used in openshift gitops(2.4.5).
I guess this is better then using an arbitrary one.

DEPENDENCIES.md
Roming22: The version is the current version as per Moemen's comment.

We're currently focusing on a single OCP branch. In the future we might decide to support multiple OCP branches, in which case we'll need to update the CI.

images/kcp-registrar/register.sh
```diff
@@ -214,6 +214,8 @@ register() {
                 --syncer-image "ghcr.io/kcp-dev/kcp/syncer:$KCP_SYNC_TAG" \
                 --resources deployments.apps,services,ingresses.networking.k8s.io,pipelines.tekton.dev,pipelineruns.tekton.dev,tasks.tekton.dev,runs.tekton.dev,networkpolicies.networking.k8s.io \
                 --output-file "$syncer_manifest"
+            # Set a restricted security context
+            sed -i "/- name: kcp-syncer/r $(dirname "$0")/syncer-patch" "$syncer_manifest"
```
Roming22, line 218: 1. `$(dirname "$0")` should be replace with `$SCRIPT_DIR` (cf other scripts)
2. I would split the bin and data in the image, instead of having everything under one folder.

images/kcp-registrar/syncer-patch
Roming22: Maybe yq can help? This bit seems a bit hacky :)

bnallapeta: Also need to add a copy command in the Dockerfile to copy the patch file.

Mo3m3n: Done
I have also added the yml extension since it is now used by `yq`

images/kcp-registrar/register.sh
Mo3m3n: Done

ckcp/openshift_dev_setup.sh
```diff
@@ -272,6 +272,12 @@ patches:
         description: This value refers to the hostAddress defined in the Route.
         value: $ckcp_route " >>"$ckcp_temp_dir/kustomization.yaml"
 
+	# Enable the "runtime/default" seccomp profile (required for restricted security context)
+	# This is enabled by default in restricted-v2 available in openshift >= 4.11
+	if ! (set +o pipefail && kubectl describe scc restricted restricted-v2 2>/dev/null |
```
bnallapeta, line 277: We are patching only restricted in the next line. Why are we describing restricted-v2 as well?

images/kcp-registrar/bin/register.sh
```diff
@@ -214,6 +219,10 @@ register() {
                 --syncer-image "ghcr.io/kcp-dev/kcp/syncer:$KCP_SYNC_TAG" \
                 --resources deployments.apps,services,ingresses.networking.k8s.io,pipelines.tekton.dev,pipelineruns.tekton.dev,tasks.tekton.dev,runs.tekton.dev,networkpolicies.networking.k8s.io \
                 --output-file "$syncer_manifest"
+            # Set a restricted security context
+            patch="$SCRIPT_DIR/../data/syncer-patch.yml" yq -i \
```
bnallapeta, line 223: sometimes, the CI won't play nice with '..'. Please use "$(dirname "$SCRIPT_DIR")")

images/kcp-registrar/data/syncer-patch.yml
```diff
@@ -0,0 +1,8 @@
+---
```
bnallapeta, line 1: Minor point. But let's maintain consistency with the yaml extension. Plus [official recommendation](https://yaml.org/faq.html) as well :)
.yml -> .yaml

ckcp/openshift_dev_setup.sh
Mo3m3n: This PR enables "runtime/default" seccomp profile in Pod's securityContext.
Openshift version < 4.11 complains if we use this seccomp profile because it is not enabled by default in the available SCCs, thus we are patching the restricted SCC here to support this seccomp profile.
The condition where we don't need to make this scc patch:
- It was already done (describing scc restricted)
- It is already available in openshift >= 4.11 (describing restricted-v2 scc) so no need to patch scc restricted.

bnallapeta:
@Mo3m3n Please take a look at the failed ckcp test. @xinredhat Kindly assist if its a CI issue rather than the PR issue.

ckcp/openshift_dev_setup.sh
```diff
@@ -276,6 +276,12 @@ patches:
...
+		grep -sq 'runtime/default') ; then
+		 kubectl patch scc restricted --type=merge -p '{"seccompProfiles":["runtime/default"]}'
```
adambkaplan, line 283: We should not patch SCCs installed by OpenShift. These changes will be removed by the cluster operators.

Mo3m3n: So I would rather create a clone of that scc, patch it and assign it to the corresponding sa.
Correct ?

adambkaplan: Am I correct in that the reason this is here is because the restricted SCC doesn't allow any value to be set for the seccomp profile? 

Since this is only for `ckcp`, I think it makes sense to declare OpenShift 4.11 as a required prerequisite. We don't deploy `ckcp` in the actual service.

Mo3m3n: Right @adambkaplan the restricted SCC in 4.10 does not allow have this [securityContext](https://github.com/openshift-pipelines/pipeline-service/pull/241/files#diff-6c3c2089e306854fc8e3fb3696029ed7f67f8424c5c0ad53836e5eeb27621cbeR79).
You may ask, why I am changing that in 4.10 while the target of the PR is 4.11? reason is that 4.11 pushes for a restricted security context(via warnings). 
The change works out of the box in 4.11 but the same securitContext requires patching scc in 4.10.
For the sake of consistency I was trying to patch scc in 4.10 to work around that.
We have multiple options:
- backport the security context change in 4.10. In this case I will need assistance from openshift support. I have tried this [doc](https://docs.openshift.com/container-platform/4.10/security/seccomp-profiles.html) and I did not succeed.
- Don't add securityContext in 4.10 and warn users that the securitycontext of kcp pod is not the same between 4.10 and 4.11
- Don't add security Context in 4.10 and declare that 4.11 is a required Prerequisite.

Roming22: @adambkaplan Correct, we would either upgrade the existing cluster, or request a new cluster to deploy the service on, before deprecating the 4.10 cluster.

Roming22: @Mo3m3n I'm fine with making 4.11 a prerequisite, we don't really have the bandwidth to support multiple OCP versions at the moment. By the time we might do it, 4.12 will be out and since we will probably limit support to `n` and `n-1` version, that would again rule out 4.10.

DEPENDENCIES.md
```diff
@@ -6,11 +6,11 @@
...
-| OpenShift                     | 4.10                                                                                                                                  | Platform                                                                     | Upgrades to next versions 4.11 need to be tested and approved                                                                                                             |
+| OpenShift                     | 4.11                                                                                                                                  | Platform                                                                     | Upgrades to next versions 4.11 need to be tested and approved                                                                                                             |
```
bnallapeta, line 9: Comments - Upgrades to next versions **4.12** need to be tested and approved

PR #245, opened 2022-09-16
Record server CA during registration

lcarva:
This ensures a more secure and correct connection from the syncer to the workspace.

I ran into certificate issues when deploying pipeline-service into an existing kcp instance in OpenShift.

### DIFF
--- a/images/kcp-registrar/register.sh
+++ b/images/kcp-registrar/register.sh
@@ -207,9 +207,24 @@ register() {
             --syncer-image "ghcr.io/kcp-dev/kcp/syncer:$KCP_SYNC_TAG" \
             --resources deployments.apps,services,ingresses.networking.k8s.io,pipelines.tekton.dev,pipelineruns.tekton.dev,tasks.tekton.dev,runs.tekton.dev,networkpolicies.networking.k8s.io \
             --output-file "$syncer_manifest"
+        add_ca_cert_to_syncer_manifest "${kcp_kcfg}" "$syncer_manifest"
         KUBECONFIG="${WORKSPACE_DIR}/credentials/kubeconfig/compute/${kubeconfigs[$i]}" kubectl apply \
             --context "${contexts[$i]}" -f "$syncer_manifest"
     done
+}
+
+add_ca_cert_to_syncer_manifest() {
+    config=$1
+    manifest=$2
+    # Display a minified version of the kubeconfig that only includes the current context. Pick
+    # the first server from the list (there should only ever be one due to --minify). Extract
+    # the host name from the server URL.
+    host="$(KUBECONFIG="${config}" kubectl config view --minify | yq '.clusters[0].cluster.server' | cut -d/ -f3)"
+    # hostname may include port number, remove it if it's there
+    servername="$(echo "$host" | cut -d: -f1)"
+    ca_cert="$(openssl s_client -showcerts -servername "${servername}" -connect "${host}" </dev/null | openssl x509 | base64 -w 0)"
+
+    ca_cert="${ca_cert}" yq -i '(select(.stringData.kubeconfig != null)) .stringData.kubeconfig |= (fromyaml | .clusters[].cluster."certificate-authority-data" = env(ca_cert) | to_yaml)' "${manifest}"
 }
 
 configure_synctarget_ws() {


images/kcp-registrar/register.sh
```diff
@@ -214,12 +214,30 @@ register() {
...
     done
 }
 
+add_ca_cert_to_syncer_manifest() {
+    config=$1
+    manifest=$2
+    # Display a minified version of the kubeconfig that only includes the current context. Pick
+    # the first server from the list (there should only ever be one due to --minify). Extract
+    # the host name from the server URL.
+    host="$(KUBECONFIG=${config} kubectl config view --minify | yq .clusters[0].cluster.server | cut -d/ -f3)"
```
ramessesii2, line 230:
```suggestion
    host="$(KUBECONFIG=${config} kubectl config view --minify | yq '.clusters[0].cluster.server' | cut -d/ -f3)"
```
for some reason, yq needed them single quotes in my case.

ramessesii2 (commented):
Couldn't test it completely as of yet. Cluster expired and `ckcp` CI would only be able to test it once we add `openssl` to it.
Is it somehow needed for Openshift 4.11?

lcarva:
> Couldn't test it completely as of yet. Cluster expired and `ckcp` CI would only be able to test it once we add `openssl` to it. Is it somehow needed for Openshift 4.11?

This was needed for OpenShift 4.10 in my case.

images/kcp-registrar/register.sh
```diff
@@ -214,12 +214,30 @@ register() {
...
+    # Display a minified version of the kubeconfig that only includes the current context. Pick
+    # the first server from the list (there should only ever be one due to --minify). Extract
+    # the host name from the server URL.
+    host="$(KUBECONFIG=${config} kubectl config view --minify | yq '.clusters[0].cluster.server' | cut -d/ -f3)"
+    # hostname may include port number, remove it if it's there
+    servername="$(echo $host | cut -d: -f1)"
+    ca_cert="$(openssl s_client -showcerts -servername ${servername} -connect ${host} </dev/null | openssl x509 | base64 -w 0)"
+
+    ca_cert="${ca_cert}" yq -i '(select(.stringData.kubeconfig != null)) .stringData.kubeconfig |= (fromyaml | .clusters[].cluster."certificate-authority-data" = env(ca_cert) | to_yaml)' "${manifest}"
+    # name="$(KUBECONFIG=${config} kubectl config view --minify | yq .clusters[0].name)"
```
Roming22, line 236: Why keep those commented bits in?

images/kcp-registrar/register.sh
```diff
@@ -214,12 +214,30 @@ register() {
                 --syncer-image "ghcr.io/kcp-dev/kcp/syncer:$KCP_SYNC_TAG" \
                 --resources deployments.apps,services,ingresses.networking.k8s.io,pipelines.tekton.dev,pipelineruns.tekton.dev,tasks.tekton.dev,runs.tekton.dev,networkpolicies.networking.k8s.io \
                 --output-file "$syncer_manifest"
+            add_ca_cert_to_syncer_manifest "${kcp_kcfg}" "$syncer_manifest"
```
Roming22, line 217: That seems strange to me, we should not have to mess with the manifest generated by kcp to get a secure connection.
If adding a ca_cert is helpful/requried, it seems that this should be implemented on kcp side.

images/kcp-registrar/register.sh
lcarva: Apologies. It's leftover from previous iteration. Removed.

images/kcp-registrar/register.sh
lcarva: I think that's true only if the underlying cluster itself uses a valid identity cert which is not the case in most development environments.

Roming22: Any chance the certificate we're trying to extract might not exists? I don't see any error handling around that, which is fine with me if we know that the certificate should always be there on a properly configured cluster.

lcarva: I think we can assume that it'll always be there because we're querying the OpenShift API which is always served via HTTPS. If a cert does not exist, the cluster is likely non-functional.

lcarva:
Failures appear to be unrelated to the change?

Roming22:
Shellcheck failure is related to the change and mentions the line that generate the error.

Roming22:
For the ckcp-test, please make sure you have rebased against `main`. If you have rebased and still see the issue, please contact @xinredhat to get support on the CI.

bnallapeta:
@lcarva Here is the snippet from the [logs](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-pipelines_pipeline-service/245/pull-ci-openshift-pipelines-pipeline-service-main-plnsvc-shellcheck/1572200368098512896). Please fix this.

```
In ./images/kcp-registrar/register.sh line 233:
    ca_cert="$(openssl s_client -showcerts -servername ${servername} -connect ${host} </dev/null | openssl x509 | base64 -w 0)"
                                                       ^-----------^ SC2086: Double quote to prevent globbing and word splitting.
                                                                              ^-----^ SC2086: Double quote to prevent globbing and word splitting.
Did you mean: 
    ca_cert="$(openssl s_client -showcerts -servername "${servername}" -connect "${host}" </dev/null | openssl x509 | base64 -w 0)"

```

Roming22:
@lcarva Your PR now has conflicts, please rebase when you fix the shellcheck issue.

Roming22 (approved):

PR #122, opened 2022-06-24
Remove TaskRuns syncing from kcp workspace

avinal:
## Description 

- remove task and taskrun sync
- add TektonConfig to only prune TaskRuns
- update related documentations

## Result

- This will enable clearing of TaskRun resources from the workload clusters. 
- As of now removing TaskRuns from workload cluster results in reconciling of the resources and the resources get recreated. 

Signed-off-by: Avinal Kumar <avinkuma@redhat.com>

PR #246, opened 2022-09-16
Sync TaskRuns, not Runs

lcarva:
TaskRun is to a Task what PipelineRun is to a Pipeline [0]. A Run resource is used to execute a CustomTask[1].

We may want Run resources in the future, but for now, let's stick with the basic Tekton resources.

[0] https://tekton.dev/docs/pipelines/taskruns
[1] https://tekton.dev/docs/pipelines/runs/

Signed-off-by: Luiz Carvalho <lucarval@redhat.com>

### DIFF
--- a/ckcp/README.md
+++ b/ckcp/README.md
@@ -93,7 +93,7 @@ locations                                      scheduling.kcp.dev/v1alpha1      
 placements                                     scheduling.kcp.dev/v1alpha1            false        Placement
 <br>pipelineruns                      pr,prs       tekton.dev/v1beta1                     true         PipelineRun</br>
 <br>pipelines                                      tekton.dev/v1beta1                     true         Pipeline</br>
-<br>runs                                           tekton.dev/v1alpha1                    true         Run</br>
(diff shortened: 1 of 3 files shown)


avinal:
Hi @lcarva, we have intentionally removed the syncing of TaskRuns because of pruning issues. Please see #122 and https://github.com/openshift-pipelines/pipeline-service/blob/main/docs/phase1_limitations.md#pruning-of-pipelineruns-and-taskruns. Until we have a kcp aware pruning mechanism this must be kept this way.

lcarva:
> Hi @lcarva, we have intentionally removed the syncing of TaskRuns because of pruning issues. Please see #122 and https://github.com/openshift-pipelines/pipeline-service/blob/main/docs/phase1_limitations.md#pruning-of-pipelineruns-and-taskruns. Until we have a kcp aware pruning mechanism this must be kept this way.

Aha! Thanks for the pointer. What about Runs? Do we actually want to sync those?

PR #256, opened 2022-09-23
Provide CRs to sync as parameter in register script

Mo3m3n:
Custom resources to sync with kcp can now be provided as a parameter for register.sh script.
Also CR_TO_SYNC was renamed to CRS_TO_SYNC since we are dealing with multiple CRs

### DIFF
--- a/ckcp/openshift_dev_setup.sh
+++ b/ckcp/openshift_dev_setup.sh
@@ -105,9 +105,9 @@ init() {
   GIT_REF=$(yq '.git_ref // "main"' "$CONFIG")
 
   # get list of CRs to sync
-  read -ra CR_TO_SYNC <<< "$(yq eval '.cr_to_sync | join(" ")' "$CONFIG")"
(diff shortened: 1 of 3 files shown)


config/config.yaml
```diff
@@ -12,7 +12,7 @@ git_url: https://github.com/openshift-pipelines/pipeline-service.git
 git_ref: main
 
 # cr_to_sync is a list of the CRs which need to be synced.
```
Roming22, line 14: You forgot to update that comment

ckcp/openshift_dev_setup.sh
```diff
@@ -363,13 +363,16 @@ install_pipeline_service() {
 }
 
 register_compute() {
+  resources="$(printf '%s,' "${CRS_TO_SYNC[@]}")"
```
Roming22, line 366: I'm not familiar with the syntax of these 2 lines, can you please explain what they do?

Roming22 (approved):

PR #260, opened 2022-09-28
WIP: Openshift 4.11 Pod security standards

CathalOConnorRH:
Created new SCC to match baseline pod security standards by setting allowPrivilegeEscalation: false and updating clusterrole to use new SCC. This is required from Openshift 4.11 onwards.

### DIFF
--- a/gitops/argocd/tektoncd/clusterrole.yaml
+++ b/gitops/argocd/tektoncd/clusterrole.yaml
@@ -1,0 +1,14 @@
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: pipelines-scc-clusterrole
+rules:
+  - verbs:
+      - use
+    apiGroups:
+      - security.openshift.io
+    resources:
+      - securitycontextconstraints
+    resourceNames:
+      - pipelines-scc-v2
--- a/gitops/argocd/tektoncd/kustomization.yaml
+++ b/gitops/argocd/tektoncd/kustomization.yaml
@@ -5,3 +5,5 @@ resources:
   - allow-argocd-to-manage.yaml
   - openshift-operator.yaml
   - tekton-config.yaml
+  - clusterrole.yaml
+  - pipelines-scc-v2.yaml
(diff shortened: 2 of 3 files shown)


bnallapeta:
@CathalOConnorRH @Mo3m3n has already worked on implementing this in his PR #241 and there is a discussion going on [there](https://github.com/openshift-pipelines/pipeline-service/pull/241#discussion_r978615250). Could you please check that and see if this PR is bringing other changes than what he submitted there?

CathalOConnorRH:
> @CathalOConnorRH @Mo3m3n has already worked on implementing this in his PR #241 and there is a discussion going on [there](https://github.com/openshift-pipelines/pipeline-service/pull/241#discussion_r978615250). Could you please check that and see if this PR is bringing other changes than what he submitted there?

I believe these are different, @Mo3m3n is implementing pod security on ckcp deployment, this PR is implementing pod security on the pods created by the pipeline runs which will be on the workload clusters

Mo3m3n:
@CathalOConnorRH Shouldn't this rather be handled by the Openshift pipelines operator ?
There are multiple applications deployed in this project, some of them via operators (pipelines, gitops, etc) others directly by our scripts (kcp pod, syncer pod, etc).
My understanding is that it is the scope of the operator to make sure deployments runs in the cluster.
One way to double check this is to try running Openshift pipelines operator in 4.11 and see we get any warnings about restricted security policy.

CathalOConnorRH:
@Mo3m3n  There is an upstream story opened for this for v1.9 but no release date set yet. This will secure pipelines sooner and allow us to resolve any issues we may encounter.

Roming22:
Setting as draft, as my understanding is that this PR might not be merged.

adambkaplan (commented):
@CathalOConnorRH the item that's missing here is the role binding.

Looking at what the Tekton operator does, it creates a `RoleBinding` in every namespace to go along with the `pipelines` service account. It seems that we would need to replicate this logic in the workspace controller.

Roming22:
@CathalOConnorRH has this PR gone stale? If it is not needed anymore please close it. If it is still relevant please respond to the feedback and rebase the changes.

bnallapeta:
@CathalOConnorRH Could you check if this PR is relevant anymore with the changes to the repo? If not, please close it.

PR #263, opened 2022-09-29
PLNSRVCE-671 : Automate the process to upgrade pipeline-service on staging cluster

bnallapeta:
Context:
Gitlab CI uses cluster-setup and kcp-registrar images with 'latest' tag at the moment. We want to provide the latest commit SHA as the tags instead of latest. I am working on automating this process where we always fetch the last commit ID and provide that as the tag in Gitlab. Once this is merged, it will trigger a CI to update our staging cluster with the latest code.

Problem:
Now, the Github action build-push-images.yaml is based on path filtering. It would only build and create new images only when there are changes to that image content. Otherwise, it will just have a older commit tag as the latest. So there is a possibility of non-existent image tag which Gitlab will try to fetch and fail. 

Solution:
This PR that **does not** build a new image if there are no changes but **only tags** it with the latest commit ID.

Uploading images with latest tags to quay.io
- If the image has been built and pushed to quay, then it won't tag it again and push
- But if the image is not built, then this change will make sure that the image is tagged with the latest commit and push it to quay.

This goes along with the PR raised on Gitlab (this one needs to be merged first).

### DIFF
--- a/.github/workflows/build-push-images.yaml
+++ b/.github/workflows/build-push-images.yaml
@@ -68,7 +68,9 @@ jobs:
           password: ${{ secrets.QUAY_TOKEN }}
       - name: Print image url
         if: steps.filter.outputs.access-setup == 'true'
-        run: echo "Image pushed to ${{ steps.push-to-quay-access-setup.outputs.registry-paths }}"
+        run: |
+          echo "Image pushed to ${{ steps.push-to-quay-access-setup.outputs.registry-paths }}"
+          echo "::set-output name=access_setup_pushed::true"
 
       # Build and push cluster-setup image, tagged with latest and the commit SHA.
       - name: Build cluster-setup Image
@@ -93,7 +95,9 @@ jobs:
           password: ${{ secrets.QUAY_TOKEN }}
       - name: Print image url
         if: steps.filter.outputs.cluster-setup == 'true'
-        run: echo "Image pushed to ${{ steps.push-to-quay-cluster-setup.outputs.registry-paths }}"
+        run: |
+          echo "Image pushed to ${{ steps.push-to-quay-cluster-setup.outputs.registry-paths }}"
+          echo "::set-output name=cluster_setup_pushed::true"
 
       # Build and push gateway-deployment image, tagged with latest and the commit SHA.
       - name: Build gateway-deployment Image
@@ -118,7 +122,9 @@ jobs:
           password: ${{ secrets.QUAY_TOKEN }}
       - name: Print image url
         if: steps.filter.outputs.gateway-deployment == 'true'
-        run: echo "Image pushed to ${{ steps.push-to-quay-gateway-deployment.outputs.registry-paths }}"
+        run: |
+          echo "Image pushed to ${{ steps.push-to-quay-gateway-deployment.outputs.registry-paths }}"
+          echo "::set-output name=gateway_deployment_pushed::true"
 
       # Build and push kcp-registrar image, tagged with latest and the commit SHA.
       - name: Build kcp-registrar Image
@@ -145,7 +151,9 @@ jobs:
           password: ${{ secrets.QUAY_TOKEN }}
       - name: Print image url
         if: steps.filter.outputs.kcp-registrar == 'true'
-        run: echo "Image pushed to ${{ steps.push-to-quay-kcp-registrar.outputs.registry-paths }}"
+        run: |
+          echo "Image pushed to ${{ steps.push-to-quay-kcp-registrar.outputs.registry-paths }}"
+          echo "::set-output name=kcp_registrar_pushed::true"
 
       # Build and push devenv image, tagged with the branch name and the commit SHA.
       - name: Build devenv Image
@@ -170,7 +178,9 @@ jobs:
           password: ${{ secrets.QUAY_TOKEN }}
       - name: Print image url
         if: steps.filter.outputs.devenv == 'true'
-        run: echo "Image pushed to ${{ steps.push-to-quay-devenv.outputs.registry-paths }}"
+        run: |
+          echo "Image pushed to ${{ steps.push-to-quay-devenv.outputs.registry-paths }}"
+          echo "::set-output name=devenv_pushed::true"
 
       # Build and push shellcheck image, tagged with latest and the commit SHA.
       - name: Build shellcheck Image
@@ -195,7 +205,9 @@ jobs:
           password: ${{ secrets.QUAY_TOKEN }}
       - name: Print image url
         if: steps.filter.outputs.shellcheck == 'true'
-        run: echo "Image pushed to ${{ steps.push-to-quay-shellcheck.outputs.registry-paths }}"
+        run: |
+          echo "Image pushed to ${{ steps.push-to-quay-shellcheck.outputs.registry-paths }}"
+          echo "::set-output name=shellcheck_pushed::true"
 
       # Build and push vulnerability-scan image, tagged with latest and the commit SHA.
       - name: Build vulnerability Image
@@ -220,4 +232,84 @@ jobs:
           password: ${{ secrets.QUAY_TOKEN }}
       - name: Print image url
         if: steps.filter.outputs.vulnerability == 'true'
-        run: echo "Image pushed to ${{ steps.push-to-quay-vulnerability-scan.outputs.registry-paths }}"
+        run: |
+          echo "Image pushed to ${{ steps.push-to-quay-vulnerability-scan.outputs.registry-paths }}"
+          echo "::set-output name=vulnerability_scan_pushed::true"
+
+      # Tag images with the latest commit ID if they are not changed.
+      - name: Tag latest commit ID to quay.io
+        id: tag-commit-quay-access-setup
+        if: steps.build-image-access-setup.outputs.access_setup_pushed != 'true'
+        env:
+          image: access-setup
+          registry: quay.io/redhat-pipeline-service
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_TOKEN }}
+        run: |
+          ./images/quay-upload/image-upload.sh
+
+      - name: Tag latest commit ID to quay.io
+        id: tag-commit-quay-cluster-setup
+        if: steps.build-image-cluster-setup.outputs.cluster_setup_pushed != 'true'
+        env:
+          image: cluster-setup
+          registry: quay.io/redhat-pipeline-service
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_TOKEN }}
+        run: |
+          ./images/quay-upload/image-upload.sh
+
+      - name: Tag latest commit ID to quay.io
+        id: tag-commit-quay-gateway-deployment
+        if: steps.build-image-gateway-deployment.outputs.gateway_deployment_pushed != 'true'
+        env:
+          image: gateway-deployment
+          registry: quay.io/redhat-pipeline-service
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_TOKEN }}
+        run: |
+          ./images/quay-upload/image-upload.sh
+
+      - name: Tag latest commit ID to quay.io
+        id: tag-commit-quay-kcp-registrar
+        if: steps.build-image-access-setup.outputs.kcp_registrar_pushed != 'true'
+        env:
+          image: kcp-registrar
+          registry: quay.io/redhat-pipeline-service
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_TOKEN }}
+        run: |
+          ./images/quay-upload/image-upload.sh
+
+      - name: Tag latest commit ID to quay.io
+        id: tag-commit-quay-devenv
+        if: steps.build-image-access-setup.outputs.devenv_pushed != 'true'
+        env:
+          image: devenv
+          registry: quay.io/redhat-pipeline-service
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_TOKEN }}
+        run: |
+          ./images/quay-upload/image-upload.sh
+
+      - name: Tag latest commit ID to quay.io
+        id: tag-commit-quay-shellcheck
+        if: steps.build-image-access-setup.outputs.shellcheck_pushed != 'true'
+        env:
+          image: shellcheck
+          registry: quay.io/redhat-pipeline-service
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_TOKEN }}
+        run: |
+          ./images/quay-upload/image-upload.sh
+
+      - name: Tag latest commit ID to quay.io
+        id: tag-commit-quay-vulnerability-scan
+        if: steps.build-image-vulnerability-scan.outputs.vulnerability_scan_pushed != 'true'
+        env:
+          image: vulnerability-scan
+          registry: quay.io/redhat-pipeline-service
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_TOKEN }}
+        run: |
+          ./images/quay-upload/image-upload.sh
--- a/.github/workflows/build-push-images.yaml
+++ b/.github/workflows/build-push-images.yaml
@@ -34,6 +34,12 @@ jobs:
             kcp-registrar:
               - '.github/workflows/build-push-images.yaml'
               - 'images/kcp-registrar/**'
+            quay-upload:
+              - '.github/workflows/build-push-images.yaml'
+              - 'test/images/quay-upload/**'
+            update-pipeline-service:
+              - '.github/workflows/build-push-images.yaml'
+              - 'images/update-pipeline-service/**'
             devenv:
               - '.github/workflows/build-push-images.yaml'
               - 'config/dependencies.yaml'
@@ -70,7 +76,16 @@ jobs:
         if: steps.filter.outputs.access-setup == 'true'
         run: |
           echo "Image pushed to ${{ steps.push-to-quay-access-setup.outputs.registry-paths }}"
-          echo "::set-output name=access_setup_pushed::true"
+      - name: Tag latest commit ID to quay.io
+        id: tag-commit-quay-access-setup
+        if: steps.filter.outputs.access-setup != 'true'
+        env:
+          image: access-setup
+          registry: quay.io/redhat-pipeline-service
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_TOKEN }}
+        run: |
+          ./test/images/quay-upload/image-upload.sh
 
       # Build and push cluster-setup image, tagged with latest and the commit SHA.
       - name: Build cluster-setup Image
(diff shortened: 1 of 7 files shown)


images/quay-upload/image-upload.sh
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
```
Roming22, line 1: I would this image to `test`, as strictly speaking it is not an image use to operate pipeline-service, but used during development (on the other hand, if we put the image to update the gitops repo in this repo, it is part of operations and would be in `images`).

images/quay-upload/image-upload.sh
```diff
@@ -0,0 +1,74 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# fetching values from env vars
```
Roming22, line 21: I would put that in an `init` function and make sure that the values are set.

.github/workflows/build-push-images.yaml
```diff
@@ -220,4 +232,84 @@ jobs:
           password: ${{ secrets.QUAY_TOKEN }}
       - name: Print image url
         if: steps.filter.outputs.vulnerability == 'true'
-        run: echo "Image pushed to ${{ steps.push-to-quay-vulnerability-scan.outputs.registry-paths }}"
+        run: |
+          echo "Image pushed to ${{ steps.push-to-quay-vulnerability-scan.outputs.registry-paths }}"
+          echo "::set-output name=vulnerability_scan_pushed::true"
+
+      # Tag images with the latest commit ID if they are not changed.
+      - name: Tag latest commit ID to quay.io
```
Mo3m3n, line 240: It would be cleaner to keep the steps ordered by image context, example:
- Build access-setup Image
- Push to quay.io
- Pint image url
- Tag latest commit ID to quay.io

instead of grouping all the "Tag latest commit ID to quay.io" at the end

.github/workflows/build-push-images.yaml
```diff
@@ -220,4 +232,84 @@ jobs:
           password: ${{ secrets.QUAY_TOKEN }}
       - name: Print image url
         if: steps.filter.outputs.vulnerability == 'true'
-        run: echo "Image pushed to ${{ steps.push-to-quay-vulnerability-scan.outputs.registry-paths }}"
+        run: |
+          echo "Image pushed to ${{ steps.push-to-quay-vulnerability-scan.outputs.registry-paths }}"
+          echo "::set-output name=vulnerability_scan_pushed::true"
```
Mo3m3n, line 237: Is the second output var (example `steps.build-image-access-setup.outputs.vulnerability_scan_pushed`) necessary? 
Can't we reuse `steps.filter.outputs.vulnerability` ?
(same for the rest of the builds)

Mo3m3n (commented):
It is not clear to my why we want to retag all images with latest commit from main even-though the images themselves did not change. 
Can you provide more context about this (or even better in the commit message body)?

images/quay-upload/image-upload.sh
```diff
@@ -0,0 +1,74 @@
...
+}
+
+main() {
+  fetch_commits
+  tag_and_push
+}
+
+if [ "${BASH_SOURCE[0]}" == "$0" ]; then
+  main "$@"
+fi
```
ramessesii2, line 82: A new line here.

images/quay-upload/image-upload.sh
```diff
@@ -0,0 +1,74 @@
...
+  image=$(podman images "$image_path":latest --format json | jq '.[0].Names')
+
+  if [[ "$image" == "null" ]]; then
+    printf "Image was not pulled due to some issue. Exiting.\n"
+    exit 1
+  else
+    printf "Image pull was successful.\n"
+  fi
+
+  podman tag "$image_path":latest "$image_path":"$latest_commit"
```
ramessesii2, line 70: Would it be a good idea to have `$latest_commit` as a global variable? Since it's only defined in different function `fetch_commits`.

images/quay-upload/image-upload.sh
bnallapeta: As per the discussion on Gitlab, I moved it here.

images/quay-upload/image-upload.sh
bnallapeta: Done.

.github/workflows/build-push-images.yaml
bnallapeta: All the steps mentioned are ordered except for the last one as it is not certain that it will be executed each time. That's why I kept these at the end. Let me know if you still feel its better grouped at the top. I shall update it.

.github/workflows/build-push-images.yaml
bnallapeta: Because, unless the path is changed for a particular image and the step is executed, steps.filter.outputs.vulnerability will be non-existent.

Mo3m3n: let's take this example
```
- name: Print image url
        if: steps.filter.outputs.vulnerability == 'true'
        run: |
          echo "Image pushed to ${{ steps.push-to-quay-vulnerability-scan.outputs.registry-paths }}"
          echo "::set-output name=vulnerability_scan_pushed::true"
```
If `steps.filter.outputs.vulnerability` does not exist (path did not change for this image) then `vulnerability_scan_pushed` won't exist either no ?

.github/workflows/build-push-images.yaml
```diff
@@ -220,4 +232,84 @@ jobs:
...
+          image: shellcheck
+          registry: quay.io/redhat-pipeline-service
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_TOKEN }}
+        run: |
+          ./images/quay-upload/image-upload.sh
+
+      - name: Tag latest commit ID to quay.io
+        id: tag-commit-quay-vulnerability-scan
+        if: steps.build-image-vulnerability-scan.outputs.vulnerability_scan_pushed != 'true'
```
Mo3m3n, line 308: `vulnerability_scan_pushed` is defined in the "Print image url" step
But here you are referring to output from "Build vulnerability Image" step. 
You have tested it and it works this way ?
(same for all steps)

.github/workflows/build-push-images.yaml
Roming22: I agree with Moemen, just because it will make copy pasting the actions for a new image easier (no cherry-picking from various part of the file).

images/quay-upload/image-upload.sh
```diff
@@ -0,0 +1,82 @@
...
+init() {
+  # fetching values from env vars
+  username="$username"
+  password="$password"
+  registry="$registry"
+  image="$image"
+  image_path="$registry"/"$image"
+
+  if [[ -z $username || -z $password || -z $registry || -z $image ]]; then
+    printf "Error while fetching one of more env variables. Exiting.\n"
```
Roming22, line 30: Errors should be sent to stderr.

images/quay-upload/image-upload.sh
```diff
@@ -0,0 +1,82 @@
...
+
+init() {
+  # fetching values from env vars
+  username="$username"
+  password="$password"
+  registry="$registry"
+  image="$image"
+  image_path="$registry"/"$image"
+
+  if [[ -z $username || -z $password || -z $registry || -z $image ]]; then
```
Roming22, line 29: Shellcheck will probably complain about missing quotes. Please check CI.

images/quay-upload/image-upload.sh
```diff
@@ -0,0 +1,82 @@
...
+        printf "Error while fetching the latest commit from GitHub. Status code: %s\n" "${latest_commit_status}" >&2
+        exit 1
+      fi
+    fi
+  done
+}
+
+tag_and_push() {
+  podman login -u="$username" -p="$password" quay.io
+  podman pull -q "$image_path":latest
```
Roming22, line 58: That going to work only as long as we have a single branch. It would be best to use the previous commit instead of `latest`, so that it can work on any number of branch.

images/quay-upload/image-upload.sh
```diff
@@ -0,0 +1,82 @@
...
+  # verify that the image is actually pulled
+
+  image=$(podman images "$image_path":latest --format json | jq '.[0].Names')
+
+  if [[ "$image" == "null" ]]; then
+    printf "Image was not pulled due to some issue. Exiting.\n"
```
Roming22, line 64: Redirect to stderr

images/update_kcp_stable/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+WORKDIR /
+RUN mkdir /workspace && chmod 777 /workspace && chown 65532:65532 /workspace
+ENV HOME /tmp/home
+RUN mkdir $HOME && chmod 777 $HOME && chown 65532:65532 $HOME
+RUN YQ_VERSION="v4.24.5" && \
+    curl --fail -sSL -o "/usr/local/bin/yq" "https://github.com/mikefarah/yq/releases/download/$YQ_VERSION/yq_linux_amd64" && \
+    chmod 755 "/usr/local/bin/yq"
+RUN JQ_VERSION="1.6" && \
+    curl -sSL -o "/usr/local/bin/jq" "https://github.com/stedolan/jq/releases/download/jq-$JQ_VERSION/jq-linux64" && \
+    chmod +755 "/usr/local/bin/jq"
```
Roming22, line 26: `+755` -> `755`. Also please check hadolint ci results, it seems to be unhappy about something.

images/quay-upload/image-upload.sh
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env bash
```
Roming22, line 1: This is just a script, not an image. It should be moved someplace else, and in a folder that makes it clear that it not used for operating the service.

.github/workflows/build-push-images.yaml
```diff
@@ -220,4 +232,84 @@ jobs:
...
+        env:
+          image: vulnerability-scan
+          registry: quay.io/redhat-pipeline-service
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_TOKEN }}
+        run: |
+          ./images/quay-upload/image-upload.sh
```
Roming22, line 315: You're missing the steps for the new image you've created.

images/update_kcp_stable/Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
```
Roming22, line 1: I would rename the image to `update_pipeline_service`. `kcp_stable` is the name of our repo.

images/update_kcp_stable/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+LABEL build-date= \
+      com.redhat.build-host= \
+      description="This image provides binaries and a script to update image tags on Gitlab repo" \
```
Roming22, line 4: Do reference gitlab specifically. The description should be generic as this image might handle GitLab or other SCM in the future. E.g. `This image provides the required tooling to update the image tags on the SRE gitops repository`.
Also impact `io.k8s.description` and `summary`.

images/update_kcp_stable/bin/gitlab_open_mr.sh
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bash
+
+declare -r SOURCE_BRANCH="robot/auto_update"
```
Roming22, line 3: There might be other robots that submits update for other things (e.g. kubeconfigs to add/remove clusters, change secrets, etc).
Therefore I would make the name a bit more descriptive, such as `robot/pipeline-service-update`.

images/update_kcp_stable/bin/gitlab_open_mr.sh
```diff
@@ -0,0 +1,65 @@
...
+declare -r TARGET_BRANCH="main"
+declare -r GITLAB_TOKEN="${GITLAB_TOKEN}"
+declare -r gitlab_ci="./.gitlab-ci.yml"
+
+raise_mr_gitlab() {
+  current_commit=$1
+  latest_commit=$2
+  if [[ "$current_commit" == "$latest_commit" ]]; then
+    printf "Latest images already available\n" >&2
```
Roming22, line 12: This one is not an error, no need to redirect to stderr.

images/update_kcp_stable/bin/gitlab_open_mr.sh
```diff
@@ -0,0 +1,65 @@
...
+    exit 0
+  else
```
Roming22, line 14: `else` -> `fi` and unindent the next block.

images/update_kcp_stable/bin/gitlab_open_mr.sh
```diff
@@ -0,0 +1,65 @@
...
+
+        if [  $(echo "$existing_mr" | jq '. | length') == 0 ]; then
+          printf "Raising a new MR\n" >&2
+          resp_http_code=$(curl -k -w '%{http_code}' -o /tmp/raise_pr.json \
+              -H "PRIVATE-TOKEN: $GITLAB_TOKEN" \
+              -H "Content-Type: application/json" \
+              --data-raw "${BODY_MR}" \
+              -X POST "https://gitlab.cee.redhat.com/api/v4/projects/$CI_PROJECT_ID/merge_requests")
+          if [[ "$resp_http_code" == "201" ]]; then
+            printf "PR successfully raised\n" >&2
```
Roming22, line 43: Not an error.

images/update_kcp_stable/bin/gitlab_open_mr.sh
```diff
@@ -0,0 +1,65 @@
...
+            exit 0
+          else
+            printf "Unable to raise a new PR\n" >&2
+            exit 1
+          fi
+        else
+          printf "MR already exists. Pushed the latest tag to the same MR.\n" >&2
```
Roming22, line 50: Not an error.

images/update_kcp_stable/bin/gitlab_open_mr.sh
```diff
@@ -0,0 +1,65 @@
...
+          exit 0
+        fi
+      return
+      else
+        if [[ "$i" -lt 3 ]]; then
+          printf "Unable to commit the file. Retrying...\n" >&2
```
Roming22, line 56: Not an error.

images/update_kcp_stable/bin/update.sh
```diff
@@ -0,0 +1,98 @@
...
+Usage:
+    %s [options]
+
+Fetch the latest commit from Pipeline Service Github repo
+and update the image tags in the Gitlab repo where the config lives.
+
+Optional arguments:
+    -scm, --source-control-management SCM
+        Source Code Management tools like github, gitlab, bitbucket etc.
+        Default: gitlab
```
Roming22, line 30: Not sure it makes sense to have a default value for this.
Also, I would assume that by looking at env variables in the container, the script should be able to find the SCM on its own (e.g. if `GITLAB_PROJECT_ID` exists, then we're in GitLab).

images/update_kcp_stable/bin/update.sh
```diff
@@ -0,0 +1,98 @@
...
+    latest_commit_status=$(
+      curl -sw '%{http_code}' -o /tmp/latest_commit.json \
+      -H "Accept: application/vnd.github.VERSION.sha" \
+      "https://api.github.com/repos/openshift-pipelines/pipeline-service/commits/$TARGET_BRANCH"
+    )
+    if [[ "$latest_commit_status" == "200" ]]; then
+      latest_commit=$(cut -c -7 < /tmp/latest_commit.json)
+    else
+      if [[ "$i" -lt 3 ]]; then
+        printf "Unable to fetch the latest commit. Retrying...\n" >&2
```
Roming22, line 77: Not an error.

images/quay-upload/image-upload.sh
bnallapeta: Actually, it is going to be an image when we move to PaC. For now, I had just written the script. But I will write the Dockerfile as well in this PR.
(just like all other images/scripts under this folder, we would require some modifications with the script to make it work with OpenShift Pipelines)

images/update_kcp_stable/bin/update.sh
bnallapeta: I think its still good to provide an option for a user to mention this. Because, if we need to implement auto-detection of SCM, then we would need to do that for a lot of providers.
Default value -> The reason to mention the default value is because our project is using Gitlab. If we don't provide a default, then developers using this would need to  use the flag every time they use the script.

images/quay-upload/image-upload.sh
bnallapeta: Done. Fetching the latest tag with a GET call to quay.

.github/workflows/build-push-images.yaml
bnallapeta: You're right. I've changed the if condition check if check with `steps.filter.outputs.access-setup`.
Only if the paths are changed would it trigger image build and push. So, if this is not true, I am fetching the latest commit on Git, tagging it and pushing it to Quay.

images/quay-upload/image-upload.sh
Roming22: I would still move this image to `test` (while waiting for the refactoring of the repo we talked we should do) in the meantime, as this image is not used by operators.

images/quay-upload/image-upload.sh
```diff
@@ -0,0 +1,74 @@
...
+    latest_commit_status=$(
+      curl -sw '%{http_code}' -o /tmp/latest_commit.json \
+      -H "Accept: application/vnd.github.VERSION.sha" \
+      "https://api.github.com/repos/openshift-pipelines/pipeline-service/commits/main"
+    )
+    if [[ "$latest_commit_status" == "200" ]]; then
+      latest_commit=$(cut -c -7 < /tmp/latest_commit.json)
+    else
+      if [[ "$i" -lt 3 ]]; then
+        printf "Unable to fetch the latest commit. Retrying...\n" >&2
```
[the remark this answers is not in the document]
Roming22: I still see the redirect. There's another on line 69.

images/update-pipeline-service/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+    curl --fail -sSL -o "/usr/local/bin/yq" "https://github.com/mikefarah/yq/releases/download/$YQ_VERSION/yq_linux_amd64" && \
+    chmod 755 "/usr/local/bin/yq"
+RUN JQ_VERSION="1.6" && \
+    curl -sSL -o "/usr/local/bin/jq" "https://github.com/stedolan/jq/releases/download/jq-$JQ_VERSION/jq-linux64" && \
+    chmod 755 "/usr/local/bin/jq"
+COPY .gitlab-ci.yml /usr/local/bin/.gitlab-ci.yml
+COPY images/update_kcp_stable/bin /usr/local/bin
+RUN chmod 755 /usr/local/bin/gitlab_open_mr.sh
+RUN chmod 755 /usr/local/bin/update.sh
+RUN chmod 755 /usr/local/bin/.gitlab-ci.yml
```
Roming22, line 31: This file is not in the repo, and should not be executable.

images/quay-upload/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
```
Roming22, line 1: This image is not used for operating the service, but for developing the service. Therefore it should be moved to `test/images` for the moment. Hopefully we can refactor the repo in the future to improve the layout.

Roming22:
@bnallapeta There's quite a few CI failures. Please fix them when you're processing the comments.

images/update-pipeline-service/Dockerfile
bnallapeta: Done.

images/quay-upload/Dockerfile
bnallapeta: Done.

images/quay-upload/image-upload.sh
bnallapeta: Done. Moved it to test folder.

images/quay-upload/image-upload.sh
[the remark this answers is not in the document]
bnallapeta: Resolved.

.github/workflows/build-push-images.yaml
bnallapeta: Addressed in the other comment. Resolving this one to avoid multiple threads.

Roming22 (approved):

PR #265, opened 2022-09-30
Print kubeconfig of kcp and cluster at the end of ckcp script

bnallapeta:
A common feedback received on our script is that kcp kubeconfig is not easily visible for users' to get access to. Printing it at the end of the script.

### DIFF
--- a/ckcp/openshift_dev_setup.sh
+++ b/ckcp/openshift_dev_setup.sh
@@ -419,6 +419,13 @@ main() {
   done
   echo [sync]
   register_compute
+  printf "\nUse the below KUBECONFIG to get access to the kcp workspace and compute cluster respectively.\n"
+  printf "KUBECONFIG_KCP: %s\n" "$KUBECONFIG_KCP"
+  printf "KUBECONFIG: %s\n" "$KUBECONFIG"
+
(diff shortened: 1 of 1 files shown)


ramessesii2 (approved):

ckcp/openshift_dev_setup.sh
```diff
@@ -419,6 +419,9 @@ main() {
   done
   echo [sync]
   register_compute
+  printf "\nUse the below KUBECONFIG to get access to the kcp workspace and compute cluster respectively.\n"
+  printf "KUBECONFIG_KCP: %s\n" "$KUBECONFIG_KCP"
+  printf "KUBECONFIG: %s\n" "$KUBECONFIG"
```
Roming22, line 424: Do you think there would be value in proposing the user to set some aliases? E.g.
```
  printf "\nYou can set the following aliases to access to the kcp workspace and compute cluster respectively.\n"
  printf "alias kkcp='KUBECONFIG=%s kubectl'" "$KUBECONFIG_KCP"
  printf "alias kcompute='KUBECONFIG=%s kubectl'" "$KUBECONFIG"
```

bnallapeta: Good idea. Since not everybody uses alias, I have kept both sets of printf statements.

Roming22 (approved):

PR #270, opened 2022-10-06
Fix kcp-upgrade workflow

Roming22:
The previous workflow would always be successful because it could not discriminate between the script failing and no update being required.

Changes have been implemented to have the workflow fail in case the script fails, and be successul if there's no update required.

### DIFF
--- a/.github/workflows/kcp-upgrade.yaml
+++ b/.github/workflows/kcp-upgrade.yaml
@@ -19,10 +19,11 @@ jobs:
       - name: Check if a new kcp version exists and update the files if it does
         id: fetch-kcp-version
         run: |
-          sh ./images/kcp-upgrade/upgrade.sh
-          echo "::set-output name=NEW_KCP_VERSION::$(curl -s https://api.github.com/repos/kcp-dev/kcp/releases/latest | yq '.tag_name' | sed 's/v//')"
-          echo "::set-output name=NEW_KCP_VERSION_FOUND::true"
-        continue-on-error: true
+          ./images/kcp-upgrade/upgrade.sh
+          if [ -e "/tmp/kcp-upgrade.txt" ]; then
+            echo "::set-output name=NEW_KCP_VERSION::$(cat "/tmp/kcp-upgrade.txt" | sed 's/v//')"
+            echo "::set-output name=NEW_KCP_VERSION_FOUND::true"
+          fi
 
       - name: Create PR if a new kcp version exists
         if: ${{ steps.fetch-kcp-version.outputs.NEW_KCP_VERSION_FOUND == 'true' }}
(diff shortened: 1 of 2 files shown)


Roming22:
@bnallapeta @ramessesii2 See the result and logs of [this action](https://github.com/openshift-pipelines/pipeline-service/actions/runs/3199710102) to see why this PR is required.

.github/workflows/kcp-upgrade.yaml
```diff
@@ -19,10 +19,11 @@ jobs:
       - name: Check if a new kcp version exists and update the files if it does
         id: fetch-kcp-version
         run: |
-          sh ./images/kcp-upgrade/upgrade.sh
-          echo "::set-output name=NEW_KCP_VERSION::$(curl -s https://api.github.com/repos/kcp-dev/kcp/releases/latest | yq '.tag_name' | sed 's/v//')"
-          echo "::set-output name=NEW_KCP_VERSION_FOUND::true"
-        continue-on-error: true
+          ./images/kcp-upgrade/upgrade.sh
+          if [ -e "/tmp/kcp-upgrade.txt" ]; do
```
ramessesii2, line 23: I've not seen this syntax before - if->do->done. Shouldn't it be if->then->fi?

ramessesii2 (commented):
Although previous intent was to just exit the Workflow gracefully, I never saw this [error](https://github.com/openshift-pipelines/pipeline-service/actions/runs/3199710102/jobs/5225776854) - `./images/kcp-upgrade/upgrade.sh: 19: set: Illegal option -o pipefail`. 
Is the PR fixing above error as well?

.github/workflows/kcp-upgrade.yaml
Roming22: Fixed

Roming22:
@ramessesii2 I don't think it's going to solve the cause of the failure, which is fine with me as I would like to check that the action still exit in error when the script fails on unexpected issues.

.github/workflows/kcp-upgrade.yaml
```diff
@@ -19,10 +19,11 @@ jobs:
...
         run: |
-          sh ./images/kcp-upgrade/upgrade.sh
-          echo "::set-output name=NEW_KCP_VERSION::$(curl -s https://api.github.com/repos/kcp-dev/kcp/releases/latest | yq '.tag_name' | sed 's/v//')"
-          echo "::set-output name=NEW_KCP_VERSION_FOUND::true"
-        continue-on-error: true
+          ./images/kcp-upgrade/upgrade.sh
+          if [ -e "/tmp/kcp-upgrade.txt" ]; then
+            echo "::set-output name=NEW_KCP_VERSION::$(cat "/tmp/kcp-upgrade.txt" | sed 's/v//')"
+            echo "::set-output name=NEW_KCP_VERSION_FOUND::true"
+          done
```
ramessesii2, line 26:
```suggestion
          fi
```

Roming22: Sorry :blush: Fixed.

bnallapeta (approved):

PR #272, opened 2022-10-07
Reverting the use of annotation for kcp glbc as it is breaking syncer

bnallapeta:
Commenting the line which adds an annotation to use kcp glbc feature, as this is causing issues with the syncer being unable to sync the status from synctarget back to the kcp workspace.
This is rendering our service unusable at the moment.
Tracking the kcp issue [here](https://github.com/kcp-dev/kcp/issues/2147).

### DIFF
--- a/images/kcp-registrar/register.sh
+++ b/images/kcp-registrar/register.sh
@@ -220,7 +220,9 @@ register_cluster() {
             --output-file "$syncer_manifest"
         add_ca_cert_to_syncer_manifest "${kcp_kcfg}" "$syncer_manifest"
         # Add annotations required by kcp-glbc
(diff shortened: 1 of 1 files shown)


ramessesii2 (approved):
With this change, I tested, the syncer is able to sync back the status. 
Should we add more info as comment above that commented line before merging this PR? it would be a good reminder before that featuregate is activated once again.

Mo3m3n:
Can you please add context in the commit message or in a comment before the annotation line explaining that this is a kcp bug and provide the link to the issue tracking this

bnallapeta:
@ramessesii2 @Mo3m3n Done.

Roming22 (approved):

PR #278, opened 2022-10-10
Update Dockerfiles to use SHA instead of tags

Roming22:
Using the SHA will enable us to trigger the rebuild of images when new images are made available, especially in the case when the new image solves a vulnerability issue.


bnallapeta (approved):
/lgtm

PR #280, opened 2022-10-10
Refactor image upload

Roming22:
The current script would not upload the right images when there is multiple branches in the repository.

The refactored script also leverage environment variables that are expected to be set to decrease reliance on external services.

### DIFF
--- a/test/images/quay-upload/image-upload.sh
+++ b/test/images/quay-upload/image-upload.sh
@@ -32,68 +32,49 @@ init() {
   fi
 }
 
-fetch_commits() {
-  for i in {1..3}; do
-    latest_commit_status=$(
-      curl -sw '%{http_code}' -o /tmp/latest_commit.json \
-      -H "Accept: application/vnd.github.VERSION.sha" \
-      "https://api.github.com/repos/openshift-pipelines/pipeline-service/commits/main"
-    )
-    if [[ "$latest_commit_status" == "200" ]]; then
-      latest_commit=$(cut -c -7 < /tmp/latest_commit.json)
-    else
-      if [[ "$i" -lt 3 ]]; then
-        printf "Unable to fetch the latest commit. Retrying...\n"
-        sleep 20
-      else
-        printf "Error while fetching the latest commit from GitHub. Status code: %s\n" "${latest_commit_status}" >&2
-        exit 1
-      fi
-    fi
-  done
+get_commit_id() {
+  if [ -n "$GITHUB_SHA" ]; then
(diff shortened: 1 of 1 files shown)


bnallapeta:
> The current script would not upload the right images when there is multiple branches in the repository.

@Roming22 
At the moment, we don't have support to upload images of different branches into quay. We are currently uploading 'main', 'latest' and the latest 'commit ID' as the tags on quay.io. Please see the below image for reference (sha will be the same for all 3 tags).

image


The way I implemented image-upload.sh is that it fetches the tags of an image and grabs the latest one (latest will be main for now). Tomorrow, if we change the way we upload images onto quay.io and now, the latest tags are 'staging' and 'commit ID', then image-upload.sh will fetch whichever is the most recent tag and use that.


The change you've done in this PR is explicitly passing a name of a branch. Since this script is called from [here](https://github.com/openshift-pipelines/pipeline-service/blob/main/.github/workflows/build-push-images.yaml#L87), how do we pass the name of the branch dynamically? If not dynamically and we just fetch the name of the current branch (eg: main), then why do we need to pass this at all? Without this too, the script is fetching the current latest tag from quay.io

Roming22:
> The way I implemented image-upload.sh is that it fetches the tags of an image and grabs the latest one (latest will be main for now). Tomorrow, if we change the way we upload images onto quay.io and now, the latest tags are 'staging' and 'commit ID', then image-upload.sh will fetch whichever is the most recent tag and use that.
Given 2 branches `main` and `release-1.x` pushing images to quay.io. Given a PR merged on main triggering the rebuild of all images. When the next PR is merged on `release-1.x`, if it does not trigger an image build, then the images tagged for this commit will be the one generated by the `main` branch. This is incorrect behavior, and this is what I meant when I said that the current script does not support multiple branches.

> The change you've done in this PR is explicitly passing a name of a branch. Since this script is called from [here](https://github.com/openshift-pipelines/pipeline-service/blob/main/.github/workflows/build-push-images.yaml#L87), how do we pass the name of the branch dynamically? If not dynamically and we just fetch the name of the current branch (eg: main), then why do we need to pass this at all? Without this too, the script is fetching the current latest tag from quay.io
The name of the branch and the commit ID is retrieved from environment variables set by GitHub when running the action (`GITHUB_REF`, `GITHUB_SHA`), so there's no need to specify them in the action itself. Env vars must be specified when running the script manually. If the branch and commit_id cannot be found, no images will be tagged (i.e. there is no "default" behavior).

bnallapeta (approved):
/lgtm

PR #288, opened 2022-10-12
PLNSRVCE-646 : Sync workflows

ramessesii2:
### Changes 
Now, [IndividualImageScannerQuay](https://github.com/openshift-pipelines/pipeline-service/blob/main/.github/workflows/individual-image-scanner-quay.yaml) will run after [BuildPushImage](https://github.com/openshift-pipelines/pipeline-service/blob/main/.github/workflows/individual-image-scanner-quay.yaml) workflow and will be skipped if somehow BuildPushImage fails.
 
### DIFF
--- a/.github/workflows/individual-image-scanner-quay.yaml
+++ b/.github/workflows/individual-image-scanner-quay.yaml
@@ -5,10 +5,18 @@ on:
     branches:
       - main
   workflow_dispatch:
+  workflow_run:
+    workflows:
+      - "Build and push images to quay"
+    branches:
+      - main
+    types:
+      - completed
 
 jobs:
   scans:
     runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
     outputs:
       access-setup-output: ${{ steps.access-setup-scan.outputs.VULNERABILITIES_EXIST }}
       cluster-setup-output: ${{ steps.cluster-setup-scan.outputs.VULNERABILITIES_EXIST }}


.github/workflows/individual-image-scanner-quay.yaml
```diff
@@ -5,10 +5,16 @@ on:
     branches:
       - main
   workflow_dispatch:
+  workflow_run:
+    workflows: [Build and push images to quay]
```
Roming22, line 9: We favor the 
```
list-name:
  - item
```
syntax, as it makes diff more readable.

Roming22 (approved):

.github/workflows/individual-image-scanner-quay.yaml
```diff
@@ -105,7 +113,7 @@ jobs:
   check-results:
     runs-on: ubuntu-latest
     needs: scans
-    if: always()
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
bnallapeta, line 116: The reason for having always() in if is that we need to do the check irrespective of what happens with the previous steps for any one of the image. Eg: access-setup might go through but cluster-setup may not run. This may cause the action to exit without running check-results step.

ramessesii2: The `scans` job would `succeed` even if one of the steps is executed successfully based on path filtering. We will only get a `failure` when one of the steps which were to be executed based on path filtering doesn't go through. Is that correct? In that case, why would we need to do `check-results` because the Workflow would result in an error?

Also, I do see we won't need an `if` statement because `check-results` already has an `needs: scans` statement so it wouldn't run if `scans` job is not there or skipped!

bnallapeta: I still don't see any changes. Commit not pushed?

ramessesii2: I understand that we wanted to revert the `if` statement in `check-results` steps. The change is pushed 😓

bnallapeta (approved):
/lgtm

PR #289, opened 2022-10-12
Upgrade dependencies

Roming22:
Many dependencies are upgraded to the latest version. Fix an issue triggered by the shellcheck upgrade.
Fix whitespace formatting in DEPENDENCIES.yaml.
Add options to install.sh so it can be run by a user in their environment.


Roming22:
I will upgrade deps one by one.

PR #295, opened 2022-10-13
Fix build-push-images.yaml

Roming22:
Wrong env variable is used for the password


Roming22:
@bnallapeta Please review asap.

PR #297, opened 2022-10-14
fix workflow and reset_ckcp.sh

ramessesii2:
## This PR fixes -
* Bug in invoking one of the steps in the `individual-image-scanner-quay.yaml`
* Conceal error when a resource is not found when running `reset_ckcp.sh` using `--ignore-not-found=true`


Roming22 (approved):
Since the 2 changes are unrelated, it would be best to open 2 PRs next time.

PR #301, opened 2022-10-17
deploy gateway in the same workspace as compute

Mo3m3n:
Instead of deploying gateway on a dedicated Workspace (under SRE org), we can have it in the same SRE workspace as compute but in a different namespace.


gitops/kcp/gateway/haproxy-ns.yaml
```diff
@@ -0,0 +1,5 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: gateway
```
Roming22, line 5: The other manifests reference `haproxy` instead of `gateway`. Typo?

images/gateway-deployment/install.sh
```diff
@@ -24,19 +24,11 @@ usage() {
 
     # Parameters
     printf "The following parameters need to be passed to the script:\n"
-    printf "KCP_ORG: the organistation for which the workload clusters need to be registered, i.e.: root:pipeline-service\n"
```
Roming22, line 27: This option might be useful, I would probably keep it, but bypass `switch_org()` if unset.

Roming22 (commented):
Now that this process is out of development, I wonder if it should not be folded with either `access_setup/setup_compute.sh` or `cluster_setup/bin/install.sh` (depending on the permissions required).

That would be a follow up PR, but I think it needs to be discussed.

gitops/kcp/gateway/haproxy-ns.yaml
Mo3m3n: Not a typo, actually the NS is "gateway" and inside the ns the deployment, service, ingress are named "haproxy" since they are specific to haproxy.

images/gateway-deployment/install.sh
```diff
@@ -24,19 +24,11 @@ usage() {
...
     printf "KCP_WORKSPACE: the name of the workspace where the gateway gets deployed (created if it does not exist), i.e: gateway.\n"
```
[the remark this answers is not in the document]
Mo3m3n: Done

Mo3m3n:
> Now that this process is out of development, I wonder if it should not be folded with either access_setup/setup_compute.sh or cluster_setup/bin/install.sh (depending on the permissions required).

> That would be a follow up PR, but I think it needs to be discussed.

@Roming22 both locations may be problematic:
- access-setup is just about setting up early access to the cluster that cannot be done via gitops, once access is enabled gitops tool will handle the rest. I think that we would want gateway config to be enforced by gitops.
- setup-cluster is for now mainly about looping over existing physical cluster to install pipeline-service there.

The only image that currently deals with kcp setup via gitops is kcp-registrar.
Maybe we can have the following layout:
- **access_setup** image which handles access setup for both kcp in pcluster (setup-compute.sh + setup-kcp.sh): nothing to change here
- **cluster-setup** image which needs to be updated to work in a similar way:
  - _compute-setup.sh_ to do what is currently done via "images/cluster-setup/bin/install.sh"
  - _kcp-setup.sh_ which needs to do what "images/kcp-registrar/bin/register.sh" and "images/gateway-deployment/install.sh" are doing

Roming22 (approved):

PR #305, opened 2022-10-18
WIP: Repository refactoring proposal

Roming22:
The repository is now organized based on the actor, so that anyone can easily find the information they are interested in based on the scenario they are looking into.


Mo3m3n:
This new layout is more intuitive.
- This PR should not include any behavior change right ? (no change on scripts, images, etc. It seems to be the case from a first look but I wanted to double check.
- links in the README file need to be checked I think. Example
  - in main README:
  ```
  [here](https://github.com/openshift-pipelines/pipeline-service/tree/main/dev/ckcp)
  ```
  - in developer/ckcp/README:
  ```
   [argo-apps](../gitops/argocd/argo-apps/)
  ```
  - in developer/docs/development.md:
  ```
  [./local/kind/setup.sh](./local/kind/setup.sh)
  ```
- in developer dir we will need a main README we can just move the one in developer/docs/README.md
- I think we don't need the work directory right ? it is only required in the SRE internal repo. If yes then we remove it in a seperate commit.
- This is not related to this PR but I have just noticed it now: in `operator/images/kcp-upgrade/upgrade.sh` you will probably need to fix multiple paths there (to align them with the new layout) but it is really problematic that we need to update kcp versions in so many file and the list may still grow. We should have just one ENV file that can be used anywhere in the project in stead of harcoding the version in many files.

Roming22:
> This new layout is more intuitive.
I'm happy you like it.


> * This PR should not include any behavior change right ? (no change on scripts, images, etc. It seems to be the case from a first look but I wanted to double check.
Correct. Some paths may need to be fixed.

> * links in the README file need to be checked I think.

Correct, that will be done in the "real" PR, this one is just to show the expected layout, and get feedback. It should NOT be merged as is.

> * in developer dir we will need a main README we can just move the one in developer/docs/README.md

As part of the refactoring, I'll make sure every top level folder has an appropriate README.md. That might mean moving/splitting content.

> * I think we don't need the work directory right ? it is only required in the SRE internal repo. If yes then we remove it in a seperate commit.

I don't see a directory named `work` in the PR. Could you please clarify?

> * This is not related to this PR but I have just noticed it now: in `operator/images/kcp-upgrade/upgrade.sh` you will probably need to fix multiple paths there (to align them with the new layout) but it is really problematic that we need to update kcp versions in so many file and the list may still grow. We should have just one ENV file that can be used anywhere in the project in stead of harcoding the version in many files.

I agree with the sentiment and would like us to find a way to move towards a single source of truth for versions in the future. An env file does not seem to bad of solution, since it's easy to pass to the container engine when building an image, and can be sourced in scripts to get the required values.

Mo3m3n:
> I don't see a directory named work in the PR. Could you please clarify?
I Erred , I was looking to a zombie dir in my local repo while doing the review :)

bnallapeta:
@Roming22 
First of all, this is really great! Feels more intuitive to navigate through folders. So thanks for raising this. 

I have only reviewed the structure of the repository and not the contents of each file. I plan to do that with the individual PRs you will raise.

architect
  - If we are having a doc for kcp-registration.md, it makes sense to have one for cluster-setup. So, the flow would go like, (1) what are the components installed on a cluster and  (2) how this cluster is registered with a kcp workspace.
  
developer
  - there is a folder called docs under which there is a file called development.md. But this development.md purely contains instructions that are related to local. Would rename the file to README.md and move it to local folder.
  - The README.md present under docs folder can be removed and placed directly under developer folder.
  - The main README (under developer) should clearly highlight the different dev modes we have - devenv (its an local environment), local (kcp running on your laptop with kind clusters as compute) & ckcp (kcp running in a pod with OpenShift clusters as compute) and should have a hyperlink pointing to the respective folders.


operator
  - The docs folder purely contains examples of how sre should structure their config. It does not have actual documentation. Perhaps renaming it to examples/sre would make more sense?
  - compute/pac-manager - I know your restructuring did not bring this change, but why do we have pac-manager separated into compute in the first place? The same set of manifests already exist under kcp! I would favour keeping it under kcp (or pac?) and removing the compute folder instead.
@Mo3m3n, please provide your comments here as you're working on these components.


shared
  - if the only component using dependencies.yaml is devenv, would prefer if it can be moved into devenv folder.
  - Another reason is that any developer related config can be controlled from config.yaml under developer/ckcp


user
  - I do agree that providing instructions to successfully test pipeline-service are useful, I think it should be placed under Gitops because that's the folder a user would access to setup everything.
  - Since we already have an examples folder under operator, we could create a folder under examples called 'user' or 'test-pipeline-service' and provide these instructions there.

Mo3m3n:
> [compute](https://github.com/openshift-pipelines/pipeline-service/tree/main/gitops/compute/pac-manager)/pac-manager - I know your restructuring did not bring this change, but why do we have pac-manager separated into compute in the first place? The same set of manifests already exist under [kcp](https://github.com/openshift-pipelines/pipeline-service/tree/main/gitops/kcp/pac-manager)! I would favour keeping it under kcp (or pac?) and removing the compute folder instead.
@Mo3m3n, please provide your comments here as you're working on these components.

Because we need to setup a pac-manager user for both kcp and compute. They will be used by gitops-tool/ci to configure the project.
We need to either keep two copies or refactor that in a shared repo but removing one of them will be confusing

Roming22:
First set of PRs have been opened: #313, #314, #315, #316, #317.

Roming22:
All refactoring has been done in other PRs.

Issue #276, opened 2022-10-10
openshift_dev_setup.sh doesn't work on Mac because Podman cannot mount native source folder 


xinredhat:
**Description:**
Podman running on Mac is using remote model. By default bind volume will use source folders on podman server's host(here it is a VM running on Mac as pondman server).  When you init Poman machion with command `podman machine init`, the VM doesn't mount the any native source folder to the VM. The result is that when you try to mount the native source folder  with podman command, it reports  `no such file or directory`. Below is the error I am seeing when executing openshift_dev_setup.sh on Mac.

```
podman run --rm --env COSIGN_PASSWORD=Q9HDartEsayOVxRW --volume /var/folders/s0/jr_x9_mj0sb2np74b8427w180000gn/T/tmp.IgGQXBSs/credentials/manifests/compute/tekton-chains/tmp:/workspace:z --workdir /workspace --entrypoint /usr/bin/cosign quay.io/redhat-appstudio/appstudio-utils:eb94f28fe2d7c182f15e659d0fdb66f87b0b3b6b generate-key-pair
        Error: statfs /var/folders/s0/jr_x9_mj0sb2np74b8427w180000gn/T/tmp.IgGQXBSs/credentials/manifests/compute/tekton-chains/tmp: no such file or directory
```

There is a discussion about it  https://github.com/containers/podman/discussions/13537

adambkaplan:
Since this is using a basic `podman run` command, there's no reason why we shouldn't use `CONTAINER_ENGINE` env var so that it can be swapped with docker.

I ran into a similar issue doing development in a toolbox. In that case, I had to modify one of the scripts to use `flatpak-spawn --host podman` as the container engine.

Roming22:
@adambkaplan @xinredhat Please validate that this issue is fixed. Close the issue if your test is successful.

xinredhat:
When using CONTAINER_ENGINE=docker, this issue is gone on Mac

PR #308, opened 2022-10-19
Use CONTAINER_ENGINE instead of podman

Roming22:
Some developers do not use podman, or podman is causing trouble on their platform (e.g. macOs).
Scripts have been changed to use in order of preference:
- CONTAINER_ENGINE if set by the user
- podman
- docker

Linked to #276

### DIFF
--- a/developer/ckcp/README.md
+++ b/developer/ckcp/README.md
@@ -26,6 +26,7 @@ Before you execute the script, you need:
 4. to install [yq](https://mikefarah.gitbook.io/yq/#install)
 6. to install [kubectl kcp plugin](https://github.com/kcp-dev/kcp/blob/main/docs/kubectl-kcp-plugin.md)
    Note: ckcp uses the official kcp image in order to run kcp in a pod (latest released versions). It is advisable to use the same version for the kcp plugin as the kcp core API (especially as KCP API is evolving quickly). The current version can be found in [DEPENDENCIES.md](../../DEPENDENCIES.md). Make sure to checkout the branch listed in the doc before installing the plugin.
+7. to install [podman](https://github.com/containers/podman) or [docker](https://www.docker.com/). If you have both you can control which is used by setting the `CONTAINER_ENGINE` environment variable (e.g. `export CONTAINER_ENGINE="podman"`). While we do not recommend it for security reasons, you can prefix the binary with `sudo` to force the execution as the root user (e.g. `export CONTAINER_ENGINE="sudo podman"`).
 
 You can run the openshift_dev_setup.sh script with or without parameters as specified below:
 
--- a/ci/images/quay-upload/image-upload.sh
+++ b/ci/images/quay-upload/image-upload.sh
@@ -73,6 +73,29 @@ init() {
   image=${image:-}
 
   image_path="$registry"/"$image"
+
+  detect_container_engine
+}
+
+detect_container_engine() {
+    CONTAINER_ENGINE="${CONTAINER_ENGINE:-}"
+    if [[ -n "${CONTAINER_ENGINE}" ]]; then
+      return
+    fi
+    CONTAINER_ENGINE="podman"
+    if ! command -v podman >/dev/null; then
+        CONTAINER_ENGINE=docker
+        return
+    fi
+    if [[ "$OSTYPE" == "darwin"* && -z "$(podman ps)" ]]; then
+        # Podman machine is not started
+        CONTAINER_ENGINE=docker
+        return
+    fi
+    if [[ "$OSTYPE" == "darwin"* && -z "$(podman system connection ls --format=json)" ]]; then
+        CONTAINER_ENGINE=docker
+        return
+    fi
 }
 
 get_commit_id() {
@@ -99,18 +122,18 @@ pull_push_image() {
   source="$image_path:$branch_name"
   target="$image_path:$commit_id"
 
-  podman login -u="$username" -p="$password" quay.io
+  $CONTAINER_ENGINE login -u="$username" -p="$password" quay.io
 
-  podman pull -q "$source"
+  $CONTAINER_ENGINE pull -q "$source"
   image=$(podman images "$source" --format json | jq '.[0].Names')
   if [[ "$image" == "null" ]]; then
     printf "Image '%s' was not pulled due to some issue. Exiting.\n" "$source" >&2
     exit 1
   fi
 
-  podman tag "$source" "$target"
+  $CONTAINER_ENGINE tag "$source" "$target"
 
-  podman push "$target"
+  $CONTAINER_ENGINE push "$target"
 }
 
 main() {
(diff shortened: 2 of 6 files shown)


Roming22:
@xinredhat How can we add `ALLOW_ROOTLESS` true to the container that runs `plnsvc-ckcp-test`?

images/access-setup/content/bin/setup_compute.sh
```diff
@@ -147,6 +147,33 @@ init() {
...
+  detect_container_engine
+}
+
+detect_container_engine() {
+    CONTAINER_ENGINE="${CONTAINER_ENGINE:-}"
+    ALLOW_ROOTLESS="${ALLOW_ROOTLESS:-false}"
+    if [[ -n "${CONTAINER_ENGINE}" ]]; then
+      return
+    fi
+    CONTAINER_ENGINE="sudo podman"
```
bnallapeta, line 160: Do we need to have root privileges by default? Especially because Tekton Chains was run with just podman run (line 176) without the need for sudo.

images/access-setup/content/bin/setup_compute.sh
```diff
@@ -147,6 +147,33 @@ init() {
   manifests_dir="$WORK_DIR/credentials/manifests"
   mkdir -p "$credentials_dir"
   mkdir -p "$manifests_dir"
+
+  detect_container_engine
+}
+
+detect_container_engine() {
+    CONTAINER_ENGINE="${CONTAINER_ENGINE:-}"
```
bnallapeta, line 155: Do you see value to actually add this as a parameter to the script? Because, currently, user will have to export 'CONTAINER_ENGINE' before running the script. I don't see this documented in a README anywhere and even if we do document it, its a possibility that they might miss it.
Same goes for the other script too.

Roming22: We use that in other scripts. What do you think about documenting that in the DEPENDENCIES.md?

xinredhat:
> @xinredhat How can we add `ALLOW_ROOTLESS` true to the container that runs `plnsvc-ckcp-test`?

sure, this is the PR https://github.com/openshift/release/pull/33317

images/access-setup/content/bin/setup_compute.sh
bnallapeta: 1. I think apart from local/kcp, we are not using sudo anywhere else. (the local script has a need for sudo)
2. I see, but with the base images we are using such as UBI, we haven't come across this issue.

My point basically is that we should avoid using sudo unless we really need to.

images/access-setup/content/bin/setup_compute.sh
bnallapeta: I could only find this being used in local. And its already [documented](https://github.com/openshift-pipelines/pipeline-service/blob/main/DEVELOPMENT.md?plain=1#L31) in development.md.

Given the number of people that would really need this is less, I guess we document the same under ckcp/README.md and leave it be (instead of parameterizing it).

Roming22: I've documented that bit, please close the conversation if that matches your expectations.

images/access-setup/content/bin/setup_compute.sh
Roming22: 1. The same logic is used in `ci/images/quay-upload/image-upload.sh`, `developer/images/devenv/run.sh`, `developer/local/utils.sh`, `images/access-setup/content/bin/setup_compute.sh`.
2. This is not an issue with the base images. This is an issue with systemd in RHEL CSB 8.x that breaks podman in some mysterious ways. Podman never works after a reboot. Sometimes it can be fixed with a `sysctl` command, sometimes it can't be fixed until I reboot the machine.
 
I've added documentation around that bit, please close the conversation if that matches your expectations.

developer/ckcp/README.md
```diff
@@ -26,6 +26,7 @@ Before you execute the script, you need:
 4. to install [yq](https://mikefarah.gitbook.io/yq/#install)
 6. to install [kubectl kcp plugin](https://github.com/kcp-dev/kcp/blob/main/docs/kubectl-kcp-plugin.md)
    Note: ckcp uses the official kcp image in order to run kcp in a pod (latest released versions). It is advisable to use the same version for the kcp plugin as the kcp core API (especially as KCP API is evolving quickly). The current version can be found in [DEPENDENCIES.md](../DEPENDENCIES.md). Make sure to checkout the branch listed in the doc before installing the plugin.
+7. to install [podman](https://github.com/containers/podman) or [docker](https://www.docker.com/). If you have both you control which is used by setting the `CONTAINER_ENGINE` environment variable. If using podman, it is recommended to run `export ALLOW_ROOTLESS=true` to run the containers with your user instead of `root` (note that this is unnecessary if you've already set `CONTAINER_ENGINE` to `podman`).
```
bnallapeta, line 29: > (note that this is unnecessary if you've already set `CONTAINER_ENGINE` to `podman`).
This sentence is a bit confusing. Why would setting export ALLOW_ROOTLESS=true be unnecessary if container_engine is set to podman?

images/access-setup/content/bin/setup_compute.sh
bnallapeta: Thanks @Mo3m3n. It is very nicely summarized!
I would emphasize on the last line you wrote - "needing to use sudo maybe be an exception in certain environments but it should not be the normal way."

Roming22: @bnallapeta 

1. `podman` is now the default instead of `sudo podman`.
2. Remove references to ALLOW_ROOTLESS (except in `developer/local/kind/setup.sh` where I do have to set it to keep the existing behavior).
3. Updated the doc.

Please resolve if it matches your expectations.

developer/ckcp/README.md
Roming22: ALLOW_ROOTLESS dropped from the scripts and doc.

developer/local/utils.sh
```diff
@@ -54,14 +54,10 @@ setupTraps() {
...
 detect_container_engine() {
     CONTAINER_ENGINE="${CONTAINER_ENGINE:-}"
-    ALLOW_ROOTLESS="${ALLOW_ROOTLESS:-false}"
     if [[ -n "${CONTAINER_ENGINE}" ]]; then
       return
     fi
-    CONTAINER_ENGINE="sudo podman"
-    if [[ "${ALLOW_ROOTLESS}" == "true" ]]; then
-        CONTAINER_ENGINE=podman
-    fi
```
bnallapeta, line 64: Since utils.sh is being used by local/kcp, we need to have this block where it uses sudo podman.

images/access-setup/content/bin/setup_compute.sh
bnallapeta: @Roming22 Thank you. I have one comment which I have published. But the PR looks good to me.

developer/local/utils.sh
Roming22: Fixed.

PR #309, opened 2022-10-19
Allow workspace to be configured when deploying ckcp

Roming22:
The workspace in which the service is deployed can now be configured by changing the kcp.workspace value in config/config.yaml.

The structure of the config.yaml file has been modified to improve readability by grouping values together when possible.

### DIFF
--- a/developer/ckcp/README.md
+++ b/developer/ckcp/README.md
@@ -111,7 +111,7 @@ Considerations for testing a new component:-
 3. A user can modify the Custom Resources to be synced by the KCP Syncer by modifying [cr_to_sync](./config.yaml).
 4. Onboarding a new component requires creating a new Argo CD application in [argo-apps](../../operator/gitops/argocd/argo-apps/) and adding it to [kustomization.yaml](../../operator/gitops/argocd/argo-apps/kustomization.yaml).
 5. For testing, users need to modify only the git source path and ref of their Argo CD application to reflect their own Git repo.
-6. A user can also choose a different version of kcp than the one running by default by changing the value of [version.kcp](./config.yaml).
+6. A user can also choose a different version of kcp than the one running by default by changing the value of [kcp.version](./config.yaml).
 
 ### Reset ckcp
 
--- a/developer/ckcp/config.yaml
+++ b/developer/ckcp/config.yaml
@@ -11,24 +11,23 @@ git_url: https://github.com/openshift-pipelines/pipeline-service.git
 # git_ref refers to the git repo's ref to be considered as the source of truth for Argo CD applications.
 git_ref: main
 
-# crs_to_sync is a list of the CRs which need to be synced.
-crs_to_sync:
-  - deployments.apps
-  - services
-  - ingresses.networking.k8s.io
-  - networkpolicies.networking.k8s.io
-  - pipelines.tekton.dev
-  - pipelineruns.tekton.dev
-  - tasks.tekton.dev
-  - repositories.pipelinesascode.tekton.dev
+kcp:
+  # crs_to_sync is a list of the CRs which need to be synced.
+  crs_to_sync:
+    - deployments.apps
+    - services
+    - ingresses.networking.k8s.io
+    - networkpolicies.networking.k8s.io
+    - pipelines.tekton.dev
+    - pipelineruns.tekton.dev
+    - tasks.tekton.dev
+    - repositories.pipelinesascode.tekton.dev
+  version: v0.9.0
+  workspace: root:default:pipeline-service-compute
 
 # Applications to be deployed on the cluster
 apps:
   - pipeline_service # pipeline_service sets up Pipeline Service on the cluster.
-
-# versions
-version:
-  kcp: v0.9.0
 
 # Tekton results database credentials
 tekton_results_db:
--- a/developer/ckcp/openshift_dev_setup.sh
+++ b/developer/ckcp/openshift_dev_setup.sh
@@ -123,19 +123,26 @@ init() {
   GIT_REF=$(yq '.git_ref // "main"' "$CONFIG")
 
   # get list of CRs to sync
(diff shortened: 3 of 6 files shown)


config/config.yaml
```diff
@@ -11,25 +11,25 @@ git_url: https://github.com/openshift-pipelines/pipeline-service.git
...
+  crs_to_sync:
+    - deployments.apps
+    - services
+    - ingresses.networking.k8s.io
+    - networkpolicies.networking.k8s.io
+    - pipelines.tekton.dev
+    - pipelineruns.tekton.dev
+    - tasks.tekton.dev
+    - repositories.pipelinesascode.tekton.dev
+  # Override the version from config/dependencies.yaml
```
bnallapeta, line 25: I think it would be less confusing for the user to set values in one file before running ckcp script. By doing this (having to set kcp version from another file but rest of the values from here), it will only cause confusion from the user's perspective.

ckcp/openshift_dev_setup.sh
```diff
@@ -146,6 +154,15 @@ init() {
   fi
   cp -rf "$GITOPS_DIR/sre" "$WORK_DIR"
   update_git_reference "$GIT_URL" "$GIT_REF" "$WORK_DIR/environment/kcp/kustomization.yaml"
+  {
+    echo "---"
+    kustomize build ckcp/work/environment/kcp/workspace-controller/base | \
```
bnallapeta, line 159: This should not be the way to patch kustomize files. We should try to use overlay/kustomization.yaml to modify values as that is the purpose of overlay folder.
We could have two folders under overlay called dev and staging.
overlays/
  - dev
    - kustomization.yaml
  - staging
    - kustomization.yaml
dev could have the value provided by user.
staging could have the value from our staging cluster.

config/config.yaml
Mo3m3n: I think @bnallapeta point also is that `config/dependencies.yaml` is only used by devenv, so if you comment out the version there will be no place for ops to fetch kcp version.
By the way, and unrelated to this PR, the location of "config/dependencies.yaml" is confusing, it should rather be unedr dev directory

ckcp/openshift_dev_setup.sh
Mo3m3n: @bnallapeta ops(stable/staging) and dev should not share the same WORK_DIR in the first place I think so we don't need dev and staging folders here.
However I agree that using `sed -e` anywhere in this project seems like an anti-pattern to kustomize and we may rather want to patch via kustomize rather than sed.

Mo3m3n: After further looking into kustomize it seems it [does not cope well with vars](https://github.com/kubernetes-sigs/kustomize/issues/4120)

Roming22: I rewrote that bit and moved it to `setup_kcp.sh` as the patching is required for SRE as well. Let's track follow-up feedback on a new comment if necessary.

config/config.yaml
Roming22: I think there's a misunderstanding.

1. The kcp version for the project is set by `config/dependencies.yaml` (e.g. when building images).
2. By default ckcp will use that version.
3. If the user needs to change ckcp behavior, the only file they need to touch is `config/config.yaml`. Maybe my comment is misleading, but what it is trying to say is "if you want ckcp to use a different version, uncomment the line below and define the version you want to use".
4. Regarding the fact that `config/dependencies.yaml` is used only by `devenv`: the goal is to use it (or an upcoming replacement `common/env/dependencies.env`) to be used by anything that requires a version to have a single source of truth.

Does that clear things up for everyone?

bnallapeta: 1. In all of the images, KCP_VERSION is explicitly set to a version - 0.9.0. See [this](https://github.com/openshift-pipelines/pipeline-service/blob/main/images/kcp-registrar/Dockerfile#L34) for example.
2. By default ckcp uses the version from [config/config.yaml](url) file. Not from ckcp/dependencies.yaml. See [this](https://github.com/openshift-pipelines/pipeline-service/blob/main/ckcp/openshift_dev_setup.sh#L20) and [this](https://github.com/openshift-pipelines/pipeline-service/blob/main/ckcp/openshift_dev_setup.sh#L161).
3. This is what I addressed in my initial comment. That the user should only change the version in config.yaml
4. Single source of truth - depends on what is the requirement for it. As of now, there are two files I see that's related to this.
  - config/config.yaml - a yaml where users can modify values that will be used by ckcp.
  - DEPENDENCIES.md - a readme where users can see the version details of every single tool we use.
Why do we need to create another 'single source of truth' file?

ckcp/openshift_dev_setup.sh
bnallapeta: Since the change is only related to ckcp, it makes sense to include the patch in setup_kcp.sh. Thanks.

config/config.yaml
bnallapeta: @Roming22 Could you PTAL at this today? I believe PR 39 on kcp-stable is blocked due to this. 

P.S.: This will probably need a rebase due to the restructuring of our repo now.

Roming22: @bnallapeta Rebase done, and I've removed the bit that was causing trouble. It will come back at a later date ;)

Roming22:
@xinredhat I can't reproduce the error on `plnsvc-ckcp-test` in my dev env using a cluster-bot hypershift cluster. I tried to access the test cluster, but while I can login, I don't seem to have the rights to see anything. Could you please have a look and help me troubleshoot the issue?

bnallapeta (approved):
/lgtm

bnallapeta:
@Roming22 the ckcp-test error was a flaky error which I too have faced earlier. Triggered a retest and looks to be good. Merging the PR.

PR #335, opened 2022-10-28
Modify tekton-chains generated yaml files into a kustomization based structure

bnallapeta:


operator/images/cluster-setup/bin/install.sh
```diff
@@ -108,6 +108,58 @@ switch_cluster() {
...
+fetch_bitwarden_secrets() {
+  CREDENTIALS_DIR="$WORKSPACE_DIR/credentials"
+
+  BW_CLIENTID=${BW_CLIENTID:-}
+  BW_CLIENTSECRET=${BW_CLIENTSECRET:-}
+  BW_PASSWORD=${BW_PASSWORD:-}
+
+  if [ -z "$BW_PASSWORD" ]; then
+    printf "Unset environment variable: \$BW_PASSWORD" >&2
+    return
```
Roming22, line 120: Shouldn't that be `exit 1`?

operator/images/access-setup/content/bin/setup_compute.sh
```diff
@@ -166,57 +166,56 @@ generate_shared_manifests(){
 }
 
 tekton_chains_manifest(){
-  manifest="$manifests_dir/compute/tekton-chains/signing-secrets.yaml"
-  if [ ! -e "$manifest" ]; then
-    manifests_tmp_dir="$(dirname "$manifest")/tmp"
-    mkdir -p "$manifests_tmp_dir"
+  chains_kustomize="$manifests_dir/compute/tekton-chains/kustomization.yaml"
```
Roming22, line 169: My understanding is that:
1. This PR is forcing users to use BW.
2. This PR hard-code some BW keys ID.

Both look wrong to me.

Here's what I was expecting.

-  A config file `credentials/config/bitwarden.yaml` in the gitops repo. This config file both alerts that the operator is using bitwarden to store some/all of their credentials, and store the map of BW-UID to secret type.
- A generic script that process the YAML file, creating the files in expected locations.
- No changes to setup_compute or install.sh, except for adding the pre-processing (or in the case of setup_compute, it might be uploading the secret to BW, and add the ID to the bitwarden.yaml if the value is empty).

Roming22 (commented):
My understanding

Mo3m3n:
- Don't forget to update the docker images if we are going to need the bw client.

operator/images/access-setup/content/bin/setup_compute.sh
bnallapeta: 1. Its not.
2. We need to use the ID of a Bitwarden secret in order to fetch it. That's the only unique identifier for any secrets stored in Bitwarden. And without the login credentials and access to our organization, there's no meaning to those IDs.

ckcp
With this PR, I am not changing the way things are working currently with ckcp. The changes I've made in setup_compute.sh is just changing the format in which the files are generated (from regular .yaml files to a kustomize structure) as it is neater to maintain it and update values dynamically.

kcp-stable
In order to run pipeline-service on our staging cluster with kcp-stable configuration using bitwarden secrets, I've made changes in install.sh where I'm only going to update the files generated earlier (tekton chains and tekton results) **if** BW_PASSWORD is set while calling install.sh (I have a PR for this on Gitlab). If someone else without Bitwarden wants to use it, it would work fine as they would not have set BW_PASSWORD.

operator/images/cluster-setup/bin/install.sh
bnallapeta: We need install.sh to proceed if BW_PASSWORD is not set. (there is the install_clusters func that would run after this)

operator/images/access-setup/content/bin/setup_compute.sh
Mo3m3n: > We need to use the ID of a Bitwarden secret in order to fetch it

Does it have to be hard-coded ? If someone else wants to use our image carrying his own bitwarden secrets they will need to use a different ID I guess.

>  The changes I've made in setup_compute.sh is just changing the format in which the files are generated (from regular .yaml files to a kustomize structure) as it is neater to maintain it and update values dynamically.

I would have put that in a different commit since it is rather a general optimization, I was also thinking this change is necessary for bitwarden to work.

operator/images/cluster-setup/bin/install.sh
Mo3m3n: 1. if BW_PASSWORD is empty we have nothing to print right ?
2. just for esthetic/readability you can remove the else block and use just
```if [ -z "$BW_PASSWORD" ]; then return ;fi

bnallapeta: 1. Right. It might be confusing for an user to read a message about Bitwarden when they don't intend to use it. I have removed the print statement.
2. Done.

operator/images/access-setup/content/bin/setup_compute.sh
bnallapeta: I guess I was thinking only in terms of pipeline service using Bitwarden to hide the secrets. But if we have to generalize it so that others can use it too, I think we need to maintain a separate bitwarden.yaml like Romain has suggested. Let's discuss this before I start implementing it.

Roming22: Discussed this morning. Closing the conversation as we know have a common understanding.

operator/images/cluster-setup/bin/install.sh
Roming22: As discussed, the presence of `credentials/bitwarden.yaml` will trigger a check for the `BW_*` variables.

bnallapeta:
I shall raise another PR with the Bitwarden related changes. This PR is ready to be merged with the current changes.

PR #336, opened 2022-10-28
consumer: Add scripts to access service

Roming22:

### DIFF
--- a/consumer/README.md
+++ b/consumer/README.md
@@ -1,25 +1,55 @@
 # Documentation for consumers of Pipeline Service
 
-This folder holds all the files required to consume Pipeline Service when the service is
-operated by a third party.
+This folder holds all the files required to consume Pipeline Service when the service is operated by a third party.
 
 ## Requirements
 
 * Access to a kcp workspace.
-* Path of the kcp workspace in which the service is deployed (e.g. `root:pipeline-service`).
-This path is provided by the service operators.
+* The following information from the service provider:
+  * Path of the kcp workspace in which the service is deployed (e.g. `root:pipeline-service`).
+  * Name of the APIExport (e.g. `kubernetes`).
 
-## How to bind to the service
+## Quick start guide
 
-* Login on your kcp workspace.
-* Run `./bind_to.sh root:path:to:the:service` to bind to a Pipeline Service provider. The
-exact path will be provided by the service provider.
+This section explains how to use the provided scripts to quickly setup a workspace to use a service provided by an operator of Pipeline Service.
 
-## How to test the service
+If you prefer to do it yourself, or something is not quite working the way it should, see the `Manual setup` section to be walked through the procedure.
 
-* Login on your kcp workspace.
-* Run `./run_test_workload.sh`.
+### Binding to the service
 
-## How to connect a repository to Pipeline as Code
+* Login to your kcp workspace (e.g. `~`).
+* Run `./bind.sh --exportname 'kubernetes' --from 'root:path:to:the:service' --to '~:pipeline-service'`
+  * A new `pipeline-service` workspace will be created under `~`.
+  * The binding to the Pipeline Service provider will be created in the `~:pipeline-service` workspace.
+  * A test will automatically be run to validate that `pipelineruns` get scheduled.
+  * See the `--help` of the command for additional options (e.g. specifying the kubeconfig file).
+
+### Connecting a repository to Pipelines as Code
+
+TODO
+
+## Manual setup
+
+This section is unnecessary if you've already followed the steps of the `Quick start guide`.
+
+### How to bind to the service
+
+To bind to the service:
+* Login to kcp.
+* Go to the workspace in which you want to bind the service.
+* Make a copy of [apibinding.yaml](./manifests/apibinding/apibinding.yaml).
+* Edit `.spec.reference.workspace.path` to reference the service workspace.
+* Edit `.spec.reference.workspace.exportName` to reference the name of the export.
+* Apply the binding with `kubectl apply -f apibindings.yaml`
+
+To validate that the binding is successful:
+* Run `consumer/hack/run_test_workload.sh`
+### How to test the service
+
+To validate that the binding was successful:
+* Login to your kcp workspace.
+* Run `consumer/hack/run_test_workload.sh`.
+
+### How to connect a repository to Pipeline as Code
 
 TODO
--- a/consumer/hack/bind.sh
+++ b/consumer/hack/bind.sh
@@ -1,0 +1,109 @@
+#!/usr/bin/env bash
+
+#quit if exit status of any cmd is a non-zero value
+set -euo pipefail
+
+SCRIPT_DIR="$(
+    cd "$(dirname "$0")" >/dev/null
+    pwd
+)"
+
+# shellcheck source=consumer/hack/run_test_workload.sh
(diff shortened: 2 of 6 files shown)


consumer/README.md
```diff
@@ -6,19 +6,34 @@ operated by a third party.
...
-* Path of the kcp workspace in which the service is deployed (e.g. `root:pipeline-service`).
-This path is provided by the service operators.
+* The following information from the service provider:
+  * Path of the kcp workspace in which the service is deployed (e.g. `root:pipeline-service`).
+  * Name of the APIExport (e.g. `kubernetes`).
 
-## How to bind to the service
+## Quick start guide
 
 * Login on your kcp workspace.
```
bnallapeta, line 15: nitpicking: Login to your kcp workspace.

consumer/README.md
```diff
@@ -6,19 +6,34 @@ operated by a third party.
...
-* Run `./bind_to.sh root:path:to:the:service` to bind to a Pipeline Service provider. The
-exact path will be provided by the service provider.
+* Run `./bind.sh --from root:path:to:the:service --to '~:pipeline-service'` to bind to a Pipeline Service provider.
+  * See the `--help` of the command for additional options (e.g. specifying the kubeconfig file).
+* A test will automatically run to validate that `pipelineruns` get scheduled.
+
+## How to bind to the service
```
bnallapeta, line 20: It is not clear enough that this section "How to bind to the service" and the above section "Quick start guide" are one and the same. To me, they look like steps to be followed one after another. Could you please make it clear to the user that they either need to run the script **or** need to do this step and the next step manually.

consumer/README.md
```diff
@@ -6,19 +6,34 @@ operated by a third party.
...
+
+To bind to the service:
+* Log on kcp.
```
bnallapeta, line 23: nitpicking: Login to kcp.

consumer/README.md
```diff
@@ -6,19 +6,34 @@ operated by a third party.
...
+* Edit `.spec.reference.workspace.path` to reference the service workspace.
+* Edit `.spec.reference.workspace.exportName` to reference the name of the export.
+* Apply the binding with `kubectl apply -f apibindings.yaml`
 
+To validate that the binding is successful:
+* Run `consumer/hack/run_test_workload.sh`
 ## How to test the service
 
+To validate that the binding was successful:
 * Login on your kcp workspace.
```
bnallapeta, line 35: nitpicking: Login to your kcp workspace.

consumer/hack/bind.sh
```diff
@@ -0,0 +1,101 @@
...
+    ${0##*/} [options]
+
+Setup Pipeline Service on a kcp workspace.
+
+Mandatory arguments:
+    -f, --from SERVICE_WORKSPACE
+        Path to the workspace exposing the APIExport.
+
+Optional arguments:
+    -k, --kubeconfig KUBECONFIG
```
bnallapeta, line 26: I think we could think about mandating kubeconfig param too. Because more often than not, the default path for kubeconfig - ~/.kube/config would contain some cluster's kubeconfig and people would have the kcp's kubeconfig stored elsewhere (again, based on the experience with kcp so far, not as a rule). So, it would avoid unnecessary issues if we force the user to provide kubeconfig as manadatory param.

consumer/hack/bind.sh
```diff
@@ -0,0 +1,101 @@
...
+        Path to the kubeconfig with the connection information to the kcp workspace.
+    -t, --to SERVICE_WORKSPACE
```
bnallapeta, line 28: I think this should be WORKSPACE instead of SERVICE_WORKSPACE.
Suggestion: Change it to USER_WORKSPACE to clearly differentiate between service and user's workspaces.

consumer/hack/bind.sh
```diff
@@ -0,0 +1,101 @@
...
+        Workspace in which to bind the pipeline service.
```
bnallapeta, line 29: Suggestion: User workspace in which Pipeline Service would be bound

consumer/hack/bind.sh
```diff
@@ -0,0 +1,101 @@
...
+            shift
+            KUBECONFIG="$1"
+            ;;
+        -f | --from)
+            shift
+            SERVICE_WORKSPACE="$1"
+            ;;
+        -to | --to)
+            shift
+            WORKSPACE="$1"
```
bnallapeta, line 54: Suggestion: Change it to USER_WORKSPACE to clearly differentiate between service and user's workspaces.

Roming22: Done

consumer/hack/bind.sh
Roming22: Reworded, let me know if if is not okay.

consumer/hack/bind.sh
Roming22: Done (I used CONSUMER_WORKSPACE and OPERATOR_WORKSPACE, to reference the same personas we have defined in our repo).

consumer/README.md
Roming22: Done

consumer/README.md
Roming22: Done

consumer/README.md
Roming22: Done

consumer/hack/bind.sh
Roming22: I disagree. The user should not have to specify a path every time they want to use the tool. And with kcp only operators would deal with having to log to kcp and clusters. Everyone else should just be logged in kcp.

If we start to do too much hand holding, there's also the fact that a kubeconfig can contain multiple context, and we don't let the user choose which one either, which could also lead to failure.

What I would be open to do, is check that the current kubeconfig/context points to a kcp instance, and return a nice error if that's not the case.

consumer/README.md
Roming22: Updated. Let me know if further changes are needed.

consumer/hack/bind.sh
Roming22: I've update the code with the new error message. Feel free to reopen the conversation if that solution is not okay with you.

bnallapeta: I wasn't too strong about the idea to begin with and I agree with your point on being too strict/constricted with the experience. 
+1 to the error message you've added to check if its pointing to a kcp workspace or not.

consumer/README.md
```diff
@@ -1,25 +1,55 @@
...
-* Login on your kcp workspace.
-* Run `./run_test_workload.sh`.
+### Binding to the service
 
-## How to connect a repository to Pipeline as Code
+* Login to your kcp workspace (e.g. `~`).
+* Run `./bind.sh --exportname 'kubernetes' --from 'root:path:to:the:service' --to '~:pipeline-service'`
+  * A new `pipeline-service` workspace will be created under `~`.
+  * The binding to the Pipeline Service provider will be created in the `~:pipeline-service` workspace.
+  * A test will automatically be run to validate that `pipelineruns` get scheduled in the `run to validate that `pipelineruns` get scheduled.` workspace.
```
bnallapeta, line 24: Seems like there is a typo in this sentence.

Roming22: Fixed.

bnallapeta:
CI was green earlier. The shellcheck error is fixed and "Pipelines as Code CI / pipeline-service-static-code-analysis" is green. Merging the PR.

PR #346, opened 2022-11-03
Move script to install dependencies to shared/

Roming22:
- Replace dependencies.yaml with dependencies.env: the file is always read within a script so it was changed to a sourceable env file.
- Make install.sh available to all images: the goal is to use this script in all images in the future, so there's a single source of truth.
- Fix a bug in run.sh


bnallapeta (commented):
- dependencies.env file is missing.
- It is difficult to validate devenv each time there are changes to it. Maybe you can think of a test script to validate end to end devenv functionality.

bnallapeta:
@Roming22 

> developer/images/devenv/run.sh should build and run the image. Is there any additional tests you'd like to see?
Got it. Should be good enough.

PR #347, opened 2022-11-04
Capture return status in the function call inline to fix the issue with vulnerability scanners

bnallapeta:
From the logs of a latest run. Instead of the value of res being 0 or 1, the script exits prematurely causing the value to be set to some string. This PR fixes this issue.

```
Run res= Check the full report here: https://quay.io/repository/redhat-pipeline-service/vulnerability-scan/manifest/sha256:92f5cf046[1](https://github.com/openshift-pipelines/pipeline-service/actions/runs/3389846804/jobs/5633369718#step:9:1)bdedb54e66bf2e6a489b16f67ef941aad27cf62f1f25976c29736f?tab=vulnerabilities 
/home/runner/work/_temp/00184d58-34aa-45fb-90ff-56d51ab33496.sh: line 1: Check: command not found
Error: Process completed with exit code [12](https://github.com/openshift-pipelines/pipeline-service/actions/runs/3389846804/jobs/5633369718#step:9:13)7.
```

### DIFF
--- a/ci/images/vulnerability-scan/scan-image.sh
+++ b/ci/images/vulnerability-scan/scan-image.sh
@@ -25,12 +25,11 @@ SCRIPT_DIR="$(
 # shellcheck source=ci/images/vulnerability-scan/common/utils.sh
 source "$SCRIPT_DIR/common/utils.sh"
 
-VULNERABILITIES_EXIST=1
+VULNERABILITIES_EXIST=0
 
 main() {
     get_digest "$IMAGE_NAME"
-    get_vulnerabilities "$IMAGE_NAME"
-    VULNERABILITIES_EXIST=$?
+    get_vulnerabilities "$IMAGE_NAME" || VULNERABILITIES_EXIST=$?
     printf "%s" "$VULNERABILITIES_EXIST"
 }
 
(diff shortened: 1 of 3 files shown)


ci/images/vulnerability-scan/scan-image.sh
```diff
@@ -14,7 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -o errexit
```
Roming22, line 17: I don't think it's a good idea to remove that. This flag ensures that if any command unexpectedly returns a non-0 status, the program will stop instead of ignoring the error.

If you want to ignore a return status on a specific function, you can do the following: `myfunc $MYARG || true`

bnallapeta: Thanks for this suggestion. But using || true will make it always return 0. I'm working on making it actually capture the return status using something like || echo $?. There seems to be some issue. Will update when its done.

bnallapeta: Its done now. I've tested it with a local branch and its working.

Roming22 (approved):

PR #349, opened 2022-11-04
PLNSRVCE-818: Generalize the Github Action "Periodic image scanner" so that other teams can reuse it

bnallapeta:


.github/workflows/periodic-scanner-quay.yaml
```diff
@@ -22,6 +22,8 @@ jobs:
           echo "VULNERABILITIES_EXIST=$(tail -1 /tmp/vulnerability-scan.log)" >> $GITHUB_OUTPUT
         env:
           AUTH_BEARER_TOKEN: ${{ secrets.AUTH_BEARER_TOKEN }}
+          images: ("access-setup" "cluster-setup" "devenv" "kcp-registrar" "quay-upload" "shellcheck" "update-pipeline-service" "vulnerability-scan" "yamllint")
```
Roming22, line 25: Can you make that a list with one item per line? It will avoid having long lines, and will make future diff easier to read.

bnallapeta: Done.

PR #355, opened 2022-11-07
Shellcheck

Roming22:


bnallapeta:
@Roming22 could you take a look at the 'pipeline-service-static-code-analysis' that has failed?

PR #357, opened 2022-11-08
modified timeout value for pipeline and tasks

xinredhat:
This PR is to increase timeout value for pipeline and tasks


ramessesii2 (approved):
LGTM. Although, I wanted to know how do we know if this is the timeout time required? Will it vary according to the on the codebase size?

PR #358, opened 2022-11-08
ckcp can now be run without any app

Roming22:
This change should help deploying ckcp in a standalone manner.

### DIFF
--- a/operator/images/kcp-registrar/bin/register.sh
+++ b/operator/images/kcp-registrar/bin/register.sh
@@ -19,8 +19,8 @@ set -o nounset
 set -o pipefail
 
 SCRIPT_DIR="$(
-  cd "$(dirname "$0")" >/dev/null
-  pwd
+    cd "$(dirname "$0")" >/dev/null
+    pwd
 )"
 
 usage() {
@@ -115,8 +115,8 @@ exit_error() {
     exit 1
 }
 
-function indent () {
-        sed "s/^/$(printf "%$1s")/"
+function indent() {
+    sed "s/^/$(printf "%$1s")/"
 }
 
 prechecks() {
@@ -172,14 +172,20 @@ get_clusters() {
         for sub in "${subs[@]}"; do
             context=$(echo -n "${sub}" | cut -d ',' -f 1)
             cluster=$(echo -n "${sub}" | cut -d ',' -f 2 | cut -d ':' -f 1)
-            if ! (echo "${clusters[@]}" | grep -q "${cluster}") \
-                && (find "$WORKSPACE_DIR/environment/compute" -type d -name "${cluster}" | grep -q "${cluster}" >/dev/null); then
+            if ! (echo "${clusters[@]}" | grep -q "${cluster}") &&
+                (find "$WORKSPACE_DIR/environment/compute" -type d -name "${cluster}" | grep -q "${cluster}" >/dev/null); then
                 clusters+=("${cluster}")
                 contexts+=("${context}")
                 kubeconfigs+=("${kubeconfig}")
+            else
+                printf "WARNING: No configuration found for '%s' (declared in '%s') in '%s'" "$cluster" "$kubeconfig" "$WORKSPACE_DIR/environment/compute" >&2
             fi
         done
     done
+    if [ "${#clusters[@]}" = "0" ]; then
+        printf "No clusters found in '%s'" "$kubeconfig_dir/compute"
+        exit 0
+    fi
 }
 
 switch_org() {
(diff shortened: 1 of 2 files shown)


Roming22:
@bnallapeta This PR would help with the troubleshooting you are doing with the gitops team.

They would need to comment '.apps' and '.crs_to_sync' in 'developper/ckcp/config.yaml'

samyak-jn:
Thanks, @Roming22 for working on this! :")
You are awesome :rocket:

bnallapeta:
@Roming22 Maybe I wasn't clear in our discussion yesterday. The point I wanted to make was "Should ckcp, a dev environment created to use pipeline-service running on kcp be allowed to run just kcp **without** pipeline service itself?"

I believe we should not, because then the intention of ckcp becomes kcp + something on top of it rather than pipeline service in a developer environment. Just to clarify, I'm not saying we should not do it, but if that's what we want to provide with ckcp, then putting that under developer folder in pipeline service would not be the best idea.

Roming22:
> @Roming22 Maybe I wasn't clear in our discussion yesterday. The point I wanted to make was "Should ckcp, a dev environment created to use pipeline-service running on kcp be allowed to run just kcp **without** pipeline service itself?"

Short answer: Yes, if it requires little effort on our end and as long as users know we're on a best effort regarding support.

> I believe we should not, because then the intention of ckcp becomes kcp + something on top of it rather than pipeline service in a developer environment. 

This tool lives in our repository because we were the first one to need it, and it was natural to keep everything together. It is now apparent that other development teams find the tool valuable. To be clear ckcp is indeed 'kcp deployed as a pod in a cluster' with pipeline-service on top. Every refactoring of the script has been to try to better separate each bit.

> Just to clarify, I'm not saying we should not do it, but if that's what we want to provide with ckcp, then putting that under developer folder in pipeline service would not be the best idea.

I agree. Ideally ckcp should live somewhere else. Either within the kcp-dev repository as a dev tool, or in its own repository somewhere. The truth is that I'm not sure we can make that happen right now because we need to focus on the service itself.

fgiloux:
> Ideally ckcp should live somewhere else. Either within the kcp-dev repository as a dev tool, or in its own repository somewhere.
 
 It is there: https://github.com/kcp-dev/helm-charts/tree/main/charts/kcp
 
 and it has been there for a long time. Actually when we refined ckcp we took inspiration from the deployment in the kcp organization. Frankly I don't get why other projects don't want to consume it directly from there if they don't use Pipeline Service.
 
 my 2 cents.

Roming22:
> It is there: https://github.com/kcp-dev/helm-charts/tree/main/charts/kcp

My understanding is that:

- This helm chart only install ckcp itself. Our script "magically" takes care of all the dependencies for the user and generates a kubeconfig so they're ready to login and use it by the time the script completed.
- It install `latest` by default. I'm not familiar with Helm, so I'm not sure if there's a way to override the version. If there is, it's probably more involved than changing the version in `config.yaml`.

My take is that maybe we should be using that helm chart, but someone who wants to get started quickly and painlessly benefits immensely from our deployment script as they don't need to juggle the dependencies, versions and setting up the access.

bnallapeta:
Let me summarize our discussion on this PR.
- Teams starting out to use kcp need a self-hosted kcp in order to experiment with their product on top of kcp.
- kcp although provides a container image, our ckcp module uses that image and sets up a few different things on top.
- We go agree as a team that there is value in providing ckcp module **without** pipeline service. This helps other teams use our module without pipeline service.
- However, we also agreed that the current folder in which ckcp is available is not the ideal location.
- So, this PR can be merged for now and then a later PR would bring in changes to 
  - put this ckcp in another folder at the root of the repository.
  - or consider moving this into a separate repository.

@Roming22 @Mo3m3n @ramessesii2 
If we all think this is okay, let me know and I can press that magic button :)

operator/images/kcp-registrar/bin/register.sh
```diff
@@ -253,61 +259,62 @@ add_ca_cert_to_syncer_manifest() {
...
+        echo "${CRS_TO_SYNC[*]}"
+    )"
+    cr_regexp=$(echo "$cr_regexp" | tr "," \|)
+    readarray -td, crs_to_sync_arr <<<"$CRS_TO_SYNC"
+    declare -p crs_to_sync_arr >/dev/null
+
+    local wait_period=0
+    while [[ "$(KUBECONFIG="${kcp_kcfg}" kubectl api-resources -o name 2>&1 | grep -Ewc "$cr_regexp")" -ne ${#crs_to_sync_arr[@]} ]]; do
+        wait_period=$((wait_period + 10))
+        #when wait_period is reached, print out the CR resources that is not synced to KCP
```
ramessesii2, line 296:
```suggestion
        # when wait_period is reached, print out the CR resources that is not synced to KCP
```

ramessesii2 (approved):
lgtm. Only a minor comment.

operator/images/kcp-registrar/bin/register.sh
Roming22: Done

PR #359, opened 2022-11-08
change SA name from pac-manager

ramessesii2:
##### This PR aims at changing the name of ServiceAccount from `pac-manager` to `pipeline-service-manager`. Since, this SA is used for generating right credentials for accessing services (compute and kcp-workspace), `pac-manager` name doesn't seem be the appropriate name with respect to its function. 


developer/ckcp/openshift_dev_setup.sh
```diff
@@ -345,7 +345,7 @@ patches:
     --kcp-org "$kcp_org" \
     --kcp-workspace "$kcp_workspace" \
     --work-dir "$WORK_DIR" \
-    --kustomization "$GIT_URL/operator/gitops/kcp/pac-manager?ref=$GIT_REF" |
+    --kustomization "$GIT_URL/operator/gitops/kcp/pln-svc-manager?ref=$GIT_REF" |
```
Roming22, line 348: Please change `pln-svc-manager` to `pipeline-service-manager`.

operator/gitops/compute/pln-svc-manager/kustomization.yaml
```diff
@@ -1,6 +1,6 @@
 ---
```
Roming22, line 1: This directory will need to be renamed as well.

Roming22 (commented):
My understanding is that the ckcp test cannot pass because of the renaming of a directory referenced in a kustomization.yaml. Theferore, I'm OK with the change being merged once my comments have been processed.

operator/gitops/compute/pln-svc-manager/kustomization.yaml
ramessesii2: sure, I can change all those instances to `pipeline-service-manager`.

Roming22 (approved):

PR #362, opened 2022-11-09
Add bw cli tool in cluster-setup and devenv images

bnallapeta:


developer/images/devenv/Dockerfile
```diff
@@ -20,8 +20,7 @@ COPY shared /tmp/image-build/shared
 
 WORKDIR /tmp/image-build/shared/hack
 
-RUN set -x \
-    && ./install.sh --bin argocd,grpc_cli,hadolint,jq,kcp,kind,kubectl,oc,shellcheck,tkn,yamllint,yq \
+RUN ./install.sh --debug --bin argocd,grpc_cli,hadolint,jq,kcp,kind,kubectl,oc,shellcheck,tkn,yamllint,yq,bitwarden \
```
Roming22, line 23: Please keep the list ordered.

shared/config/dependencies.sh
```diff
@@ -11,3 +11,4 @@ export SHELLCHECK_VERSION="v0.8.0"
 export TEKTONCD_CLI_VERSION="v0.27.0"
 export YAMLLINT_VERSION="1.28.0"
 export YQ_VERSION="v4.28.1"
+export BITWARDEN_VERSION="cli-v2022.10.0/bw-linux-2022.10.0"
```
Roming22, line 14: That should probably be `v2022.10.0`, and the rest should be part of the curl url.

bnallapeta: Done.

developer/images/devenv/Dockerfile
bnallapeta: Done.

Roming22 (approved):

PR #363, opened 2022-11-09
Implement single source of truth for images

Roming22:
All versions dependencies are now controlled by
'shared/config/dependencies.sh'.


.github/workflows/build-push-images.yaml
```diff
@@ -58,7 +64,7 @@ jobs:
         uses: redhat-actions/buildah-build@v2
         with:
           image: access-setup
-          context: ./operator/images/access-setup
+          context: .
```
bnallapeta, line 67: context represents the path where a Dockerfile is present. Not able to follow the intent to change it to '.' here. Could you please explain?

Roming22: My understanding is that `containerfiles` is the path of the Dockerfile, and `context` is the path of the data that is available to the `COPY` command (see the doc for the [build-context](https://docs.podman.io/en/latest/markdown/podman-build.1.html#build-context-name-value) option, although it does not set that specific flag).

PR #365, opened 2022-11-10
Fix bitwarden installation

Roming22:
During the image creation 'bw --version' was run as root while $HOME was already set to '/tmp/home'. This generated files in $HOME/.config owned by root, causing a failure when 'bw' was called again as a different user.

- The 'install_bitwarden' function has been cleaned up.
- The Dockerfiles for 'cluster-setup' and 'devenv' have been modified so that $HOME is set after all dependencies installation.
- Add missing yq dependency to cluster-setup
- Fix issue when the workspace is read-only for the user.
- Fix issue with tracing being activated in the middle of the install.
- Fix credential leak

### DIFF
--- a/operator/images/cluster-setup/content/bin/install.sh
+++ b/operator/images/cluster-setup/content/bin/install.sh
@@ -60,6 +60,7 @@ parse_args() {
       WORKSPACE_DIR="$1"
       ;;
     -d | --debug)
+      DEBUG="--debug"
       set -x
       ;;
     -h | --help)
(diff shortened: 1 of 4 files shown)


shared/hack/install.sh
```diff
@@ -134,9 +134,8 @@ install_argocd() {
 }
 
 install_bitwarden() {
-    curl "${CURL_OPTS[@]}" -o "$TMPBIN/bw.zip" "https://github.com/bitwarden/clients/releases/download/cli-${BITWARDEN_VERSION}/bw-linux-${BITWARDEN_VERSION:1}.zip"
-    unzip "$TMPBIN/bw.zip" -d "$TMPBIN/"
-    rm -rf "$TMPBIN/bw.zip"
```
bnallapeta, line 139: Removing this line will copy the zip file into BIN_DIR as move_bin is called after this. We don't want the zip file, just the executable.

operator/images/cluster-setup/content/bin/install.sh
```diff
@@ -74,17 +75,29 @@ parse_args() {
...
+    rm -rf "$tmp_workspace_dir"
+  fi
+  cp -rf "$WORKSPACE_DIR" "$tmp_workspace_dir"
+  WORKSPACE_DIR="$tmp_workspace_dir"
 }
 
 # turns off tracing even with set -x mode enabled across the script to prevent secrets leaking
 setx_off() {
   set +x
-} 2>/dev/null
```
bnallapeta, line 82: the reason I put these is that users do not need to know that we are turning on and turning off tracing. Removing this will print set +x and set -x which is unnecessary I feel.

bnallapeta (approved):
Minor comments but going to merge this for now.

PR #367, opened 2022-11-11
Update the kubeconfig file printed for accessing kcp-workspace in ckcp

ramessesii2:
#### Right now, the kubeconfig printed for accessing default kcp-workspace is the same one used for accessing the whole kcp-server spawned by ckcp. Instead, we should use the `kubeconfig` generated using `pipeline-service-manager` SA which is limited to the default workspace created.


Mo3m3n:
LGTM
@ramessesii2 can you put the content of the message explaining this PR in the commit subject too ?
So we can understand in the future what this commit is about when looking into the git history

ramessesii2:
Done

PR #368, opened 2022-11-12
Removing tmp folder in install.sh as its breaking deployment on staging cluster

bnallapeta:
- Moving files into /tmp folder and assigning that as the workspace path is causing issues in the Gitlab CI.
- Corresponding PR has already been raised on Gitlab with more details.


ramessesii2 (approved):
lgtm

PR #375, opened 2022-11-16
Move bitwarden function to utils.sh

Mo3m3n:
bitwarden fetch was moved to utils.sh:
- As a helper function it makes more sense to be in a generic script.
- It makes it easier to reuse it when debugging: for example we can do the following:
   `source ./operator/images/cluster-setup/content/bin/utils.sh `
    ````fetch_bitwarden_secrets ````


xinredhat:
LGTM

bnallapeta (approved):

PR #376, opened 2022-11-18
PLNSRVCE-812 : define Pod and container securityContext and SCCs

ramessesii2:
With this PR, we add more configurations to deployment manifests to follow operator security standards.
    - define SCC
    - drop all capabilities
    - run as non-root user
    - resource requests and limits
    - readOnlyRootFileSystem

### DIFF
--- a/operator/gitops/argocd/tekton-chains/base/chains-controller-deployment.yaml
+++ b/operator/gitops/argocd/tekton-chains/base/chains-controller-deployment.yaml
@@ -14,6 +14,25 @@ spec:
             # Mount the secrets volume
             - mountPath: /etc/ssl/certs
               name: chains-ca-cert
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - "ALL"
+            # User 65532 is the nonroot user ID
+            runAsUser: 65532
+            runAsGroup: 65532
+            runAsNonRoot: true
+            seccompProfile:
+              type: RuntimeDefault
+            readOnlyRootFilesystem: true
+          resources:
+            requests:
+              cpu: 100m
+              memory: 100Mi
+            limits:
+              cpu: 500m
+              memory: 500Mi
       volumes:
         - name: chains-ca-cert
           secret:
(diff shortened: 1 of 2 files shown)


ramessesii2:
Hi @CathalOConnorRH @avinal, I think you've been working on deploying tekton-results with Pipeline-Service. One question I had was regarding adding liveness and readiness probes for tekton-results api and watcher. Do you think it would makes sense to add them? I tried searching for any such practices in `tekton-results` upstream config, I can't seem to find it though. What do  you think?

avinal:
IIRC Cathal did stress testing of Tekton Results some months ago.

operator/gitops/argocd/tekton-chains/base/chains-controller-deployment.yaml
```diff
@@ -14,6 +14,25 @@ spec:
             # Mount the secrets volume
             - mountPath: /etc/ssl/certs
               name: chains-ca-cert
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+              - "ALL"
+            # User 65532 is the nonroot user ID
+            runAsUser: 65532
```
Mo3m3n, line 23: Do we need to run as specific user ID ?
isn't `runAsNonRoot`  enough ?

CathalOConnorRH:
I agree with adding the liveness and readiness probes. 

When we tested tekton-results it was with postgres on cluster. its worth re running load tests once the new architecture is defined and we have agreed SLOs. Initial tests were testing the limits of what could be stored in results.

operator/gitops/argocd/tekton-chains/base/chains-controller-deployment.yaml
ramessesii2: Hey, the rationale was to define an explicit userID. Although I'm aware that the userID defaults to the UID defined in the image metadata but since we have defined `runAsNonRoot`, it won't default to root UID anyway.

ramessesii2:
Thanks Cathal, I've created Jira story `[PLNSRVCE-866]` for adding those probes. 
@Mo3m3n @bnallapeta @avinal  please give a review. We can merge it if it looks good to you.

PR #379, opened 2022-12-02
Automate update of base images in Dockerfiles

Roming22:
- Move operator/images/kcp-upgrade to developer/images/dependencies-update
- Add script to update sha of base images
- Update kcp upgrade script to work with the new layout
- Fix an issue on a path in the kcp upgrade script


developer/images/dependencies-update/tasks/update_dockerfiles_base_images.sh
```diff
@@ -0,0 +1,109 @@
...
+  if grep --max-count=1 --quiet "FROM $base_image_name@$base_image_sha" "${DOCKERFILES[@]}"; then
+    echo "No update"
+  else
+    update_base_image_sha
+    echo "Updated to latest sha"
+  fi
+}
+
+get_base_image_sha() {
+  base_image_sha=$(skopeo inspect "docker://$base_image_name" | yq ".Digest")
```
xinredhat, line 90: We need to add skopeo as one of dependent tools in [DEPENDENCIES.md](https://github.com/openshift-pipelines/pipeline-service/blob/main/DEPENDENCIES.md) and [dependencies.sh](https://github.com/openshift-pipelines/pipeline-service/blob/main/shared/config/dependencies.sh)

developer/images/dependencies-update/update.sh
```diff
@@ -0,0 +1,110 @@
...
+    esac
+    shift
+  done
+}
+
+init() {
+  if [ -z "${TASKS[*]}" ]; then
+    TASKS=( "${DEFAULT_TASKS[@]}" )
+  fi
+  COMMIT_MSG="$TMPDIR/update_commit_msg.txt"
```
xinredhat, line 88: When I manually executed `update.sh` without parametes, it reported error `TMPDIR: unbound variable`.  It needs to clarify exporting the value of `TMPDIR` before executing `update.sh` in the `usage` function.

developer/images/dependencies-update/tasks/update_kcp_version.sh
```diff
@@ -0,0 +1,111 @@
...
+  latest_kcp_version="$(curl -s https://api.github.com/repos/kcp-dev/kcp/releases/latest | yq '.tag_name' | sed 's/v//')"
+  if [ -z "$latest_kcp_version" ]; then
+    printf "[ERROR] Could not retrieve the latest kcp version\n" >&2
+    exit 1
+  fi
+}
+
+update_kcp_version() {
+  printf "Upgrade kcp version from '%s' to '%s'\n" "$current_kcp_version" "$latest_kcp_version" | tee "$COMMIT_MSG"
+  sed -i "s,$current_kcp_version,$latest_kcp_version,g" .github/workflows/build-push-images.yaml
```
xinredhat, line 83: I verified it in my Linux VM and it reported errror
>sed: can't read .github/workflows/build-push-images.yaml: No such file or directory

developer/images/dependencies-update/update.sh
Roming22: Fixed.

developer/images/dependencies-update/tasks/update_kcp_version.sh
Roming22: Should be fixed now, please retry. Make sure to call it via `update.sh --task update_kcp_version`, as it will set a required env var.

developer/images/dependencies-update/tasks/update_dockerfiles_base_images.sh
Roming22: Add to [DEPENDENCIES.md](https://github.com/openshift-pipelines/pipeline-service/blob/main/DEPENDENCIES.md) only.  Skopeo is currently not installable through [install.sh](https://github.com/openshift-pipelines/pipeline-service/blob/main/shared/hack/install.sh), so the version does not need to be fixed in [dependencies.sh](https://github.com/openshift-pipelines/pipeline-service/blob/main/shared/config/dependencies.sh).

PR #400, opened 2022-12-14
Fix broken tekton-chains test

ramessesii2:
Fix broken tekton-chains test   
    - use Skopeo instead of Buildah in the Pipeline example manifest to copy image.
    - remove checks for argo tekton-chains app since it's no more deployed as an argo-app.

#### How to test
Do a `KUBECONFIG=~/.kube/config CASES=chains ./operator/test/test.sh` once Pipeline-Service is deployed on the cluster.

### DIFF
--- a/operator/test/manifests/test/tekton-chains/simple-copy-pipeline.yaml
+++ b/operator/test/manifests/test/tekton-chains/simple-copy-pipeline.yaml
@@ -26,15 +26,20 @@ spec:
           Task to copy a container image from a container repository to another.
         params:
           - name: IMAGE_SRC
-            description: Reference of the image buildah will pull.
+            description: Reference of the image skopeo will pull.
           - name: IMAGE_DST
-            description: Reference of the image buildah will push.
-          - name: BUILDER_IMAGE
-            description: The location of the buildah builder image.
-            default: quay.io/buildah/stable:v1.18.0
-        workspaces:
-          - name: sslcertdir
-            optional: true
+            description: Reference of the image skopeo will push.
+          - name: SKOPEO_IMAGE
+            description: The location of the skopeo image.
+            default: quay.io/skopeo/stable:v1.9.0
+          - name: srcTLSverify
+            description: Verify the TLS on the src registry endpoint
+            type: string
+            default: "true"
+          - name: destTLSverify
+            description: Verify the TLS on the dest registry endpoint
+            type: string
+            default: "true"
         results:
           - name: IMAGE_DIGEST
             description: Digest of the image just built.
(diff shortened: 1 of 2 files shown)


Roming22 (commented):
@ramessesii2 LGTM, but you'll need to fix the yamllint errors before ita can be merged.

Roming22 (approved):

operator/test/manifests/test/tekton-chains/simple-copy-pipeline.yaml
```diff
@@ -26,44 +26,48 @@ spec:
...
+                   --src-tls-verify="$(params.srcTLSverify)" \
+                   --dest-tls-verify="$(params.destTLSverify)"
+                echo "$(params.IMAGE_DST)" > "$(results.IMAGE_URL.path)"
+                cat "/tmp/image-digest" > "$(results.IMAGE_DIGEST.path)"
+              else
+                return 1
+              fi
+            securityContext:
+              runAsNonRoot: true
+              runAsUser: 65532
```
Mo3m3n, line 70: Is this required ?

Mo3m3n:
I suggest that in a separate commit (but in the same PR) we activate the chains test so that we can see if it works
Because for now it is not working for me:
````
-> % CASES=chains ./operator/test/test.sh
[test_chains]
Reset namespace 'test-tekton-chains'
Error from server (NotFound): namespaces "test-tekton-chains" not found
namespace/test-tekton-chains created
pipeline.tekton.dev/simple-copy created
OK
PipelineRun started: simple-copy-run-w4fxz
Waiting for logs to be available...
[image-copy : copy] Getting image source signatures
[image-copy : copy] Copying blob sha256:99a592b45028613a395793f9544b9b5c0e343ce2eb062acf1a0f07e338dd69b0
[image-copy : copy] Copying blob sha256:207e252fc3105530c479355645f87c5ff12df70567a0d6eb80d6f575efb5fea0
[image-copy : copy] Copying blob sha256:db8873ce6d61f6480c2a7c6c611ff45b4f08a3dba00070f2455e141e73daa168
[image-copy : copy] Copying blob sha256:117d566d60723e6389a42381d6dfcec6b3354382b52d652453720ed6131eb784
[image-copy : copy] Copying config sha256:3fea15c479c10166290f4812f3c6ab687d6288232040e0f4788a1f1a231a1fa7
[image-copy : copy] Writing manifest to image destination
[image-copy : copy] Storing signatures

Pipeline signed: .Failed
[ERROR] Unsigned pipeline (simple-copy-run-w4fxz)
```

operator/test/manifests/test/tekton-chains/simple-copy-pipeline.yaml
ramessesii2: `runAsUser` is not required. Although, defining SecurityContext is a good practice. Do you think it should be removed for any reason from this pipeline manifest?

ramessesii2:
Hi @Mo3m3n @Roming22, the tests for tekton-chains fails locally since Bitwarded isn't setup for tekton-chains when one does `dev_setup.sh`, and so Tekton-chains logs report `x509 certificate signed by unknown authority` and that's how Pipelines are rejected by chains while signing. 
Although, one quick way to test it is using pln-svc-stage cluster. The test passes since Bitwarded is setup on that cluster (using @bnallapeta 's account).

operator/test/manifests/test/tekton-chains/simple-copy-pipeline.yaml
Mo3m3n: > defining SecurityContext is a good practice.

runAsNonRoot should be enough I think. There should be a reason to run as a specific user otherwise I don't see the need for that.

Mo3m3n:
> Hi @Mo3m3n @Roming22, the tests for tekton-chains fails locally since Bitwarded isn't setup for tekton-chains when one does `dev_setup.sh`, and so Tekton-chains logs report `x509 certificate signed by unknown authority` and that's how Pipelines are rejected by chains while signing. Although, one quick way to test it is using pln-svc-stage cluster. The test passes since Bitwarded is setup on that cluster (using @bnallapeta 's account).

Probably not related to bitwarden, this [job](https://github.com/openshift-pipelines/pipeline-service/blob/main/operator/gitops/argocd/pipeline-service/tekton-chains/overlays/openshift/chains-secrets-config-patch.yaml) is used to update the ca cert of tekton-chains, which is required to  get around the `x509 certificate signed by unknown authority` message. 
But the job is only run after everything is synced by argocd (`  argocd.argoproj.io/hook: PostSync`) However we currently have an issue with tekton-results not being deployed correctly (tekton-results-api) which make this job never run. 
So maybe we need in a seperate PR fix tekton-results and enable its CI test.

operator/test/manifests/test/tekton-chains/simple-copy-pipeline.yaml
ramessesii2: That makes sense. I'll quickly make the changes.

Roming22 (approved):

operator/test/manifests/test/tekton-chains/simple-copy-pipeline.yaml
Mo3m3n: Actually I found that this was required :) 
More about it here: https://github.com/openshift-pipelines/pipeline-service/pull/423

PR #408, opened 2022-12-19
Fix tekton-results

Mo3m3n:
- Fix the config mount path for tekton-results-api. This seems to have
      changed after latest image update.
    - instruction to install postgres-db in dev environment was accidentally
      removed in 64ceee84af. This has been fixed

### DIFF
--- a/developer/openshift/dev_setup.sh
+++ b/developer/openshift/dev_setup.sh
@@ -199,6 +199,8 @@ install_pipeline_service() {
   "$PROJECT_DIR/operator/images/cluster-setup/content/bin/install.sh" \
     ${DEBUG:+"$DEBUG"} \
     --workspace-dir "$WORK_DIR" | indent 2
+  echo -n "- Install tekton-results DB: "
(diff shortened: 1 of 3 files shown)


adambkaplan (approved):
/approve

Do we have an e2e test which verifies that the apiserver is deployed correctly? What is the current behavior with the wrong mount path?

Roming22 (approved):

Mo3m3n:
> Do we have an e2e test which verifies that the apiserver is deployed correctly? What is the current behavior with the wrong mount path?

In [operator/test/test.sh](https://github.com/openshift-pipelines/pipeline-service/blob/main/operator/test/test.sh) we have e2e tests covering  pipelineruns, results and chains. But only pipelineruns were tested in CI which let some issues slip to the other components. We are fixing that now then enabling testing of all components in CI

>  What is the current behavior with the wrong mount path?

With a wrong mount path the pod  is never ready with logs showing that config was not found

PR #417, opened 2022-12-22
fix tekton-results test and enable it in CI 

Mo3m3n:


Mo3m3n:
Enabling ci tests for tekton-results requires grpc_cli.
Building from source grpc_cli at ci runtime make the ci job quite slow.
I am closing this PR until we are able to have an image for ci with tools already installed like described in https://github.com/openshift-pipelines/pipeline-service/issues/421.
And opening  https://github.com/openshift-pipelines/pipeline-service/pull/422 instead

PR #419, opened 2022-12-23
[WIP] Add minio tenant for tekton results

AndrienkoAleksandr:
Add minio log storage for tekton-results for development instances.

### Referenced issue:
https://issues.redhat.com/browse/PLNSRVCE-769


AndrienkoAleksandr:
@adambkaplan I resolved merge conflict and retested it, take a look, please, when you will be free.

developer/openshift/dev_setup.sh
```diff
@@ -181,6 +182,26 @@ setup_compute_access(){
     indent 2
 }
 
+install_minio() {
+  APP="minio"
+
+  local ns="$APP"
+  #############################################################################
+  # Install the gitops operator
```
Roming22, line 190: `Install the minio operator`

operator/gitops/argocd/pipeline-service/tekton-results/overlays/dev/kustomization.yaml
```diff
@@ -0,0 +1,31 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../../base
```
Roming22, line 5: missing indentation

operator/images/cluster-setup/content/bin/utils.sh
```diff
@@ -48,6 +48,36 @@ check_deployments() {
...
+  printf -- "- pod with label %s: " "$label"
+  
+  #a loop to check if the pod exists
+  local amountAttepts=20
+  local i=0
+  while [ -z "$(kubectl get pods -l $label -n $ns --no-headers -o custom-columns=':metadata.name')" ]; do
+    printf '.'; sleep 10;
+    i=$((i+1))
+    if [[ $i -eq "${amountAttepts}" ]]; then
+      printf " not found by timeout \n"
```
Roming22, line 64: `printf "[ERROR] Pod %s not found by timeout \n" "$label" >&2`

operator/images/cluster-setup/content/bin/utils.sh
```diff
@@ -48,6 +48,36 @@ check_deployments() {
...
+
+  printf "Exists"
+
+  #a loop to check if the pod is Available and Ready
+  if kubectl wait --for=condition=ready pod -l "$label" -n "$ns" --timeout=100s >/dev/null; then
+    printf ", Ready\n"
+  else
+    kubectl -n "$ns" describe pod -l "$label" 
+    kubectl -n "$ns" logs -l "$label" 
+    exit 1
```
Roming22, line 77: Add an error message such as `printf "[ERROR] Pod %s failed to start\n" "$label" >&2`

operator/images/cluster-setup/content/bin/utils.sh
```diff
@@ -48,6 +48,36 @@ check_deployments() {
...
 }
 
+check_pod_by_label() {
+  local ns="$1"
+  local label="$2"
+
+  printf -- "- pod with label %s: " "$label"
+  
+  #a loop to check if the pod exists
+  local amountAttepts=20
```
bnallapeta, line 58: amountAttepts -> numOfAttempts

operator/gitops/argocd/pipeline-service/tekton-results/overlays/dev/minio-user-secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+data:
+  S3_ACCESS_KEY_ID: bWluaW8=
+  S3_SECRET_ACCESS_KEY: bWluaW8xMjM=
```
bnallapeta, line 4: you need to move this file to Bitwarden and then get the secrets from Bitwarden in the shell script before it is used.

developer/openshift/operators/minio/tenant/tenant-config.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: storage-configuration
+  namespace: tekton-results
+type: Opaque
+stringData:
+  config.env: |-
+    export MINIO_ROOT_USER="minio"
```
bnallapeta, line 10: are there just dummy values that are going to be replaced dynamically or actual credentials? If they are actual credentials, then you need to move this file to Bitwarden and then get the secrets from Bitwarden in the shell script before it is used.

developer/openshift/operators/minio/minio.yaml
```diff
@@ -0,0 +1,12 @@
+---
```
bnallapeta, line 1: Please add the minio operator in DEPENDENCIES.md

Roming22:
Setting as a Draft PR, since it's not active and has pending comments.

Roming22:
@AndrienkoAleksandr You'll need to rebase against main for the tests to pass. Please have a look at the comments and set them to "Resolved" if you've taken the feedback into account.

developer/openshift/operators/minio/tenant/tenant-config.yaml
AndrienkoAleksandr: done.

AndrienkoAleksandr:
I provided some workaround security context for tenant.yaml:
```
      securityContext:
        runAsNonRoot: true
        runAsUser: 1000610001
        runAsGroup: 100620002
```
because I faced with some bug in the new minio operators 4.5.6 which was released recently. Ideally I would like to merge patch: https://github.com/minio/operator/pull/1407 , but I'm not sure if that's possible...

gabemontero:
@AndrienkoAleksandr @adambkaplan I was able to verify an approximate form of https://github.com/openshift-pipelines/pipeline-service/pull/419/commits/e65cf69997d4f848c5299d2e581aa70be4efdc9b and adding the nonroot SCC to the tenant's SA over in my https://github.com/openshift-pipelines/pipeline-service/pull/443

my tweaks were 
- around some file names / file location to better fit with my helm based install of 4.5.8
- with 4.5.8 at least we can avoid setting the uid/gid at all and just leave the securityContext as empty.  Now, that said, I do see where they add the uid: 1000 under the covers, and where @AndrienkoAleksandr PR upstream would have value, if their operator's defaulting would leave well enough alone, and let OpenShift's random ID generation hold, but certainly I would not view it as blocker for merging my #443 at this point.

gabemontero:
> @AndrienkoAleksandr @adambkaplan I was able to verify an approximate form of [e65cf69](https://github.com/openshift-pipelines/pipeline-service/commit/e65cf69997d4f848c5299d2e581aa70be4efdc9b) and adding the nonroot SCC to the tenant's SA over in my #443
> 
> my tweaks were
> 
>     * around some file names / file location to better fit with my helm based install of 4.5.8
> 
>     * with 4.5.8 at least we can avoid setting the uid/gid at all and just leave the securityContext as empty.  Now, that said, I do see where they add the uid: 1000 under the covers, and where @AndrienkoAleksandr PR upstream would have value, if their operator's defaulting would leave well enough alone, and let OpenShift's random ID generation hold, but certainly I would not view it as blocker for merging my [[WIP] Add minio tenant for tekton results #443](https://github.com/openshift-pipelines/pipeline-service/pull/443) at this point.

a bit of a follow up on your upstream PR @AndrienkoAleksandr - I don't think it is exactly what we want for running on openshift, as you are still setting uid/gid by default:  https://github.com/minio/operator/blob/8aec1cbd758a00964070df0064a350e42e480452/pkg/resources/statefulsets/minio-statefulset.go#L420-L427

I think **IF** we pursue contributing upstream to minio, it would be better to use the discovery client to see if we are on openshift, where:
- if we are not on openshift, we default that uid to 1000, etc.
- if we are on openshift, we leave the security context blank, and then openshift pick the random non root uid,gid
- then in our testing, we sort out if we need any SCC for the tenant SA, i.e. non-root vs. restricted, for the pod to come up

AndrienkoAleksandr:
> I don't think it is exactly what we want for running on openshift, as you are still setting uid/gid by default: https://github.com/minio/operator/blob/8aec1cbd758a00964070df0064a350e42e480452/pkg/resources/statefulsets/minio-statefulset.go#L420-L427

No, actually if we have an empty security context with nil values: https://github.com/minio/operator/blob/8aec1cbd758a00964070df0064a350e42e480452/pkg/resources/statefulsets/minio-statefulset.go#L438 , then this values will stay nil.
But I like idea with Openshift detector. Maybe this will make pr more reasonable to merge for maintainers. I will try to handle this proposal.

PR #422, opened 2023-01-02
Fix and reorganize tekton-results

Mo3m3n:
- Fix leader election issue with tekton-results-watcher 
- remove cert-manager since it is no longer required
- Install db manifests as part of the tekton-results application
- Update TektonConfig


Roming22:
Can this PR be split in 2 with one PR fixing the issue, the other removing the cert-manager, or are the 2 tightly connected?

Mo3m3n:
Actually this PR is never going to be green because CI is always looking for argo apps from main.
So I am going to force merge this one and I have raised this issue in https://github.com/openshift-pipelines/pipeline-service/issues/429

Issue #421, opened 2023-01-02
Improve ci by using a pre-built image with required tools


Mo3m3n:
Currently we install,at [runtime](https://github.com/openshift-pipelines/pipeline-service/blob/86cc667aeac0f3ab6d3f926f43eb1f4486510f43/.tekton/pipeline-service-test.yaml#L209) (when ci is triggered), the required tools for ci tests.
Which makes CI slow and hard to keep tools versions aligned with the dependencies versions.
Using an image with required tools already installed would make this simpler

Roming22:
@Mo3m3n is that fixed by #427?

Mo3m3n:
yes it is

PR #423, opened 2023-01-03
Fix and enable tekton-chains test in CI

Mo3m3n:
When using runAsNonRoot we need one of the following:
- Use the `USER` directive in Dockerfile to specify a non root user
- use the `runAsUser` in kubernetes SecurityContext. 
In the case of tekton-chains test the used image does not specify a non root USER so `runAsUser` is required.

### DIFF
--- a/.tekton/pipeline-service-test.yaml
+++ b/.tekton/pipeline-service-test.yaml
@@ -317,7 +317,7 @@ spec:
 
                 echo "Start executing pipeline cases ..."
                 TEST_DIR=$(find "$PWD" -type f -name test.sh -exec dirname {} +)
-                KUBECONFIG=/kubeconfig CASES=pipelines $TEST_DIR/test.sh
+                KUBECONFIG=/kubeconfig CASES=pipelines,chains $TEST_DIR/test.sh
                 EOF
 
                 chmod +x plnsvc_test.sh
--- a/operator/test/manifests/test/tekton-chains/simple-copy-pipeline.yaml
+++ b/operator/test/manifests/test/tekton-chains/simple-copy-pipeline.yaml
@@ -67,6 +67,7 @@ spec:
               fi
             securityContext:
               runAsNonRoot: true
+              runAsUser: 65532
       params:
         - name: IMAGE_SRC
           value: $(params.image-src)


.tekton/pipeline-service-test.yaml
```diff
@@ -317,7 +317,7 @@ spec:
 
                 echo "Start executing pipeline cases ..."
                 TEST_DIR=$(find "$PWD" -type f -name test.sh -exec dirname {} +)
-                KUBECONFIG=/kubeconfig CASES=pipelines $TEST_DIR/test.sh
+                KUBECONFIG=/kubeconfig CASES=pipelines, chains $TEST_DIR/test.sh
```
ramessesii2, line 320:
```suggestion
                KUBECONFIG=/kubeconfig CASES=pipelines,chains $TEST_DIR/test.sh
```
Ah, there is an extra space.

ramessesii2 (changes requested):

.tekton/pipeline-service-test.yaml
Mo3m3n: Fixed

Mo3m3n:
CI is failing because of `tkn` not available. This is another reason to handle https://github.com/openshift-pipelines/pipeline-service/issues/421

adambkaplan:
I thought `runAsUser` presents issues for OpenShift. Does OpenShift 4.11 change the restricted SCC to allow non-zero `runAsUser` UIDs?

Mo3m3n:
@adambkaplan 

>I thought runAsUser presents issues for OpenShift. Does OpenShift 4.11 change the restricted SCC to allow non-zero runAsUser UIDs?

That's usually the case yes,  maybe the issue does not happen here because the test is deploymed by the admin user directly (kubectl) rather than via a service account.

gabemontero:
> @adambkaplan
> 
> > I thought runAsUser presents issues for OpenShift. Does OpenShift 4.11 change the restricted SCC to allow non-zero runAsUser UIDs?
> 
> That's usually the case yes, maybe the issue does not happen here because the test is deploymed by the admin user directly (kubectl) rather than via a service account.

if you were creating a `Pod`, yes that would be the reason, but with a `PipelineRun` the tekton controller is the requesting user.  So the tekton controller's permissions are unioned with the permissions of the serviceaccount for the `PipelineRun` when `Pod` admission validation is ultimately performed.

Roming22:
@Mo3m3n Rebasing against main should solve your CI issue.

gabemontero:
> > @adambkaplan
> > > I thought runAsUser presents issues for OpenShift. Does OpenShift 4.11 change the restricted SCC to allow non-zero runAsUser UIDs?
> > 
> > 
> > That's usually the case yes, maybe the issue does not happen here because the test is deploymed by the admin user directly (kubectl) rather than via a service account.
> 
> if you were creating a `Pod`, yes that would be the reason, but with a `PipelineRun` the tekton controller is the requesting user. So the tekton controller's permissions are unioned with the permissions of the serviceaccount for the `PipelineRun` when `Pod` admission validation is ultimately performed.

I was curious @adambkaplan and @Mo3m3n so a bit of a follow up:

- if you look at the tekton controller itself, it sets its runs as with `65532` per https://github.com/tektoncd/pipeline/blob/f9021b1385c0d5bfa4e61b9473597ca03d173cb6/config/controller.yaml#L134 and it is setting restricted profile per https://github.com/tektoncd/pipeline/blob/2ae8b6a54a1bab401dbd3b1cc533a3b087ca54ea/config/100-namespace/100-namespace.yaml#L22
- now, when you install on openshift, it still does its "random user ID" thing which has been around since the 3.x days ... a yaml snippet from a live tekton controller pod installed via the pipelines operator on my 4.11 cluster:

```
      runAsGroup: 65532
      runAsUser: 1000670000
```

but bottom line, I think you only get in trouble when the user is sufficiently low enough to move beyond restricted, and 65532 is "safe"

Mo3m3n:
closing this one in favor of https://github.com/openshift-pipelines/pipeline-service/pull/430

PR #439, opened 2023-01-16
Move pipeline-service-tests logic to shell scripts

Roming22:
This change aims to make the pipeline yaml more readable by abstracting the logic in scripts, as well as being able to apply the CI to the content of those scripts.

### DIFF
--- a/.tekton/pipeline-service-test.yaml
+++ b/.tekton/pipeline-service-test.yaml
@@ -31,6 +31,8 @@ spec:
       - name: shared-workspace
     tasks:
       - name: produce-cluster-name
+        runAfter:
+          - "fetch-repository"
         taskSpec:
           results:
             - name: cluster-name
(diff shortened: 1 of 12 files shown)


.tekton/pipeline-service-test.yaml
```diff
@@ -31,6 +31,8 @@ spec:
       - name: shared-workspace
     tasks:
       - name: produce-cluster-name
+        runAfter:
+          - "fetch-repository"
```
bnallapeta, line 35: Is there a reason this has to be run after fetch-repository?

.tekton/pipeline-service-test.yaml
```diff
@@ -48,61 +50,25 @@ spec:
                 echo -n "$CLUSTER_NAME" | tee $(results.cluster-name.path)
       - name: clean-clusters
         runAfter:
-          - "produce-cluster-name"
-        params:
-          - name: namespace
-            value: "$(params.namespace)"
-          - name: cluster-name
-            value: "$(tasks.produce-cluster-name.results.cluster-name)"
+          - "fetch-repository"
```
bnallapeta, line 53: Shouldn't this ideally run after everything is done as the last step? And not after fetch-repository?

bnallapeta (approved):
LGTM apart from those minor comments.
@xinredhat PTAL and merge if it looks good to you.

.tekton/pipeline-service-test.yaml
Roming22: It makes a bit more sense when looking at flow in the ui, as `delete-clusters` does not depend on that task.

PR #444, opened 2023-01-18
Update binaries to the latest version

Roming22:


gabemontero (approved):
tests are passing 

/lgtm

shared/config/dependencies.sh
```diff
@@ -1,15 +1,15 @@
 #!/usr/bin/env bash
-export ARGOCD_VERSION="v2.4.5"
-export BITWARDEN_VERSION="v2022.10.0"
-export CHECKOV_VERSION="2.2.254"
-export GO_VERSION="1.19.4"
-export GRPC_CLI_VERSION="v1.50.0"
-export HADOLINT_VERSION="v2.10.0"
+export ARGOCD_VERSION="v2.3.13"
```
ramessesii2, line 2: This seems to be an older version to `ARGOCD_VERSION="v2.4.5"`

Roming22: Good catch. Fixed. I'll work on a patch to the task so it does not happen again.

ramessesii2 (approved):

PR #452, opened 2023-01-24
Increase pruner frequency to every 2 minutes

Roming22:
Until tekton-results prunes the pipelines as soon as the logs have been stored, we need to be more aggressive in the pruning to prevent users running into PVCs quota issues.


adambkaplan (approved):
/approve

Do we have a mechanism for measuring an SLI for this CronJob? Should we even do this if Tekton Results is going to prune PipelineRuns?

gabemontero (approved):
I'll reiterate a comment I made in the arch call comments @Roming22  - if you want to check out the cpu/mem impact of such a change, `kubectl top pod` is a useful tool.

Roming22:
@adambkaplan This is just to tie us over until Tekton Result is up and running.

PR #460, opened 2023-02-03
Fix pipeline-service-update-dependencies

Roming22:


Roming22:
Forcing merge, as this change cannot impact customers.

PR #462, opened 2023-02-06
fixed issue of trigger test

xinredhat:
This PR is to fix the issue of Trigger test, which missed a [rbac.yaml](https://raw.githubusercontent.com/tektoncd/triggers/v0.18.0/examples/rbac.yaml).


Roming22 (approved):
We should probably remove that test in the near future since we don't use triggers anymore.

PR #469, opened 2023-02-08
Revert "update tekton-results manifests and images"

Roming22:
This reverts commit 5284a2ccd6a507eda6fa48fcaee6cda84c13a569. The commit broke the CI, and the issue is hard enough to fix that reverting the commit is necessary.


adambkaplan:
Revert and roll forward is sometimes the best option. +1 for having the courage to do this.

PR #470, opened 2023-02-08
Use source branch for ci

Roming22:
Add an option to dev_setup.sh to deploy the current branch.
This option should make it easy to test the source branch in PRs.

Validation of the PR needs to happen in throw away PR #471.

### DIFF
--- a/developer/openshift/dev_setup.sh
+++ b/developer/openshift/dev_setup.sh
@@ -29,6 +29,11 @@ Usage:
 Setup Pipeline Service on a single cluster.
 
 Optional arguments:
+    --force
+        No question asked.
+    --use-current-branch
+        Use the current branch to deploy the application. In the case of a detached
+        head, the revision is used instead.
     -w, --work-dir
         Directory in which to create the gitops file structure.
         If the directory already exists, all content will be removed.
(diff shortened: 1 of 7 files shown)


bnallapeta (approved):

developer/openshift/dev_setup.sh
```diff
@@ -238,9 +271,20 @@ install_pipeline_service() {
   "$PROJECT_DIR/operator/images/access-setup/content/bin/setup_work_dir.sh" \
     ${DEBUG:+"$DEBUG"} \
     --work-dir "$WORK_DIR" \
-    --kustomization "$GIT_URL/developer/openshift/gitops/argocd?ref=$GIT_REF" |
+    --kustomization "git::$GIT_URL/developer/openshift/gitops/argocd?ref=$GIT_REF" |
```
Mo3m3n, line 274: shouldn't we also use PROJECT_DIR here too when `USE_CURRENT_BRANCH` is set ?

Mo3m3n (changes requested):

developer/openshift/dev_setup.sh
Roming22: No. What's happening is that we set everything as usual, but before running the install, we go in and modify the content, which is what's happening line 277.

PR #472, opened 2023-02-09
Update binaries

Roming22:
- argocd to v2.6.1
- checkov to 2.2.356
- go to 1.20
- grpc_cli to v1.52.0
- oc to 4.12.2
- tektoncd_cli to v0.29.1


bnallapeta:
@Roming22 I don't think we are using Go anywhere in our repo. I checked with @xinredhat too and he said he would check if would need for a future implementation for ci-runner. Nothing at the moment.
Once he confirms, we can remove go as a dependency.

xinredhat:
> @Roming22 I don't think we are using Go anywhere in our repo. I checked with @xinredhat too and he said he would check if we would need for a future implementation for ci-runner. Nothing at the moment. Once he confirms, we can remove go as a dependency.

Yes,  the future implementation for ci-runner needs Go to build hypershift cli binary

bnallapeta:
Closing this in favour of #478 as that has more latest binaries.

PR #475, opened 2023-02-09
PLNSRVCE-1098: provision and destroy hypershift by using hyperhisft cli

xinredhat:
1. Refresh installed and set up a new Hyperhsift cluster by using hypershift cli
2. CI `pipeline-service-test` connect to the new Hypershift to spawn OCP 4.12  
3. the CI load credentials from Bitwarden server
4. Removed `storageClassName` in PVC, so that it will pick the default storageclass
5. ci-runner image installs hypershift cli

Signed-off-by: Xin Jiang @xinredhat


ci/images/ci-runner/Dockerfile
```diff
@@ -30,7 +41,8 @@ RUN set -x \
         xz-5.2.5 \
     && microdnf clean all
 COPY --from=grpc_cli /usr/local/bin/grpc_cli /usr/local/bin/
+COPY --from=hypershift /usr/local/bin/hypershift /usr/local/bin/
 COPY shared /tmp/image-build/shared
-RUN /tmp/image-build/shared/hack/install.sh --debug --bin argocd,go,jq,kubectl,tkn,yq \
+RUN /tmp/image-build/shared/hack/install.sh --debug --bin argocd,go,jq,kubectl,tkn,yq,bitwarden \
```
Roming22, line 46: please keep the iist ordered.

shared/hack/install.sh
```diff
@@ -171,6 +174,21 @@ install_grpc_cli() {
     fi
 }
 
+install_hypershift() {
+    # Build from code
```
Roming22, line 178: This should checkout a specific tag set in `shared/config/dependencies.sh`

xinredhat: hyperhsift qe told me that we should take the latest one from main branch and I looked at the tag, the latest one  `2.0.0-20220406093220`  is too old.  I will put main as the version for now.

ci/images/ci-runner/Dockerfile
xinredhat: fixed

shared/hack/install.sh
Roming22: Let's use `main` as the version number for the moment then.

Roming22:
Don’t forget to squash the commits.

prietyc123:
@xinredhat I am not sure if you are aware but we have something called as `fetch_bitwarden_secret` functionality in our code base already https://github.com/openshift-pipelines/pipeline-service/blob/main/operator/images/cluster-setup/content/bin/utils.sh#L86 

I was wondering if we can also use these for our use case 🤔

xinredhat:
> @xinredhat I am not sure if you are aware but we have something called as `fetch_bitwarden_secret` functionality in our code base already https://github.com/openshift-pipelines/pipeline-service/blob/main/operator/images/cluster-setup/content/bin/utils.sh#L86
> 
> I was wondering if we can also use these for our use case 🤔

yes, but this part of logic is different. If possible, we can refactor our code to avoid the duplicate code

Roming22 (approved):

PR #476, opened 2023-02-09
Fix automatic update

Roming22:
Files were not added before commiting the changes


Roming22:
Forcing merge without review as cannot impact users, and I'd like to troubleshoot any credential issue if there is any.

PR #478, opened 2023-02-09
Update binaries

Roming22:
Generated by developer/images/dependencies-update/hack/bin/update.sh


Roming22:
@xinredhat Are we good on the Go and HyperShift versions? Or should any of them be frozen?

xinredhat:
> @xinredhat Are we good on the Go version update? Or should it be frozen?

@Roming22 I think we are good on the Go version update. lets see the CI results.

PR #485, opened 2023-02-10
Refactor operator/test/test.sh

Roming22:
* Use parameters to control behavior.
* Improve output formatting.
* Run each test in its own namespace.
* Remove unused triggers test.
* Update doc.


Mo3m3n:
Do you think it is necessary to run each test in its own NS.
Maybe tests can be linked: create pipelines, sign them and store them.
But more importantly this means we will need to each time ask for  additional permissions to access a new NS (in staging and prod envs)

Roming22:
I see. I did that to simplify troubleshooting, but that might not be a great idea for the general usecase. I'll put everything in `pipeline-service-test`.

Mo3m3n:
> I see. I did that to simplify troubleshooting, but that might not be a great idea for the general usecase

Sure it is easier to quickly identify pipelines of different tests, maybe we can prefix the piplines name with the corresponding test.

PR #511, opened 2023-02-21
Use only REST for Route and enable logs API

avinal:
## Changes
- remove gRPC from Route and remove tests based on gRPC
- add a tekton-results env config for configmap generation with updated options and replace the default config
- enable logs API
- add tekton-results test using REST API
- remove NET_BIND_SERVICE capability from watcher deployment
- part of PLNSRVCE-1115


developer/images/dependencies-update/hack/bin/tasks/update_binaries.sh
```diff
@@ -72,10 +72,6 @@ get_go_version() {
     )
 }
 
-get_grpc_cli_version() {
```
Roming22, line 75: Please remove the version in `shared/config/dependencies.sh` as well.

operator/test/test.sh
```diff
@@ -196,42 +196,44 @@ test_results() {
...
 
-  if [[ $RECORD_RESULT == *$RESULT_UID* ]]; then
-    echo "OK"
-  else
-    echo "Failed"
-    echo "[ERROR] Unable to retrieve record $RESULT_UID from pipeline run $pipeline_name" >&2
-    exit 1
-  fi
-  echo
+    QUERY_CMD=(
```
Roming22, line 212: That does not look properly formatted. I'd expect the content to be indented, and the closing `)` to be on its own line, not indented.

shared/hack/install.sh
```diff
@@ -155,27 +155,6 @@ install_go() {
     go version
 }
 
-install_grpc_cli() {
```
Roming22, line 158: Remove the build of grpc_cli from the Dockerfiles.

Roming22 (approved):

avinal:
Just FYI, this is a breaking change: gRPC based setup won't work anymore.

adambkaplan:
> Just FYI, this is a breaking change: gRPC based setup won't work anymore.

This is probably the last time we can do this. Now that we have a production environment, we cannot introduce breaking changes without a migration path.

avinal:
Sure @adambkaplan

PR #513, opened 2023-02-22
Disable the infra-deployment push pipeline

Roming22:
The pipeline won't work until we have onboarded to stonesoup. Disabling it so the build status stays green in github.


bnallapeta:
@Roming22 [results test](https://github.com/openshift-pipelines/pipeline-service/blob/main/operator/test/test.sh#L184) has failed. PTAL.

xinredhat:
@Roming22 You need to rebase. `test.sh` in main branch works well.

gabemontero (approved):
I'm good with this @Roming22 but potential heads up - Michal is reworking the bundles as part of STONEBLD-777 and it appears that includes reduction of the use of shared secrets per https://github.com/redhat-appstudio/sprayproxy/pull/27#issue-1595223784

If that ultimately impacts https://github.com/redhat-appstudio/build-definitions/blob/main/task/init/0.1/init.yaml#L24-L36 and the task our push job here references (which pull in the `init` task), then we may be able to get this push job working with just the current vanilla tekton we have in our current CI, and no longer be dependent on onboarding to stone stoup.

I hope to track him down and see if I can get an answer in the near term.

Roming22:
@gabemontero I'm 100% okay with enabling the task once Michal's work is done :smile:

gabemontero:
> @gabemontero I'm 100% okay with enabling the task once Michal's work is done smile

turns out, while the bundle ref should change to use the `0.1` tag from https://quay.io/repository/redhat-appstudio-tekton-catalog/task-update-infra-deployments?tab=tags from what I can see it still uses shared resources per https://github.com/redhat-appstudio/build-definitions/blob/main/task/init/0.1/init.yaml#L30-L36 where I *think* that is still pulled into whatever PaC creates in `tekton-ci`

for now I'll just make an update to use the new bundle, but leave the file naming the same

then we'll sync up @Roming22 on when the onboardnig to stonesoup commences

PR #528, opened 2023-03-01
create pipeline-service-storage application

Mo3m3n:
The pipeline-service-storage app will install both postgres and minio used only for development purpose


Roming22 (approved):

Roming22:
My understanding is that this PR is obsolete because of #530 .

PR #529, opened 2023-03-01
Rearrange developer folder

Mo3m3n:
Remove local folder not used any more. Currently we are focusing on openshift cluster only.


Mo3m3n:
Moving this to the backlog

PR #539, opened 2023-03-05
Remove argocd sync-wave param from minio

Mo3m3n:
When pipeline-service is imported in infra-deployments project (as kustomize overlays and not argcd app), the sync-wave params on minio makes introduces a kind of dead lock. That is tekton-results waiting for its storage backend and minio waiting for the first sync-wave to finish. This has been fixed by removing the sync-wave which should still let minio deploy correctly


Mo3m3n:
This is required to correctly import pipeline-service in https://github.com/redhat-appstudio/infra-deployments/pull/1441

PR #541, opened 2023-03-06
Reverting embedded status from minimal to both as it is breaking integration service

bnallapeta:


bnallapeta:
As per the [documentation](https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md#beta-api---default-to-full-then-both-then-minimal), changing the status to _both_ should let the other components using pipeline service to migrate without breaking changes.

PR #504, opened 2023-02-16
PLNSRVCE-1115: Update tekton results manifests to latest

avinal:
## Changes

This update removes old tekton results manifests which was based on
release 0.4.0. Since a new relase is due we will be using development
version of tekton results.

- reference the manifests from openshift-pipelines/tektoncd-results
repository and remove yamls from pipeline service
- remove db deployments and config from tekton-results for staging and
  prod
- move database deployment to local and provide secrets manually.
  Automatic secrets won't work because bitnami/postgres image requires
the data fields either one of the vars or alias define here: https://github.com/bitnami/containers/blob/main/bitnami/postgresql/15/debian-11/rootfs/opt/bitnami/scripts/postgresql-env.sh
- change how token is generated for testing tekton-result
- partially resolves #463, PLNSRVCE-1115

Signed-off-by: Avinal Kumar <avinal@redhat.com>

## Action Items

- [x] Needs https://github.com/openshift-pipelines/tektoncd-results/pull/28
- [ ] ~~Needs https://github.com/redhat-appstudio/infra-deployments/pull/1364~~

PR #545, opened 2023-03-09
Reorganize tekton-results 

Mo3m3n:
Reorganize tekton-results manifest in order to:
- Prefer patchesStrategicMerge with kustomize rather than patchesJson6902, the first is clearer.
- Move s3 logging config from developer folder to operator one
- reflect s3 secrets layout from staging/prod


operator/gitops/argocd/pipeline-service/tekton-results/base/kustomization.yaml
```diff
@@ -25,91 +26,9 @@ configMapGenerator:
...
-  # we are using the tekton-results namespace, so patch the arguments to watcher
-  - target:
-      kind: Deployment
-      name: tekton-results-watcher
-      namespace: tekton-results
-    patch: |-
-      - op: replace
-        path: /spec/template/spec/containers/0/args/1
-        value: tekton-results-api-service.tekton-results.svc.cluster.local:50051
-      - op: remove
```
Mo3m3n, line 90: @avinal Why we have to drop the securityContext capability here ?

avinal (commented):
Cool, looks much organized now 👏🏼

operator/gitops/argocd/pipeline-service/tekton-results/base/kustomization.yaml
avinal: We don't want NET_BIND permissions, it lets pod listen on any port IIRC. Let me reference the comment where @adambkaplan suggested to remove it.

avinal: https://github.com/openshift-pipelines/pipeline-service/pull/504#discussion_r1110324875

Mo3m3n: Ok so in your case, it seems you were adding it explicitly (+ sign in the diff) to postgres so I understand Adam's comment.
Do we want also to explicitly remove `NET_BIND_SERVICE` in our overlay while it is available by default in upstream ?
@adambkaplan ?

developer/openshift/gitops/argocd/pipeline-service/tekton-results/minio-user-cred-patch.yaml
```diff
@@ -1,22 +0,0 @@
----
```
Mo3m3n, line 1: @AndrienkoAleksandr This part seemed like unecessary to me since in tenant config we already have `requestAutoCert: true` which should mount automatically the ca cert. I was able to check that (automatic mounting) in this PR

AndrienkoAleksandr: We mount tls certificate to tekton-results controller to make possible do requests to minio tenant without tls error. Without this patch request will fail.

`requestAutoCert: true` mounts certificate to the minio tenant pod, not to the tekton-results controller. So, please, revert this change.

Mo3m3n: You're right I was looking into the wrong pod :) 
Do you know if there is someway ( a param) to set another name than `storage-tls` for the ca secret ?
This way we can have the same name we use in staging/prod for that secret so the config would work for all env

AndrienkoAleksandr: I think this name was generated by pattern: "tenant name" + "-" + "tls" . Our tenant name is "storage". So secret name is "storage-tls".

Mo3m3n: Yes I figured that part, however I have not found the param to configure a custom name for the secret

operator/images/access-setup/content/bin/setup_work_dir.sh
```diff
@@ -203,8 +203,11 @@ tekton_results_manifest(){
       --dry-run=client -o yaml > "$results_db_secret"
 
     kubectl create secret generic -n tekton-results tekton-results-s3 \
-    --from-literal=S3_ACCESS_KEY_ID="$TEKTON_RESULTS_S3_USER" \
-    --from-literal=S3_SECRET_ACCESS_KEY="$TEKTON_RESULTS_S3_PASSWORD" \
+    --from-literal=aws_access_key_id="$TEKTON_RESULTS_S3_USER" \
+    --from-literal=aws_secret_access_key="$TEKTON_RESULTS_S3_PASSWORD" \
+    --from-literal=aws_region='us-east-1' \
```
Roming22, line 208: I feel like maybe the AWS region should not be hardcoded?

Mo3m3n: this is only a "filling the gap" content for the dev env, we don't really use a s3 backend in this case but we use minio which simulate a s3 storage

Roming22: Then maybe put a value that describes that? E.g. `--from-literal=aws_region='not-applicable'`

developer/openshift/gitops/argocd/pipeline-service/tekton-results/minio-user-cred-patch.yaml
Mo3m3n: Done

operator/images/access-setup/content/bin/setup_work_dir.sh
Mo3m3n: Done

Roming22 (approved):

developer/openshift/gitops/argocd/pipeline-service/tekton-results/minio-user-cred-patch.yaml
AndrienkoAleksandr: It's not configurable: https://github.com/minio/operator/blob/master/pkg/apis/minio.min.io/v2/names.go#L64

AndrienkoAleksandr:
@Mo3m3n , Sorry, but I provided for you information that you should revert changes regarding to minio ca sertificate https://github.com/openshift-pipelines/pipeline-service/pull/545/files#r1132263560 , because that's a breaking change, and connection beetween tekton-results api server and minio tenant server will be broken. But you skip it. Could you explain, please?

AndrienkoAleksandr:
@Mo3m3n I proposed https://github.com/openshift-pipelines/pipeline-service/pull/552

PR #547, opened 2023-03-13
merger downgrade tasks into upgrade ci

xinredhat:
Merge downgrade test path into upgrade CI as part of test path


.tekton/pipeline/upgrade-tests.yaml
```diff
@@ -137,6 +137,101 @@ spec:
...
+          workspace: source
+      params:
+        - name: repo_url
+          value: $(params.repo_url)
+        - name: revision
+          value: $(params.revision)
+        - name: target_branch
+          value: $(params.target_branch)
+      taskSpec:
+        workspaces:
```
prietyc123, line 169: I think we already have such task defined in the [task](https://github.com/openshift-pipelines/pipeline-service/tree/main/.tekton/tasks) folder. Can we use the existing one and create a new task if not defined already. So that our pipeline looks clean

prietyc123 (changes requested):

.tekton/pipeline/upgrade-tests.yaml
```diff
@@ -137,6 +137,101 @@ spec:
...
+      params:
+        - name: target_branch
+          value: $(params.target_branch)
+      workspaces:
+        - name: kubeconfig-dir
+          workspace: upgrade-shared-workspace
+        - name: source
+          workspace: source
+      taskSpec:
+        params:
```
prietyc123, line 217: same as [comment](https://github.com/openshift-pipelines/pipeline-service/pull/547#discussion_r1135126462)

prietyc123 (changes requested):

.tekton/pipeline/upgrade-tests.yaml
xinredhat: fixed

.tekton/pipeline/upgrade-tests.yaml
```diff
@@ -137,6 +137,49 @@ spec:
           workspace: upgrade-shared-workspace
         - name: source
           workspace: source
+    - name: fetch-repository-post-upgrade
```
prietyc123, line 140: Suggestion: 
```suggestion
    - name: clone-pipeline-service-git-post-upgrade
```

xinredhat: fixed

PR #551, opened 2023-03-15
Temp fix rbac permissions for tekton-results watcher.

AndrienkoAleksandr:


operator/gitops/argocd/pipeline-service/tekton-results/base/kustomization.yaml
```diff
@@ -32,3 +32,10 @@ patchesStrategicMerge:
   - api-service-tls.yaml
   - watcher-config.yaml
   - delete-postgres.yaml
+
+patches:
```
Mo3m3n, line 36: Can you put the patch under patchesStrategicMerge for better consistency ?
Then we will update all our kustomize to the new  `patches`  directive in a separate commit.
Also please add a comment the patching file explaining the reason of the patch and referencing the upstream fix

Mo3m3n: Ah your're right Oleksandr there is no key to use strategic merge. Ok let's keep the  current version with `patches` then.

PR #552, opened 2023-03-16
Revert tls patch to fix connection tekton-result api and Minio S3

AndrienkoAleksandr:
# Related pr

https://github.com/openshift-pipelines/pipeline-service/pull/545

### DIFF
--- a/developer/openshift/gitops/argocd/pipeline-service/kustomization.yaml
+++ b/developer/openshift/gitops/argocd/pipeline-service/kustomization.yaml
@@ -7,6 +7,7 @@ resources:
 
 patchesStrategicMerge:
   - tekton-results/minio-create-bucket.yaml
+  - tekton-results/minio-tls.yaml
 
   # Skip applying the Tekton operands while the Tekton operator is being installed.
   # See more information about this option, here:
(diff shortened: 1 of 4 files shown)


Roming22 (approved):
@Mo3m3n If you don't have any comments, please merge.

Mo3m3n:
@AndrienkoAleksandr currently connections between tekton-results-api and Minio s3 are working since results-api returns the expected result (records and logs) in the corresponding e2e test.
So we need to understand why this patch is needed. I will try to have a look this week

adambkaplan:
@Mo3m3n the current e2e test can be best described as a smoke test - it verifies the API is available and is not returning an error result, but that is it. In my own demo from last sprint, the actual log data we are looking for does not exist in the record.

I am helping @avinal refine https://issues.redhat.com/browse/PLNSRVCE-1182 - we hope to run the "upstream" e2e test for Results, potentially in parallel with the other pipeline service tests.

AndrienkoAleksandr:
I found one more bug reproducible in the slowly clusters: init container doesn't fail when minio route is not ready, because in the entrypoint shell command was used `;` separator instead of `&&`. So I included this fix to this pr.

PR #561, opened 2023-03-24
Enable completed_run_grace_period for tekton-results

Roming22:
Delete PipelineRuns 10 minutes after completion.

### DIFF
--- a/operator/gitops/argocd/pipeline-service/openshift-pipelines/tekton-config.yaml
+++ b/operator/gitops/argocd/pipeline-service/openshift-pipelines/tekton-config.yaml
@@ -46,9 +46,4 @@ spec:
       kube-api-qps: 50
       kube-api-burst: 50
   pruner:
-    keep: 10
-    resources:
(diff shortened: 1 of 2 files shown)


enarha:
The way I read the documentation https://tekton.dev/docs/operator/tektonconfig/#note both `disabled: false` AND `schedule:` with empty value are required to disable the Tekton pruner. That does not make much sense and it looks more like a documentation issue, but I'm rising it anyway. If you have tested the above configuration and it disabled the pruner, then we are good.

Roming22:
> The way I read the documentation https://tekton.dev/docs/operator/tektonconfig/#note both `disabled: false` AND `schedule:` with empty value are required to disable the Tekton pruner. That does not make much sense and it looks more like a documentation issue, but I'm rising it anyway. If you have tested the above configuration and it disabled the pruner, then we are good.

@enarha In this case the key value is `disabled: true`. So I do not think the note applies.

PR #562, opened 2023-03-24
test/tekton-results: check the right number of non empty records

Mo3m3n:
This PR updates the tekton-results e2e test from only checking the right record name to:

- checking that records have non empty data
- checking we have the right number of records corresponding to the tested pipelines

### DIFF
--- a/operator/test/test.sh
+++ b/operator/test/test.sh
@@ -79,7 +79,10 @@ init() {
 setup_test() {
   echo "[Setup]"
   echo -n "  - Namespace configuration: "
-  if ! kubectl get namespace $NAMESPACE >/dev/null 2>&1; then
+  if kubectl get namespace $NAMESPACE >/dev/null 2>&1; then
+    # Delete old pipelines resources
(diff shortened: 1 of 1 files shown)


adambkaplan:
@Mo3m3n can you please update the PR description to describe why this change is needed?

operator/test/test.sh
```diff
@@ -79,15 +79,18 @@ init() {
...
+  else
     echo -n "."
     kubectl create namespace "$NAMESPACE" >/dev/null
+    # Wait for pipelines to set up all the components
+    while [ "$(kubectl get serviceaccounts -n "$NAMESPACE" | grep -cE "^pipeline ")" != "1" ]; do
+      echo -n "."
+      sleep 2
+    done
   fi
-  # Wait for pipelines to set up all the components
```
Roming22, line 86: I would not move that while block. The fact that the namespace already exists is not a guarantee that the SA has been setup.

Mo3m3n: Fixed

Roming22 (approved):

Mo3m3n:
@Roming22 I need to fix the tests after the last rebase

PR #572, opened 2023-04-04
Disable Tekton Results DB auto migration

enarha:
With the existing DB, it's causing a failure in api. Pod does not start.


adambkaplan (approved):
/approve

/lgtm

Disabling auto migration should not impact the service. Pods are currently crashlooping because the migration logic fails on apiserver start.

adambkaplan:
/retest

Known "failed to list pods" flake.

Roming22 (approved):

enarha:
Yes, we figured that earlier and that's why we moved to the "revert" strategy. I'm closing this PR.

PR #573, opened 2023-04-05
PLNSRVCE-1268: change to allow cluster alive for 2 hours when CI task fail

xinredhat:
**Changes:**
1. When one of CI tasks fail, such as `plnsvc-setup` , CI won't destroy the cluster
2. The cluster will be kept alive for around 2 hours, when it exceeds  2 hours, it probably will be destroyed automatically

### DIFF
--- a/.tekton/pipeline/acceptance-tests.yaml
+++ b/.tekton/pipeline/acceptance-tests.yaml
@@ -115,6 +115,12 @@ spec:
         - input: "$(tasks.deploy-cluster.status)"
           operator: notin
           values: ["None"]
+        - input: "$(tasks.plnsvc-setup.status)"
+          operator: notin
+          values: ["Failed"]
+        - input: "$(tasks.pipeline-service-tests.status)"
+          operator: notin
+          values: ["Failed"]
       params:
         - name: cluster-name
           value: "$(tasks.generate-cluster-name.results.cluster-name)"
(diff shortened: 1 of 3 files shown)


adambkaplan:
LGTM - @xinredhat if you think this PR is ready, go ahead and remove "WIP" from the title.

xinredhat:
> LGTM - @xinredhat if you think this PR is ready, go ahead and remove "WIP" from the title.

Done.

Roming22:
@adambkaplan I'm concerned about letting all clusters run for 2 hours.
1. Costs. A successful test takes about 20-25 mins. A failed test would cost 5 times as much. I just want to make sure we have budget for this.
2. Resources. I believe there's a limited number of Hypershift cluster that we can instantiate at any time. Any long running cluster is taking away from that pool, and could prevent other PRs from running.

I'm OK merging the PR as is for the moment, but I'd rather see us implementing a change that requires an action from the dev to prevent the cluster deletion.
Proposal: In case of failure, `/tmp/ready-to-delete.txt` is created, the task waits for 15mins, then check if still exists before deleting the cluster. The developer has 15mins to open a terminal on that pod and remove the file if they want to keep the cluster.

Roming22:
Merging the PR. Any issue I've brought up about cost/resources can be dealt with at a later date.

PR #578, opened 2023-04-07
WIP: Run Upgrade Test from AppStudio sha

adambkaplan:
Ensure that the Pipeline Service upgrade test runs a full upgrade from what is currently deployed in the AppStudio staging cluster. This should help us catch any upgrade failures/risks before they are deployed in AppStudio.

- Add `start_revision` parameter to upgrade test pipeline.
- Add fuller descriptions to upgrade test pipeline parameters.
- dev_setup.sh: Print the git URL and ref that Pipeline Service uses for its ArgoCD deployment.

This is an attempt to simulate the failure that caused [PLNSRVCE-1262](https://issues.redhat.com/browse/PLNSRVCE-1262)

### DIFF
--- a/.tekton/pipeline-service-upgrade-test.yaml
+++ b/.tekton/pipeline-service-upgrade-test.yaml
@@ -27,6 +27,8 @@ spec:
       value: "{{ revision }}"
     - name: target_branch
       value: "{{ target_branch }}"
+    - name: start_revision
+      value: 8f3ebf86101a4cf21fe7e86335e85a3fb6e3671f
   timeouts:
     pipeline: "1h0m0s"
   workspaces:
--- a/.tekton/pipeline/upgrade-tests.yaml
+++ b/.tekton/pipeline/upgrade-tests.yaml
@@ -6,8 +6,23 @@ metadata:
 spec:
   params:
     - name: repo_url
+      type: string
+      description: URL to git repository
     - name: revision
+      type: string
+      description: |
+        Commit SHA that Pipeline Service will upgrade to. This will either be
+        the commit for a pull request, or the commit of merged code.
     - name: target_branch
+      type: string
+      description: |
+        Target branch that initiated the pipeline - either a contributor's
+        branch for a pull request, or the branch name from a push event.
+    - name: start_revision
+      type: string
+      description: |
+        Commit SHA that Pipeline Service will upgrade from. This should be the
+        commit for a deployed version of Pipeline Service.
   timeouts:
     finally: "0h20m0s"
   workspaces:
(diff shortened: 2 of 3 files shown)


gabemontero:
generally LGTM @adambkaplan 

comparing `plnsvc-setup` and `plnsvc-upgrade-setup` and searching on `github.com/openshift-pipelines/pipeline-service/developer/openshift/gitops/argocd?ref` I see how it went from  `8f3ebf86101a4cf21fe7e86335e85a3fb6e3671f` to `fe2c5c3367c973042b76e2632bba6a95d7ba5812` 

one "non-blocking" thought, as much a question as a suggestion - I wonder if it is worth tracking the image shas in the `tekton-results` namespace (perhaps chains,pac,pipelines as well) between pre and post upgrade.  And then flag whether a pipeline-service bump actually included a bump in one of those components.  Maybe it is not even something we keep long term, but keep until we get a warm fuzzy that our results tests are catching upgrade issues like 1262.

one idea for the future - once we can re-enabled https://github.com/openshift-pipelines/pipeline-service/blob/main/.tekton/infra-deployment-push.yaml.disable we can add a new pac job that submits a PR against this repo to bump https://github.com/openshift-pipelines/pipeline-service/pull/578/files#diff-85bb240671fd16c74c6c85e06780d895b4bd524e8ecaf801bb64cca6283ed7d4R31 .... it would like like https://github.com/openshift-pipelines/tektoncd-results/blob/downstream/.tekton/tekton-results-update-pipeline-service.yaml but that sed script at https://github.com/openshift-pipelines/tektoncd-results/blob/downstream/.tekton/tekton-results-update-pipeline-service.yaml would manipulate the correct file

adambkaplan:
/hold

I think we have a bug or feature gap in our dev setup scripts (which is what the CI test relies on). Our upgrade test pipeline ultimately calls `openshift/dev_setup.sh` (through a long chain of tasks that needs to be better documented). This deploys the ArgoCD application that is hard-coded to target the `main` branch: https://github.com/openshift-pipelines/pipeline-service/blob/8f3ebf86101a4cf21fe7e86335e85a3fb6e3671f/developer/openshift/gitops/argocd/pipeline-service.yaml#L14. Setting `git_ref` in `config.yaml` deploys the `main` branch, no matter what.

@Roming22 @xinredhat @prietyc123 I frankly wonder if our normal CI is testing the right thing, if it ends up deploying the `main` branch of the developer's repo...

xinredhat:
> coded

@adambkaplan our CI uses the argument `--use-current-branch` to deploy the proper branch when calling `openshift/dev_setup.sh`, see the [command](https://github.com/openshift-pipelines/pipeline-service/blob/main/ci/images/ci-runner/hack/sidecar/bin/plnsvc_setup.sh#L18). Then it will modify `targetRevision` and `repoURL` with the proper values, see [here](https://github.com/openshift-pipelines/pipeline-service/blob/8f3ebf86101a4cf21fe7e86335e85a3fb6e3671f/developer/openshift/dev_setup.sh#L240-L251)

PR #582, opened 2023-04-11
install: Check Application Sync Status

adambkaplan:
When installing Pipeline Service, check that the ArgocD application has synced in addition to the deployment status of Pipeline Service components.  Also ensure that the ArgoCD application reports "Healthy".

- Add function check_applications to cluster-setup `utils.sh`
- Update cluster-setup `install.sh` to check ArgoCD sync status.

Fixes https://issues.redhat.com/browse/PLNSRVCE-1279

### DIFF
--- a/operator/images/cluster-setup/content/bin/install.sh
+++ b/operator/images/cluster-setup/content/bin/install.sh
@@ -130,6 +130,9 @@ install_clusters() {
     printf -- "- Checking application status\n"
     check_applications "openshift-gitops" "pipeline-service" | indent 4
 
+    printf -- "- Checking subscription status\n"
+    check_subscriptions "openshift-operators" "openshift-pipelines-operator" | indent 4
+
     #checking if the pipelines and triggers pods are up and running
     printf -- "- Checking deployment status\n"
     tektonDeployments=("tekton-pipelines-controller" "tekton-triggers-controller" "tekton-triggers-core-interceptors")
--- a/operator/images/cluster-setup/content/bin/utils.sh
+++ b/operator/images/cluster-setup/content/bin/utils.sh
@@ -32,7 +32,6 @@ check_applications() {
     if ! timeout 300s bash -c "while ! kubectl get application/$app -n $ns >/dev/null 2>/dev/null; do printf '.'; sleep 10; done"; then
       printf "%s not found (timeout)\n" "$app"
       kubectl get application/"$app" -n "$ns"
-      kubectl -n "$ns" get events | grep Warning
       exit 1
     else
       printf "Exists"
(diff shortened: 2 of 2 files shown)


operator/images/cluster-setup/content/bin/utils.sh
```diff
@@ -20,6 +20,35 @@ exit_error() {
...
+      printf "%s not found (timeout)\n" "$app"
+      kubectl get application/"$app" -n "$ns"
+      kubectl -n "$ns" get events | grep Warning
+      exit 1
+    else
+      printf "Exists"
+    fi
+
+     #a loop to check if the ArgoCD application is synced
+    if kubectl wait --for=jsonpath="{.status.sync.status}"="Synced" "application/$app" -n "$ns" --timeout=200s >/dev/null; then
```
gabemontero, line 42: over in infra-deployments they do something similar but slightly different:  https://github.com/redhat-appstudio/infra-deployments/blob/main/hack/preview.sh#L200-L203

first blush, I wonder if the distinction between `Synced` and `Healthy Synced` matters ... I myself would not claim to be knowledgeable enough to say

adambkaplan: Good catch - we need the health status as well: https://argo-cd.readthedocs.io/en/stable/operator-manual/health/

adambkaplan:
Looking good!

```
  - Deploy applications:
    - Source: https://github.com/openshift-pipelines/pipeline-service/tree/3f7bdc1c2190b87db0427ff1dd90af3fe66c1fd5
    - No Bitwarden secrets
    - Installing shared manifests... 
        namespace/tekton-results configured
        secret/minio-storage-configuration configured
        secret/tekton-results-database configured
        secret/tekton-results-s3 configured
    - Installing applications via Openshift GitOps... 
        application.argoproj.io/pipeline-service configured
        application.argoproj.io/pipeline-service-storage configured
    - Checking application status
        - pipeline-service: Exists, Synced, Healthy
```

operator/images/cluster-setup/content/bin/utils.sh
```diff
@@ -20,6 +20,42 @@ exit_error() {
...
+        printf ", Healthy\n"
+      else
+        printf ", Unhealthy\n"
+        kubectl -n "$ns" describe "application/$app"
+        exit 1
+      fi
+    else
+      printf ", OutOfSync\n"
+      kubectl -n "$ns" describe "application/$app"
+      exit 1
```
bnallapeta, line 54: This will cause the script to exit even when one of the application is out of sync. IMO, it would be better to log an error here and make the script continue to check the status for the all the applications.
Additionally, we could maintain an array "failed_applications" and then log it out at the end of the script.

operator/images/cluster-setup/content/bin/install.sh
```diff
@@ -127,6 +127,9 @@ install_clusters() {
     printf -- "- Installing applications via Openshift GitOps... \n"
     install_applications | indent 4
 
+    printf -- "- Checking application status\n"
+    check_applications "openshift-gitops" "pipeline-service" | indent 4
```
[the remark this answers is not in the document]
Mo3m3n: Also since we have check_applications now, do we still need check_deployments ?
My understanding is that argo apps are "synced" and "healthy" only if all the deployments are ready right ?
If that's the case then we can get rid of check_deployments.

Roming22:
Merging the PR as
* It is required to increase the level of confidence when merging #593.
* Comments address improvements rather than flaws, and can therefore be addressed in a follow-up PR.

operator/images/cluster-setup/content/bin/install.sh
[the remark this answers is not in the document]
adambkaplan: I think we would need to verify that the "AtLatestKnown" status is all we need to verify health. That may not be the case - for instance, OLM can say it installed the operator, but the operator can report that a deployment it manages ("operand") is not healthy.

PR #588, opened 2023-04-13
Enable usage of a different repo and branch for argo apps

Mo3m3n:
In developer/config.yaml we can reference a different git repo and branch for the developer deployment however this taken into account when installing argocd apps. This commit fixes it by updating argocd apps if GIT_URL and GIT_REF variables are set.


Roming22:
I'm not a fan of changing the source for the apps after they have been created. If anything starts being synced before the source change is picked up by ArgoCD, the behavior would not be the one expected.

However, you are fixing the issue with config.yaml not being very useful, which is nice.

I've built upon your PR and propose #601. It keeps your fix, but keeps the behavior of rendering the app of apps prior to the install so that everything look as expected from the get go.

PR #589, opened 2023-04-13
Added more debug information when deploy cluster fail

xinredhat:
This PR will print out more debug logs when cluster deployment fail


ci/images/ci-runner/hack/bin/deploy-cluster.sh
```diff
@@ -19,6 +19,22 @@ fetch_bitwarden_secrets() {
     get_aws_credentials
 }
 
+print_debug_info() {
+    printf "Print debug info......\n" | indent 2
+
+    printf "HostedCluster CR Status: " | indent 4
+    kubectl -n clusters get hostedcluster "$CLUSTER_NAME" -o yaml
+
+    printf "Control Plan pods of HostedCluster: " | indent 4
```
gabemontero, line 28:
```suggestion
    printf "Control Plane pods of HostedCluster: " | indent 4
```

gabemontero (commented):
based on today's experience @xinredhat  I would add a 

`kubectl get co -o yaml` so we can get the debug messages from any openshift operators that are stuck

ci/images/ci-runner/hack/bin/deploy-cluster.sh
xinredhat: fixed.

PR #594, opened 2023-04-13
Allow pipeline-service-manager to view subscriptions

Roming22:
This permission is required when installing the cluster with dev_setup.sh


gabemontero (approved):
/lgtm

PR #597, opened 2023-04-14
Destroy CI clusters when there is no active investigation

Roming22:
To prevent too many resources from being consumed, keeping the cluster is not the default action.
If a developer wants to investigate, they need to connect to a pod and remove the 'destroy_cluster.txt' file located at the root of the source directory.

### DIFF
--- a/ci/docs/continuous_integation.md
+++ b/ci/docs/continuous_integation.md
@@ -22,3 +22,10 @@ After that, you need to add a Configmap for storing the kubeconfig of ROSA clust
 ## Setup GitHub app
 
 - Need to configure GitHub app for Pipelines as Code configuration into Pipeline Service repository.
+
+## Debugging an issue during the CI execution
+The CI will destroy the test cluster at the end of the pipeline by default.
+This is an issue when troubleshooting is required.
+
+To bypass deletion of the test cluster, delete the `destroy-cluster.txt` file in created in the root of the cloned repository.
+There is a 15 minutes window to log onto the container running the `destroy-cluster` step and delete the file.
--- a/ci/images/ci-runner/hack/bin/destroy-cluster.sh
+++ b/ci/images/ci-runner/hack/bin/destroy-cluster.sh
@@ -9,6 +9,18 @@ SCRIPT_DIR="$(
   pwd
 )"
 
+# Give developers 15mins to connect to a pod and remove the file
+# if they want to investigate the failure
+if [ -e "$PWD/destroy-cluster.txt" ]; then
+    sleep 900
+    if [ -e "$PWD/destroy-cluster.txt" ]; then
+      echo "Failure is not being investigated, cluster will be destroyed."
+    else
+      echo "Failure under investigation, cluster will not be destroyed."
+      exit 1
+    fi
+fi
+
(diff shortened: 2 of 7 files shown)


gabemontero (changes requested):
I think an update to https://github.com/openshift-pipelines/pipeline-service/blob/main/ci/docs/continuous_integation.md is warranted for explaining the deal here, and should be included with this change

Roming22:
@gabemontero Thanks for raising the point. I've added the info in the doc.

ci/images/ci-runner/hack/bin/destroy-cluster.sh
```diff
@@ -9,6 +9,18 @@ SCRIPT_DIR="$(
...
 )"
 
+# Give developers 15mins to connect to a pod and remove the file
+# if they want to investigate the failure
+if [ -e "$PWD/destroy-cluster.txt" ]; then
+    sleep 900
+    if [ -e "$PWD/destroy-cluster.txt" ]; then
+      echo "Failure is not being investigated, cluster will be destroyed."
+    else
+      echo "Failure under investigation, cluster will not be destroyed."
```
xinredhat, line 19: When the cluster is reserved , how long it will be destroyed depends on the MAX_DURATION_MINS defined in [destroy-clusters.sh](https://github.com/openshift-pipelines/pipeline-service/blob/main/ci/images/ci-runner/hack/bin/destroy-clusters.sh). Do we need to extend the MAX_DURATION_MINS bigger?

ci/docs/continuous_integation.md
```diff
@@ -22,3 +22,12 @@ After that, you need to add a Configmap for storing the kubeconfig of ROSA clust
 ## Setup GitHub app
 
 - Need to configure GitHub app for Pipelines as Code configuration into Pipeline Service repository.
+
+## Debugging an issue during the CI execution
+The CI automatically destroy cluster at the end of the pipeline, which is an issue when
```
gabemontero, line 27:
```suggestion
The CI automatically by default destroys the test cluster at the end of the pipeline, which is an issue when
```

ci/docs/continuous_integation.md
```diff
@@ -22,3 +22,12 @@ After that, you need to add a Configmap for storing the kubeconfig of ROSA clust
...
+troubleshooting is required.
+
+To solve that issue, in case of failure, a `destroy-cluster.txt` file in created in the
```
gabemontero, line 30:
```suggestion
To bypass deletion of the test cluster, delete the `destroy-cluster.txt` file that is created in the
```

gabemontero (changes requested):

ci/images/ci-runner/hack/bin/destroy-cluster.sh
Roming22: I think 2hrs is long enough for the moment. We can change it later if we get feedback that it's too short.

ci/docs/continuous_integation.md
Roming22: Rephrased.

ci/docs/continuous_integation.md
Roming22: Rephrased.

gabemontero (approved):

PR #601, opened 2023-04-17
Fix dev_setup.sh

Roming22:
Changing the values in config.yaml did not yield the same result as passing the '--use-current-branch' parameter.

### DIFF
--- a/developer/openshift/dev_setup.sh
+++ b/developer/openshift/dev_setup.sh
@@ -117,8 +117,17 @@ init() {
   # get the list of APPS to be installed
   read -ra APP_LIST <<< "$(yq eval '.apps // [] | join(" ")' "$CONFIG")"
 
-  GIT_URL=$(yq '.git_url // "https://github.com/openshift-pipelines/pipeline-service.git"' "$CONFIG")
-  GIT_REF=$(yq '.git_ref // "main"' "$CONFIG")
+  # Get the repository/branch used by ArgoCD as the source of truth
+  if [ -n "${USE_CURRENT_BRANCH:-}" ]; then
+    GIT_URL="$(git remote get-url origin | sed "s|git@github.com:|https://github.com/|")"
+    GIT_REF="$(git branch --show-current)"
+    # In the case of a PR, there's no branch, so use the revision instead
+    GIT_REF="${GIT_REF:-$(git rev-parse HEAD)}"
+  else
+    GIT_URL=$(yq '.git_url // "https://github.com/openshift-pipelines/pipeline-service.git"' "$CONFIG")
+    GIT_REF=$(yq '.git_ref // "main"' "$CONFIG")
+  fi
+  GIT_URL=$(echo "$GIT_URL" | sed '/\.git$/! s/$/.git/')
 
   # Create SRE repository folder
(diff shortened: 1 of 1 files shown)


developer/openshift/dev_setup.sh
```diff
@@ -232,23 +240,18 @@ install_pipeline_service() {
     --kustomization "git::$GIT_URL/developer/openshift/gitops/argocd?ref=$GIT_REF" |
     indent 2
 
+  # Render the content of the app of apps and change the url/branch to target the
+  # expected repository/branch
+  manifest_dir="$(find "$WORK_DIR/environment/compute" -mindepth 1 -maxdepth 1 -type d)"
+  kubectl create -k "$manifest_dir" --dry-run=client -o yaml >"$manifest_dir/pipeline-service.yaml"
+  yq -i ".spec.source.repoURL=\"$GIT_URL\" | .spec.source.targetRevision=\"$GIT_REF\"" "$manifest_dir/pipeline-service.yaml"
```
Mo3m3n, line 247: This is going to replace source in all argo apps right ?  This may be an issue if we start using external argocd apps like in https://github.com/openshift-pipelines/pipeline-service/pull/585 (where we use a Helm chart as source of the argo app).
So we need to replace only sources of apps pointing to the `pipeline-service` repository that's why I am grepping for such apps here https://github.com/openshift-pipelines/pipeline-service/pull/588/files#diff-342e128bf9808c7ef44522ed6360d1b5f838cc9308f916b6c4489300abfc09cbR162

Roming22: That's correct. I've updated the PR to use kustomize patch mechanism to set the applications with the proper source.

developer/openshift/dev_setup.sh
```diff
@@ -229,26 +236,32 @@ install_pipeline_service() {
...
+kind: Application
+metadata:
+  name: $app
+  namespace: openshift-gitops
+spec:
+  source:
+    repoURL: $GIT_URL
+    targetRevision: $GIT_REF
+EOF
+    yq -i ".patches += [\"patch-$app.yaml\"]" "$manifest_dir/kustomization.yaml"
```
xinredhat, line 257: Because of kustomize #[4911](https://github.com/kubernetes-sigs/kustomize/pull/4911), it has to use patchesStrategicMerge explicitly. 
`yq -i ".patchesStrategicMerge += [\"patch-$app.yaml\"]" "$manifest_dir/kustomization.yaml"`

xinredhat: we can use the way [Patch using Path Strategic Merge](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/patches/#patch-using-path-strategic-merge)
`yq -i ".patches += [{\"path\": \"patch-$app.yaml\"}]"  "$manifest_dir/kustomization.yaml"`

Roming22: Fixed. You can merge if there are no other comments.

developer/openshift/dev_setup.sh
Roming22: I'll mark the conversation as resolved. Let me know if you do not agree with the latest solution.

PR #603, opened 2023-04-19
Upgrade Pipelines Operator to 1.10

bnallapeta:
The PR brings in the below changes:
- Upgrades OpenShift Pipelines Operator to 1.10
- Removes Pipelines as Code files (independent installation) as PaC (version 0.17.x) comes with 1.10 operator
- Remove disabling PaC installation from TektonConfig
- Update Security tests to remove pipelines-as-code namespace as PaC components are now part of openshift-pipelines namespace.


operator/gitops/argocd/pipeline-service/kustomization.yaml
```diff
@@ -4,7 +4,6 @@ kind: Kustomization
 
 resources:
   - openshift-pipelines
-  - pipelines-as-code
```
xinredhat, line 7: @bnallapeta As we don't need to install Pac as a separate resource. Do we need to delete its related manifest [files](https://github.com/openshift-pipelines/pipeline-service/tree/main/operator/gitops/argocd/pipeline-service/pipelines-as-code)

bnallapeta: yes, will do that later. That's why marked it as WIP. 
Wanted to get the tests triggered first.

bnallapeta: Done.

operator/gitops/argocd/pipeline-service/kustomization.yaml
```diff
@@ -4,7 +4,6 @@ kind: Kustomization
...
   - tekton-chains
```
Roming22, line 7: I would expect chains to be installed by the operator as well since it's GA.

gabemontero: +1
that was our understanding back during our last round of feature refinement, as installing it from the operator, aside from "reduced duplication", gets us "use RH based images" etc.

gabemontero: in fact the test failure implies to me "let's get the chains installed via 1.10 to get versions correct:
```
- Public key: Error from server (BadRequest): error when creating "/source/operator/test/manifests/test/tekton-chains/public-key.yaml": admission webhook "webhook.pipeline.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "serviceAccountName"
```

Roming22:
Closing in favor of #612 and #613.

PR #605, opened 2023-04-19
Log all issues during cluster setup

Roming22:
Instead of exiting on the first error, report as many faults as possible.


Roming22:
@adambkaplan @bnallapeta @Mo3m3n This PR is a follow-up to #582.

- It addresses Bharath's comment about exiting on the first error instead of trying to log as much as possible.
- There was an open conversation between Moemen and Aaam about still needing `check_deployments` that was not resolved, so the function is still there.
- It removes an old function that we don't seem to be using anymore.

operator/images/cluster-setup/content/bin/utils.sh
```diff
@@ -28,16 +28,17 @@ check_applications() {
...
     printf -- "- %s: " "$app"
 
-    #a loop to check if the ArgoCD application exists
+    # Check if the ArgoCD application exists
     if ! timeout 300s bash -c "while ! kubectl get application/$app -n $ns >/dev/null 2>/dev/null; do printf '.'; sleep 10; done"; then
       printf "%s not found (timeout)\n" "$app"
       kubectl get application/"$app" -n "$ns"
-      exit 1
+      export INSTALL_FAILED=1
+      return
```
gabemontero, line 36: I get why the `exit 1` is removed but why do we turn here?

Per the PR description of collecting as many problems as possible, would we not want to just set `INSTALL_FAILED=1` and then move onto the next check in this function, vs. calling return and existing the function?

Maybe it is because these checks are dependent.... i.e. if the ArgoCD application does not exist, say the OLM Subscription will never exist.

Admittedly, this comment may be as much for my understanding ... if so, thanks for indulging me.

Roming22: Your reasoning was my reasoning as well.

PR #612, opened 2023-04-21
Upgrading Openshift Pipelines Operator from 1.9 to 1.10

bnallapeta:
The PR brings in the below changes:
- Upgrades OpenShift Pipelines Operator to 1.10


Roming22:
Overriden by #618 as it needed to be rebased.

PR #620, opened 2023-04-25
Remove independent install of PaC as it is installed via the operator

Roming22:


operator/gitops/argocd/pipeline-service/kustomization.yaml
```diff
@@ -4,7 +4,6 @@ kind: Kustomization
 
 resources:
   - openshift-pipelines
-  - pipelines-as-code
```
gabemontero, line 7: question in https://github.com/openshift-pipelines/pipeline-service/pull/618#issuecomment-1522047401 answered :-)

are you doing chains in a separate PR for isolation purposes @Roming22 ?

thanks

Roming22: Correct.

PR #623, opened 2023-04-26
Update kustomization files to latest standard

Roming22:
This removes deprecation warning when applying kustomize to the files


gabemontero:
were you able to confirm the deprecation warnings were gone before the pipelineruns got pruned @Roming22 ?

otherwise, 
/lgtm

Roming22:
@gabemontero There are some warning left because they come from tekton-results resources. I've submitted a change for that.

I ran `kustomize build` locally, I didn't look at the logs, as there's no reason it should behave differently.

PR #617, opened 2023-04-25
Update tekton results version and configuration

sayan-biswas:
Use new version of Tekton Results and update configuration to reflect on the upstream changes.

PR #627, opened 2023-04-28
Add sync waves to tekton-results

sayan-biswas:
Allows ArgoCD to wait for other resources (database, minio, etc) to sync before deploying API server, and wait for API server to be sync before deploying watcher. 
Fixes restart of API server and watcher pods.


gabemontero:
we have already vetted this change in #617 but I agree with @sayan-biswas that this change was not directly related to the results bump, so let's separate

gabemontero (approved):

PR #629, opened 2023-05-01
[DNM] Add a dedicated service account to run Tekton pipelines

Roming22:
The service account is to be created automatically on tenant namespace creation. The definition of the ClusterRole is managed by pipeline service.

### DIFF
--- a/operator/gitops/argocd/pipeline-service/openshift-pipelines/appstudio-pipelines-runner-clusterrole.yaml
+++ b/operator/gitops/argocd/pipeline-service/openshift-pipelines/appstudio-pipelines-runner-clusterrole.yaml
@@ -1,0 +1,54 @@
+---
+# Any change to this role must be communicated to the devsandbox team
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: appstudio-pipelines-runner-clusterrole
+rules:
+  - apiGroups:
+      - tekton.dev
+    resources:
+      - pipelineruns
+    verbs:
+      - get
+      - list
+      - update
+      - patch
+  - apiGroups:
+      - tekton.dev
+    resources:
+      - taskruns
+    verbs:
+      - get
+      - patch
+  - apiGroups:
+      - tekton.dev
+    resources:
+      - taskruns/status
+    verbs:
+      - patch
+  - apiGroups:
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - get
+      - create
+      - patch
+      - delete
+  - apiGroups:
+      - sharedresource.openshift.io
+    resourceNames:
+      - redhat-appstudio-user-workload
+    resources:
+      - sharedsecrets
+    verbs:
+      - use
+  - apiGroups:
+      - security.openshift.io
+    resourceNames:
+      - csi-scc
+    resources:
+      - securitycontextconstraints
+    verbs:
+      - use
--- a/operator/test/manifests/test/tekton-chains/public-key.yaml
+++ b/operator/test/manifests/test/tekton-chains/public-key.yaml
@@ -20,4 +20,4 @@ spec:
                 set -x
                 PUBLIC_KEY=$(oc get secret public-key -n tekton-chains -o jsonpath='{.data.cosign\.pub}')
                 echo "$PUBLIC_KEY" | base64 -d
-  serviceAccountName: pipeline
+  serviceAccountName: appstudio-pipeline
--- a/operator/gitops/argocd/pipeline-service/openshift-pipelines/kustomization.yaml
+++ b/operator/gitops/argocd/pipeline-service/openshift-pipelines/kustomization.yaml
@@ -3,5 +3,6 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - allow-argocd-to-manage.yaml
+  - appstudio-pipelines-runner-clusterrole.yaml
   - openshift-operator.yaml
   - tekton-config.yaml
--- a/operator/test/manifests/test/tekton-chains/simple-copy-pipeline.yaml
+++ b/operator/test/manifests/test/tekton-chains/simple-copy-pipeline.yaml
@@ -2,7 +2,6 @@
 apiVersion: tekton.dev/v1beta1
 kind: Pipeline
 metadata:
-  annotations:
   name: simple-copy
 spec:
   params:
--- a/operator/test/manifests/test/tekton-pipelines/appstudio-pipeline-service-account.yaml
+++ b/operator/test/manifests/test/tekton-pipelines/appstudio-pipeline-service-account.yaml
@@ -1,0 +1,34 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: appstudio-pipeline
+  namespace: plnsvc-tests
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: appstudio-pipelines-edit-rolebinding
+  namespace: plnsvc-tests
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: edit
+subjects:
+  - kind: ServiceAccount
+    name: appstudio-pipeline
+    namespace: plnsvc-tests
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: appstudio-pipelines-scc-rolebinding
+  namespace: plnsvc-tests
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: pipelines-scc-clusterrole
+subjects:
+  - kind: ServiceAccount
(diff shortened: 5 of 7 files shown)


Roming22:
@adambkaplan please double check the permissions granted to the new service account.

operator/gitops/argocd/pipeline-service/openshift-pipelines/appstudio-pipelines-runner-clusterrole.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - imagestreamtags
+      - imagetags
+  - verbs:
+      - get
+      - update
+    apiGroups:
+      - ''
+      - image.openshift.io
+    resources:
+      - imagestreams/layers
```
adambkaplan, line 49: We use imagestreams in app studio? Do we also use the internal registry? 

If so, @gabemontero is this the right set of permissions needed to push to the internal registry?

operator/gitops/argocd/pipeline-service/openshift-pipelines/appstudio-pipelines-runner-clusterrole.yaml
```diff
@@ -0,0 +1,49 @@
+---
+# Any change to this role must be communicated to the devsandbox team
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: appstudio-pipelines-runner-clusterrole
```
adambkaplan, line 6: cc @Michkov @stuartwdouglas - I think the app studio build definitions need more permissions than what are listed here.

adambkaplan:
Documenting the plan here:

1. This PR creates the cluster role that app studio pipelines should use.
2. https://github.com/codeready-toolchain/host-operator/pull/797 will create the `appstudio-pipeline` service account in each tenant namespace, and the role binding within the namespace.
3. We switch the operator to use `appstudio-pipeline` as the default service account for pipelines:
   ```yaml
   kind: TektonConfig
   metadata: ...
   spec:
    pipelines:
      default-service-account: appstudio-pipeline
   ```
4. App studio components remove any hard-codes of the `pipeline` service account.
5. We turn off the automatic RBAC creation for the operator:
   ```yaml
   kind: TektonConfig
   metadata: ...
   spec:
    params:
      - name: createRbacResource
        value: "false"
   ```

operator/gitops/argocd/pipeline-service/openshift-pipelines/appstudio-pipelines-runner-clusterrole.yaml
Michkov: I am not aware of usage of imagestreams

operator/test/manifests/test/tekton-chains/public-key.yaml
```diff
@@ -20,4 +20,4 @@ spec:
                 set -x
                 PUBLIC_KEY=$(oc get secret public-key -n tekton-chains -o jsonpath='{.data.cosign\.pub}')
                 echo "$PUBLIC_KEY" | base64 -d
-  serviceAccountName: pipeline
+  serviceAccountName: appstudio-pipeline
```
Michkov, line 23: Do we want to set this SA as global -> https://github.com/openshift-pipelines/pipeline-service/blob/main/operator/gitops/argocd/pipeline-service/openshift-pipelines/tekton-config.yaml#L12 ?

adambkaplan: I would say yes - see my comment https://github.com/openshift-pipelines/pipeline-service/pull/629#issuecomment-1531520847

operator/gitops/argocd/pipeline-service/openshift-pipelines/appstudio-pipelines-runner-clusterrole.yaml
gabemontero: yeah way back when, like back in 2022 when I first joined appstudio, I remember there being some options to use imagestreams as a destination for builds in the build-definitions tests or something like that, but I don't see it in that repo anymore

and certainly I would consider @Michkov the source of truth wrt along with @sbose78 for use *within* app studio.

That said, if a customer can provide their own pipelines via PaC, in theory they could want to use imagestreams.  The openshift-pipelines team I believe done work to facilitate that, with the `pipeline` service account

Why do we need both this new service account and the openshift-pipelines `pipeline` SA ?

Pending the answers to those questions, if these permissions are needed, instead of what is done here, openshift creates ClusterRoles for poth pulling and pushing images from/to the internal image registry, and in fact, whenever a namespace or project is created, RoleBindings for those ClusterRoles are created for imagestreams defined in that namepsace and associated with predefined SA's like the openshift `builder` SA.

For example:

```
gmontero ~ $ oc get rolebinding system:image-pullers -o yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  annotations:
    openshift.io/description: Allows all pods in this namespace to pull images from
      this namespace.  It is auto-managed by a controller; remove subjects to disable.
  creationTimestamp: "2023-05-02T16:05:42Z"
  name: system:image-pullers
  namespace: ggmtest
  resourceVersion: "90370"
  uid: bf7efe78-5f35-4e06-820e-084ba40b496d
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: system:image-puller
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: system:serviceaccounts:ggmtest
gmontero ~ $ oc get rolebinding system:image-builders -o yaml 
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  annotations:
    openshift.io/description: Allows builds in this namespace to push images to this
      namespace.  It is auto-managed by a controller; remove subjects to disable.
  creationTimestamp: "2023-05-02T16:06:48Z"
  name: system:image-builders
  namespace: ggmtest2
  resourceVersion: "90862"
  uid: 9e85e373-ba7b-4d21-a457-9963854f089d
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: system:image-builder
subjects:
- kind: ServiceAccount
  name: builder
  namespace: ggmtest2
gmontero ~ $ 
```

So @Roming22 you should just reuse the existing `system:image-builders` and `system:image-pullers` ClusterRoles and create RoleBindings and/or ClusterRoleBindings as needed to allow interaction with imagestreams and the internal image registry.

Roming22: The reason behind these permissions is that they are used by the pipeline-service tests. The test for tekton-chains will copy an image to the internal registry and make sure that the attestation has been created.

I'll remove those from the "staging/production" configuration, and add it as part of the "development" configuration.

Roming22: Fixed.

Roming22:
@adambkaplan @Michkov Please provide your feedback on the updated permissions. We need to communicate them to the sandbox team as soon as possible.

operator/gitops/argocd/pipeline-service/openshift-pipelines/appstudio-pipelines-runner-clusterrole.yaml
```diff
@@ -0,0 +1,54 @@
...
+    resourceNames:
+      - redhat-appstudio-user-workload
+    resources:
+      - sharedsecrets
+    verbs:
+      - use
+  - apiGroups:
+      - security.openshift.io
+    resourceNames:
+      - csi-scc
```
gabemontero, line 50: maybe add a comment that this can go away when the last vestiges of shared resources are removed 

Last I understood, https://issues.redhat.com/browse/STONEBLD-867 was the tracker for that, but @Michkov would be the definitive source of truth for that.

operator/gitops/argocd/pipeline-service/openshift-pipelines/appstudio-pipelines-runner-clusterrole.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - get
+      - create
+      - patch
+      - delete
+  - apiGroups:
+      - sharedresource.openshift.io
```
gabemontero, line 40: maybe add a comment that this can go away when the last vestiges of shared resources are removed

Last I understood, https://issues.redhat.com/browse/STONEBLD-867 was the tracker for that, but @Michkov would be the definitive source of truth for that.

gabemontero (commented):
a minor suggestion

operator/gitops/argocd/pipeline-service/openshift-pipelines/appstudio-pipelines-runner-clusterrole.yaml
Michkov: yes, it should be removed trying to get information about time.

Michkov (approved):

Roming22:
ClusterRole is deployed by the DevSandbox.

PR #631, opened 2023-05-03
Use downstream images for pipelines-as-code and tekton-chains

Roming22:


operator/gitops/argocd/pipeline-service/pipelines-as-code/kustomization.yaml
```diff
@@ -20,3 +20,30 @@ patchesJson6902:
...
           argocd.argoproj.io/managed-by: openshift-gitops
+  - target:
+      kind: Deployment
+      name: pipelines-as-code-controller
+      namespace: pipelines-as-code
+    patch: |-
+      - op: replace
+        path: "/spec/template/spec/containers/0/image"
+        value:
+          registry.redhat.io/openshift-pipelines/pipelines-pipelines-as-code-rhel8@sha256:6ba7db7a829bdac4db7554a4643283d0e97fc7a407ed7b6297888124fb5af3f3
```
gabemontero, line 31: I'll add what I mentioned in planning @Roming22 for completeness.

Per https://catalog.redhat.com/software/containers/openshift-pipelines/pipelines-pipelines-as-code-rhel8/61f93fe495bac8d0de150ad8 how about trying `v1.10.0-6` instead of the specific sha

The way those tags work in registry.redhat.io,  `v1.10.0-6` is not a rolling tag, but will only ever point to the sha it is initially seeded with.

The makes it clear to humans when looking at the deployment which version we are using here.

operator/gitops/argocd/pipeline-service/tekton-chains/kustomization.yaml
```diff
@@ -39,3 +39,12 @@ patches:
...
       $patch: delete
+  # Use downstream image
+  - target:
+      kind: Deployment
+      name: tekton-chains-controller
+      namespace: tekton-chains
+    patch: |-
+     - op: replace
+       path: /spec/template/spec/containers/0/image
+       value: registry.redhat.io/openshift-pipelines/pipelines-chains-controller-rhel8@sha256:f4ad9be541f927d0b0348d4515639f490657e699ca2b0fdfb8552f144793beca
```
gabemontero, line 50: I'll add what I mentioned in planning @Roming22 for completeness.

Per https://catalog.redhat.com/software/containers/openshift-pipelines/pipelines-chains-controller-rhel8/61f9058895bac8d0de150838 how about trying `v1.10.0-4` instead of a specific sha

The way those tags work in registry.redhat.io, v1.10.0-4 is not a rolling tag, but will only ever point to the sha it is initially seeded with.

The makes it clear to humans when looking at the deployment which version we are using here.

gabemontero (commented):
Not an "uncompromising" suggestion @Roming22 @adambkaplan, and yes, we are all programmed now "to use shas", but I think we have a unique usability opportunity given these are `registry.redhat.io` refs instead of `quay.io`.

But again, if you all feel srtongly the other way, so be it.

operator/gitops/argocd/pipeline-service/pipelines-as-code/kustomization.yaml
Roming22: Fixed.

operator/gitops/argocd/pipeline-service/tekton-chains/kustomization.yaml
Roming22: Fixed.

gabemontero (approved):
/lgtm

Roming22:
test failure is on `destroy-cluster`, so merging the PR.

PR #637, opened 2023-05-10
Change default SA for Pipelines to appstudio-pipeline

Roming22:
The pipeline service account permissions are too open by default. Therefore the account needs to be replaced with an account with more limited permissions.

The service account is managed by the DevSandbox team.

### DIFF
--- a/operator/test/test.sh
+++ b/operator/test/test.sh
@@ -79,15 +79,7 @@ init() {
 setup_test() {
   echo "[Setup]"
   echo -n "  - Namespace configuration: "
-  if ! kubectl get namespace $NAMESPACE >/dev/null 2>&1; then
-    echo -n "."
-    kubectl create namespace "$NAMESPACE" >/dev/null
-  fi
-  # Wait for pipelines to set up all the components
-  while [ "$(kubectl get serviceaccounts -n "$NAMESPACE" | grep -cE "^pipeline ")" != "1" ]; do
-    echo -n "."
-    sleep 2
-  done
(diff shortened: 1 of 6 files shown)


operator/test/test.sh
```diff
@@ -79,15 +79,7 @@ init() {
 setup_test() {
   echo "[Setup]"
   echo -n "  - Namespace configuration: "
-  if ! kubectl get namespace $NAMESPACE >/dev/null 2>&1; then
-    echo -n "."
-    kubectl create namespace "$NAMESPACE" >/dev/null
-  fi
-  # Wait for pipelines to set up all the components
-  while [ "$(kubectl get serviceaccounts -n "$NAMESPACE" | grep -cE "^pipeline ")" != "1" ]; do
```
gabemontero, line 87: rather than remove this test, why not confirm that the `appstudio-pipeline` is created in the test namespace?

gabemontero: is that because of the openshift-pipelines bug I saw you mention in slack?

If so, instead of deleting, comment the test out, with a TODO comment that we should add it back with the new SA name once the openshift-pipelines fix in question is available

gabemontero (changes requested):

operator/test/test.sh
Roming22: We know the account is created by the kustomization, therefore there is no need to wait for it.
Previously the account was created asynchronously by a controller, which is why we had that test.

gabemontero: ok I missed kustomization creating it

PR #639, opened 2023-05-11
Revert "Use downstream images for pipelines-as-code and tekton-chains"

Roming22:
**This commit should only be merged by @gabemontero or @Roming22**

This reverts commit 04a0df2558ebb6879c2a413fc4f7d1ade53f5d6f.

This is a workaround in case we're still having issues with the AppStudio e2e tests at the end of the day.

### DIFF
--- a/operator/gitops/argocd/pipeline-service/pipelines-as-code/kustomization.yaml
+++ b/operator/gitops/argocd/pipeline-service/pipelines-as-code/kustomization.yaml
@@ -20,30 +20,3 @@ patchesJson6902:
         path: "/metadata/labels"
         value:
           argocd.argoproj.io/managed-by: openshift-gitops
-  - target:
-      kind: Deployment
-      name: pipelines-as-code-controller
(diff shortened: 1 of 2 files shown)


gabemontero (approved):
More work to be done with chains / enterprise contract and moving to the downstream images:  https://github.com/redhat-appstudio/e2e-tests/pull/515#issuecomment-1545584754

Roming22:
@gabemontero I would prefer if you did not revert until after the change to the default SA account is loaded. Those are 2 big changes, and I'd prefer if we loaded them separately. It will be easier to troubleshoot the root cause if anything goes wrong.

gabemontero:
> @gabemontero I would prefer if you did not revert until after the change to the default SA account is loaded. Those are 2 big changes, and I'd prefer if we loaded them separately. It will be easier to troubleshoot the root cause if anything goes wrong.

sure thing @Roming22 

turns out they were not able to unravel everything before EOB anyway

PR #644, opened 2023-05-17
Disable pipeline SA

Roming22:
The PR has 3 commits. This is because some of the changes are temporary and will need to be rolled back in the future.

### DIFF
--- a/operator/gitops/argocd/pipeline-service/openshift-pipelines/allow-argocd-to-manage.yaml
+++ b/operator/gitops/argocd/pipeline-service/openshift-pipelines/allow-argocd-to-manage.yaml
@@ -14,6 +14,16 @@ rules:
       - patch
       - create
       - delete
+  - apiGroups:
+      - security.openshift.io
+    resources:
+      - securitycontextconstraints
+    verbs:
+      - get
+      - list
+      - patch
+      - create
+      - delete
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
--- a/operator/gitops/argocd/pipeline-service/openshift-pipelines/pipelines-scc.yaml
+++ b/operator/gitops/argocd/pipeline-service/openshift-pipelines/pipelines-scc.yaml
@@ -1,0 +1,37 @@
+---
+kind: SecurityContextConstraints
+metadata:
+  name: pipelines-scc
+allowHostDirVolumePlugin: false
+allowHostIPC: false
+allowHostNetwork: false
+allowHostPID: false
+allowHostPorts: false
+allowPrivilegeEscalation: false
+allowPrivilegedContainer: false
+allowedCapabilities:
+  - SETFCAP
+apiVersion: security.openshift.io/v1
+defaultAddCapabilities: null
+fsGroup:
+  type: MustRunAs
+groups:
+  - system:cluster-admins
+priority: 10
+readOnlyRootFilesystem: false
+requiredDropCapabilities:
+  - MKNOD
+runAsUser:
+  type: RunAsAny
+seLinuxContext:
+  type: MustRunAs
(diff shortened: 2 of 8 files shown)


operator/gitops/argocd/pipeline-service/openshift-pipelines/allow-argocd-to-manage.yaml
```diff
@@ -14,6 +14,16 @@ rules:
...
+  - apiGroups:
+      - security.openshift.io
+    resources:
+      - securitycontextconstraints
+    verbs:
+      - get
+      - list
+      - patch
+      - create
+      - delete
```
gabemontero, line 26: do we really need delete ?   In the wrong hands could be destructive

operator/gitops/argocd/pipeline-service/openshift-pipelines/pipelines-scc.yaml
```diff
@@ -0,0 +1,37 @@
...
+metadata:
+  name: pipelines-scc
+allowHostDirVolumePlugin: false
+allowHostIPC: false
+allowHostNetwork: false
+allowHostPID: false
+allowHostPorts: false
+allowPrivilegeEscalation: false
+allowPrivilegedContainer: false
+allowedCapabilities:
```
gabemontero, line 12: I mentioned this in @ramessesii2 PR on this as well .... some comments on why we need the deltas from the restricted SCC would help here

gabemontero: I suspect they are related to building with buildah ... did you perhaps just copy what was being done from the pipeline SA, or is this a delta of what was done for the pipeline SA ?

operator/gitops/argocd/pipeline-service/openshift-pipelines/allow-argocd-to-manage.yaml
gabemontero: I'm guessing that perhaps the openshift pipelines operator had this as part of doing uninstall?

If so, I guess ok .... but if somebody ever hacked a pod with this permission ....

Roming22: Those permissions are needed to create the pipelines-scc resource. It is a temporary solution and will be removed in the future.

operator/gitops/argocd/pipeline-service/openshift-pipelines/pipelines-scc.yaml
Roming22: Those permissions were provided by @Michkov and ensures that build-service can run its pipelines.
We are temporarily responsible for those permissions, and we expect those changes to be reverted (which is why they are bundled in their own commit).

ramessesii2: I also tried comparing the one created automatically by Pipelines to the one provided by Michal and they seem the same to me. It's just we'll be creating them instead on Pipelines for now.

operator/gitops/argocd/pipeline-service/openshift-pipelines/allow-argocd-to-manage.yaml
gabemontero: ok

operator/gitops/argocd/pipeline-service/openshift-pipelines/pipelines-scc.yaml
gabemontero: ok so yes to my question (buildah related == what Michal would provide) and they are temporary

gabemontero (approved):

PR #645, opened 2023-05-17
Fix possible race issue

Roming22:
It was possible for the tekton-chains-trusted-cabundle job to run before the config-trusted-cabundle was cleaned by the OSP operator


adambkaplan (approved):
/approve

/lgtm

This should mitigate race condition concerns. Not guaranteed, but less likely since the "0" wave will disable the RBAC resource feature, causing the operator to begin tearing down the ca-bundle ConfigMaps (or leave them in place? unclear the behavior here).

PR #650, opened 2023-05-22
Fix package versions in Dockerfiles

Roming22:


Roming22:
/test pipeline-service-test

pipeline-service-upgrade-test is fine, failed on `destroy-cluster`.

PR #661, opened 2023-06-02
Add Initial Pipeline Serivce Dashboard

adambkaplan:
A Grafana dashboard will make it easy for operators to monitor Pipeline Service components at a glance. The dashboard definitions here can be deployed using Kustomize on a cluster with the Grafana operator installed.

This contains an initial dashboard definition, with the intent that future contributors can add additional graphs and charts. The dashboard consists of the following:

- Line graphs of completed TaskRuns and PipelineRuns, expressed as a rate per hour.
- A success gauge measuring the percent of successful TaskRuns and PipelineRuns.

The dashboard assumes that Prometheus will be used as the data source, and that the following metrics are recorded and exported via a ServiceMonitor:

- `tekton_pipelines_controller_pipelinerun_count`
- `tekton_pipelines_controller_taskrun_count`

### DIFF
--- a/operator/gitops/README.md
+++ b/operator/gitops/README.md
@@ -10,4 +10,14 @@ We want to make the onboarding experience to use Pipeline Service as easy and cu
 
 Before installing the prerequisites, refer [DEPENDENCIES.md](../../DEPENDENCIES.md) to verify the versions of products, operators and tools used in Pipeline Service.
 
-## TODO
+## Components
+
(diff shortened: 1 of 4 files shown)


adambkaplan:
Note - I am working on a follow up PR that will deploy Grafana and all the other obeservability config via `dev_setup.sh`. The important bit is just the JSON that folks can import, and the directory structure that lets us add the kustomization to infra-deployments.

Roming22 (approved):
/lgtm

PR #663, opened 2023-06-06
Increase log buffer size to create few chunks

sayan-biswas:
This PR increases the buffer size for streaming a response from the GetLogs API. This creates less number chunks and is a workaround for the bug [PLNSRVCE-1322](https://issues.redhat.com/browse/PLNSRVCE-1322).


gabemontero (approved):
/lgtm

PR #673, opened 2023-06-12
Remove usage of chains-ca-cert

lcarva:
The chains-ca-cert ConfigMap was introduced in order to allow the Tekton Chains controller to push content to the integrated OpenShift registry. Since then, we have moved away from using this registry altogether in favor of quay.io.

This change removes usage of the chains-ca-cert ConfigMap as well as the ArgoCD Jobs responsible for setting it up.

https://github.com/openshift-pipelines/pipeline-service/pull/673


lcarva:
NOTE: There's an e2e test that checks for the presence of this secret. Here's the PR to remove that test altogether: https://github.com/redhat-appstudio/e2e-tests/pull/584

PR #677, opened 2023-06-13
PLNSRVCE-1310: bump exporter mem from 256 to 512 Mib

gabemontero:
Have pattern where the exporter in production hovers around 256 MiB for several hours, and when we bump up too much against it, the kubelet restarts us.  No error message in prev log.  So consistent with mem limit rebump.

Going to 512 to start.

As point of ref, tekton pipeline controller in production is hovering around 1GiB (and it has no mem limit).

We watch less that ^^ so let's see the 512 run first.


Roming22:
Forcing merge as it won't impact the CI.

PR #682, opened 2023-06-14
Grafana Observational graphs for Pipelines Controller Resource Utilization

ramessesii2:
### The following metrics are being queried-
    - container_cpu_usage_seconds_total is the counter of cpu usage. 
      The "sum of rate" gives us the average vCPU usage over the past 5 minutes.
    - container_memory_working_set is the gauge metric of memory used,.
      and is what the kubelet uses to determine OOM kills.

### Preview
image

### DIFF
--- a/operator/gitops/argocd/grafana/dashboards/pipeline-service-dashboard.json
+++ b/operator/gitops/argocd/grafana/dashboards/pipeline-service-dashboard.json
@@ -20,7 +20,8 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
-  "id": 6,
+  "id": 18,
+  "iteration": 1686742425906,
   "links": [],
   "panels": [
     {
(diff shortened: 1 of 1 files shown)


ramessesii2:
> An example of the panel I am talking about: https://console-openshift-console.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/k8s/ns/openshift-pipelines/pods/pipeline-metrics-exporter-54fd88cb4b-9rbnf/metrics

I couldn't see the actual queries being triggered in the console but the queries proposed by me are standard ones which I found in my study on those metrics.

gabemontero:
> > An example of the panel I am talking about: https://console-openshift-console.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/k8s/ns/openshift-pipelines/pods/pipeline-metrics-exporter-54fd88cb4b-9rbnf/metrics
> 
> I couldn't see the actual queries being triggered in the console but the queries proposed by me are standard ones which I found in my study on those metrics.

What I'm saying @ramessesii2 is we can ask the folks who work on the openshift console where the code is that does those panels, and we can cross reference on the precise stats they use.  The key players are actually part of the developer org.

My experience is that there can sometimes be more than one prevailing standard :-) 

We can get into it during our next stand up and decide if we ask them and who sends the query to them, etc.

adambkaplan (approved):
/lgtm

`container_memory_working_set` is the gauge metric of memory used, and is what the kubelet uses to determine OOM kills.

`container_cpu_usage_seconds_total` is the counter of cpu usage. The "sum of rate" gives us the average vCPU usage over the past 5 minutes.

For future PRs, it may be helpful to track down the help text associated with the metric that Prometheus scrapes. I am not sure how easy this is to track, especially for things coming from core k8s.

adambkaplan:
There are lots of "kube pod dashboard" examples that use these two metrics - here's one: https://grafana.com/grafana/dashboards/747-pod-metrics/

ramessesii2:
> My experience is that there can sometimes be more than one prevailing standard :-)

Yes @gabemontero, I do agree with that. I'll add this action item on my todo list. (I clearly missed what you meant in your first comment 😅, sorry for that. )

ramessesii2:
Thanks @adambkaplan for adding the precise help text about the kube metrics.

gabemontero:
So I think between @adambkaplan note around https://grafana.com/grafana/dashboards/747-pod-metrics/ and a simple sanity check with the console team on what they do for that panel, satisfying due diligence on which metric(s) we are using here is achieved IMO @ramessesii2

ramessesii2:
Hello @adambkaplan & @gabemontero, it seems the [Console team](https://redhat-internal.slack.com/archives/C0VMT03S5/p1687429573824339) uses the same metrics (& queries on it) for `cpu usage` & `memory`. 
Do you think there's any other action item for me on this PR?

gabemontero (approved):
/lgtm

PR #685, opened 2023-06-15
PLNSRVCE-1069:  get exporter metrics to work from dashboards

gabemontero:
These are the changes I needed to get the pipeline-service-exporter metrics to be visible from the grafana dashboards.

Turns out we need to move the servicemonitor def from infra-deployments to this repo

A couple of minor selector labeling tweaks on top.

### DIFF
--- a/operator/gitops/argocd/pipeline-service/metrics-exporter/deployment.yaml
+++ b/operator/gitops/argocd/pipeline-service/metrics-exporter/deployment.yaml
@@ -20,6 +20,7 @@ spec:
           image: quay.io/redhat-appstudio/pipeline-service-exporter:placeholder
           ports:
             - containerPort: 9117
+              name: metrics
           resources:
             requests:
               memory: "128Mi"
--- a/operator/gitops/argocd/pipeline-service/metrics-exporter/kustomization.yaml
+++ b/operator/gitops/argocd/pipeline-service/metrics-exporter/kustomization.yaml
@@ -5,6 +5,7 @@ resources:
   - clusterrolebinding.yaml
   - deployment.yaml
   - service.yaml
+  - servicemonitor.yaml
 
 images:
   - name: quay.io/redhat-appstudio/pipeline-service-exporter
--- a/operator/gitops/argocd/pipeline-service/metrics-exporter/servicemonitor.yaml
+++ b/operator/gitops/argocd/pipeline-service/metrics-exporter/servicemonitor.yaml
@@ -1,0 +1,21 @@
+---
+# for RHTAP, the pipeline-service monitor is defined in infra-deployments, but we define here in our developer folder (vs. the operator folder) to define this out of dev_setup.sh
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: pipeline-service
+  namespace: openshift-pipelines
+spec:
+  jobLabel: app
+  namespaceSelector:
+    matchNames:
+      - openshift-pipelines
+  endpoints:
+    - path: /metrics
+      port: metrics
+      interval: 15s
+      scheme: http
+      honorLabels: true
+  selector:
+    matchLabels:
+      app: pipeline-metrics-exporter


developer/openshift/gitops/argocd/pipeline-service/metrics-exporter/servicemonitor.yaml
```diff
@@ -0,0 +1,21 @@
+---
+# for RHTAP, the pipeline-service monitor is defined in infra-deployments, but we define here in our developer folder (vs. the operator folder) to define this out of dev_setup.sh
```
gabemontero, line 2: The RHTAP def of the service monitor in infra deployments is over at https://github.com/redhat-appstudio/infra-deployments/blob/main/components/monitoring/prometheus/base/servicemonitors/pipeline-service.yaml#L31-L44

You's see some diffs as of the writing of this comment, namely the `namespaceSelector` and the use of `honorLabels`.

If we merge these changes, and then bump the pipeline-service level in infra-deployments, and then when I test, I still cannot access the metrics from the openshift console, I will be changing the data at https://github.com/redhat-appstudio/infra-deployments/blob/main/components/monitoring/prometheus/base/servicemonitors/pipeline-service.yaml#L31-L44 to exactly match what I have in my dev folder here.

operator/gitops/argocd/pipeline-service/metrics-exporter/servicemonitor_openshiftconsole_rbac.yaml
```diff
@@ -0,0 +1,31 @@
+---
```
gabemontero, line 1: I believe this is redundant, as the openshift-pipelines operator is given access to the tekton metrics which are in the same namespace, but I did this because
- to show @ramessesii2 what exactly is meant by "adding RBAC"
- make an explicit statement that the exporter wants to expose stats to openshift-monitoring, vs. just piggy-backing on top of what  the tekton operator does 


that said, once @ramessesii2 signs off that he sees this and know understand what I am getting at, I'm fine with removing this redundancy if reviewers prefer I do

ramessesii2 (approved):

gabemontero:
/hold 

per discussion in scrum, @adambkaplan prefers to leverage dashboards and the inner dev loop work he is doing instead of the openshift console for direct usage of the exporter stats

that said, I'm leaving this PR open for a short while, to assist @ramessesii2 when he returns for sorting out accessing the exporter metrics from grafana

I also asked @adambkaplan and he agreed to use exporter metrics as part of his work, where perhaps he can debug access exporter metrics from grafana

gabemontero:
OK @adambkaplan @ramessesii2 I was able to prove that with the changes in this PR, including a hack in the latest dashboard to call exporter metrics vs. tekton ones, coupled with a test branch in infra-deployments pointing to this PR, and using bootstrap / preview from infra-deployments, we can access the exporter's metrics.

The test branch in infra deployments created by boostrap / preview is  https://github.com/gabemontero/infra-deployments/commits/preview-test-pipeline-svc-grafana 

I'll attach a couple of screen shots.  One from the metrics panel.  Another from the grafana one.  The prometheus query itself is not something we would use long term, but it worked with my overriding the TaskRuns per Hour chart, and it produced the same value sligthlly above 6.

I'll start iterating on this PR to get it to the meets min.  I think among other things, the servicemonitor needs to move from the developer folder to the operator one.  Examining the system live, there were no other servicemonitors that linked to the pipeline-metrics-exporter deployment/service vian the selector, labels, etc.
grafana-dashboard

metrics-dashboard

gabemontero:
confirmed with @amisstea that we should move our servicemonitor out of infra-deployments and into our repo

will use this PR for that @adambkaplan @ramessesii2

gabemontero:
The 2 `Red Hat Trusted App Pipeline` failures are part of @xinredhat working on moving our CI to RHTAP.  Those will start working when he is done.

The 2 jobs on the pipeline-service CI are running per usual.

PR #659, opened 2023-05-26
Create a CronJob to fix RHTAPBUGS-256

Roming22:
This fix is temporary and meant to be reverted once PaC is deployed by OSP, as this will bump the PaC
version to a version that contains the bugfix.

PR #693, opened 2023-06-23
Create a CronJob to fix RHTAPBUGS-256

Roming22:
This fix is temporary and meant to be reverted once PaC is deployed by OSP, as this will bump the PaC
version to a version that contains the bugfix.

This is a follow-up on #659


Roming22:
@adambkaplan The issue happened again today.

Changes since #659:
* The script was modified to loop through namespaces instead of making a cluster wide query.
* The image was replaced to use an RH built image.

Is that good enough to help us until OSP 1.11 is deployed?

Roming22:
Closed in favor of an SOP.

PR #688, opened 2023-06-21
Upgrade to OSP 1.11

Roming22:
We can clean-up the commits if the tests are successful.

For chains:
    - configured chains to use the old namespace, so that we don't have to worry about migrating the signing key for the moment.
    - put the tekton-chains config in TektonConfig
    - no longer use the upstream release payload, and allow the operator to install chains
 For PaC, remove all PaC artifacts and let the operator install

PR #700, opened 2023-06-28
add ConfigMap and Service for chains metrics

avinal:
- Add ConfigMap and Service for observalibilty in Tekton Chains
- Add port configuration and required env for metrics in deployment
  patch
- Manully adding until PLNSRVCE-1359 is fixed


gabemontero (commented):
There are some similarities with what I mentioned in https://github.com/openshift-pipelines/pipeline-service/pull/701#pullrequestreview-1503711267 for PaC.

However, because of the chains certs, and what @Roming22 said in https://redhat-internal.slack.com/archives/C032EJ007C0/p1687524351832529, *maybe* disabling the operator's install of chains (and continue to simply set the deployment to use the downstream images) when 1.11.1 is available , that we may can continue with this one.

That said, at the moment, in @Roming22 's https://github.com/openshift-pipelines/pipeline-service/pull/688 , he is allowing the operator to install chains, and is doing the cert / key stuff in the old tekton-chains namespace.

So yeah, a little muddier than PaC.  We'll see how far I get with 1.11 before I go on PTO the beginning of July, and we should see where @Roming22 is at when he returns from PTO beginning of July.

avinal:
Ok, I will wait for opinions from Romain.

Roming22:
The priority right now is to be able to deploy OSP 1.11.   There's 2 scenario:
* We manage to work out the Chains migration before OSP 1.11.1 is out, and your change will be rolled back.
* We deploy OSP 1.11.1 and do not migrate Chains, then we can prepare the Chains migration but wait for the right Chains version to be deployed by OSP before migrating.
Which ever we can accomplish first will decide what happens to Chains.

The TL;DR is I'm perfectly fine merging this PR, but there's a chance its content might be rolled back.

As we talked during the stand-up, please add some e2e that validates that the expected metrics are there. It will be helpful during the migration.

operator/gitops/argocd/pipeline-service/tekton-chains/chains-controller-deployment.yaml
```diff
@@ -0,0 +1,31 @@
+---
```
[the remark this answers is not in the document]
avinal: I had them added (in my original commits), since it was not working, I was trying with alternatives.

operator/gitops/argocd/pipeline-service/tekton-chains/config-observability.yaml
```diff
@@ -0,0 +1,45 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-observability
+  namespace: tekton-chains
```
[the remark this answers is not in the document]
avinal: I see. Namespace would be a problem. I will try with openshift-pipelines.

operator/gitops/argocd/pipeline-service/tekton-chains/observability-service.yaml
```diff
@@ -0,0 +1,21 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: tekton-chains-metrics
```
sayan-biswas, line 5:
```suggestion
  name: tekton-chains
```

operator/gitops/argocd/pipeline-service/tekton-chains/observability-service.yaml
```diff
@@ -0,0 +1,21 @@
...
+  namespace: tekton-chains
+  labels:
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-chains
+    app: tekton-chains-controller
```
sayan-biswas, line 10:
```suggestion
```

operator/gitops/argocd/pipeline-service/tekton-chains/observability-service.yaml
```diff
@@ -0,0 +1,21 @@
...
+spec:
+  ports:
+    - name: http-metrics
```
sayan-biswas, line 13:
```suggestion
    - name: metrics
```

sayan-biswas:
No other configuration are generally required other than the service.
But still incase you want add custom config, the chains section in the tekton config can be used like this.. 

```
chain:
  targetNamespace: openshift-pipelines
  controllerEnvs:
    - name: CONFIG_OBSERVABILITY_NAME  
      value: tekton-chains-config-observability 
```
Refernce:

https://github.com/tektoncd/operator/blob/6fb0477761a82cc71cce6a0efb1a3d1e48c07037/pkg/reconciler/kubernetes/tektonchain/tektonchain.go#L403

avinal:
Thanks Sayan and Romain, I have done the changes.

gabemontero (changes requested):
New error I have not seen  before in the upgrade test @avinal @sayan-biswas  ... the results api deployment is missing:

```
+ timeout 300s bash -c 'while ! kubectl get deployment/tekton-results-api -n tekton-results >/dev/null 2>/dev/null; do printf '\''.'\''; sleep 10; done'
++ printf %4s
+ sed 's/^/    /'
+ printf '%s not found (timeout)\n' tekton-results-api
+ kubectl get deployment/tekton-results-api -n tekton-results
Error from server (NotFound): deployments.apps "tekton-results-api" not found
        - tekton-results-api: ..............................tekton-results-api not found (timeout)
```

Now, that check is https://github.com/openshift-pipelines/pipeline-service/blob/main/operator/images/cluster-setup/content/bin/utils.sh#L94-L100 

And with the series of dots in the output i.e. `..............................` I believe it is waiting the full 5 minutes before giving up.

Any sense @avinal @sayan-biswas if a hiccup in the gitops chains rollout prevents to rollout of results?

I do see that there is no checks for the chains deployments in https://github.com/openshift-pipelines/pipeline-service/blob/main/operator/images/cluster-setup/content/bin/install.sh#L137-L143 ... perhaps something got lost when @Roming22 did the 1.11 change and have the operator install chains.

@avinal please add a check of the chains deployment in install.sh.  Let's then retest and see if chains is in fact getting deployed correctly with this change.

Roming22:
@avinal At the next stand up, let's talk about whether we should be using the mechanism demoed by Gabe to override the chains image to use the one that has the required metrics

avinal:
@gabemontero I don't have any idea if gitops chains rollout is causing any issue. I will try to find that.

For the chains' deployment part, we have multiple tests that directly depend on the deployment, including the new added metrics test. Should I add an explicit check for chains deployment?

gabemontero:
> @gabemontero I don't have any idea if gitops chains rollout is causing any issue. I will try to find that.

that's fine ... speculation on my part, where the deployment check I have asked you to add will help clarify that 
> 
> For the chains' deployment part, we have multiple tests that directly depend on the deployment, including the new added metrics test. Should I add an explicit check for chains deployment?

yess add the chains deployment check ... that check will run before the tests, and be a clearer indication of what is wrong

operator/test/test.sh
```diff
@@ -144,6 +144,13 @@ test_metrics() {
 }
 
 test_chains() {
+  echo -n "  - Check Deployment: "
```
gabemontero, line 147: per clarification/confirmation in stand up this check needs to move to the checks in install.sh around the https://github.com/openshift-pipelines/pipeline-service/blob/main/operator/images/cluster-setup/content/bin/install.sh#L137-L143 area

avinal: In that case it is already there: 
https://github.com/openshift-pipelines/pipeline-service/blob/8619a4e55ac4ffeee4d3fc6d54712aeaf73075e3/operator/images/cluster-setup/content/bin/install.sh#L138
this line checks for all deployment status

avinal: This is what I wanted to point to but couldn't find. I have removed the checks from test.sh

gabemontero:
merging

PR #710, opened 2023-07-12
Fix tekton results structure

sayan-biswas:
## Changes
- Simplify tekton results folder structure 
- Add service monitor config.


avinal (approved):
Looks good to me

gabemontero:
OK @sayan-biswas we have to remember that after this merges, rather than using the auto-generated PR to update infra-deployments, you have to create a manual version of that PR, where you include the same changes it has, but also remove the servicemonitor currently being created in infra-deployments

sayan-biswas:
@gabemontero Yes. Need to remove the service monitor components there and update the kustomize.

PR #713, opened 2023-07-18
Set pipeline enable-api-fields alpha

enarha:
Requested by users to allow use of StepSpecs.

### DIFF
--- a/operator/gitops/argocd/pipeline-service/openshift-pipelines/tekton-config.yaml
+++ b/operator/gitops/argocd/pipeline-service/openshift-pipelines/tekton-config.yaml
@@ -34,7 +34,7 @@ spec:
     transparency.enabled: "false"
   pipeline:
     default-service-account: appstudio-pipeline
-    enable-api-fields: beta
+    enable-api-fields: alpha
     enable-bundles-resolver: true
     enable-cluster-resolver: true
     enable-git-resolver: true


gabemontero:
are there any Jira's or slack threads for reference @enarha ?

Also, IMO @Roming22 this change deserves treatment similar to what we did with 1.11 where we created a test branch, then created a test PR off of infra-deployments to run the e2e's.

Would you agree @Roming22 ?

enarha:
https://issues.redhat.com/browse/PLNSRVCE-1368. The Slack thread and few other Jiras are linked into this one.

gabemontero:
> https://issues.redhat.com/browse/PLNSRVCE-1368. The Slack thread and few other Jiras are linked into this one.

thanks for the ref and memory refresh @enarha 

yeah given recent experiences I definitely think we minimally need to do the pre-merge infra-deployments test PR

probably wouldn't hurt to discuss in general ... in addition to the potential concerns you noted in the Jira, I *think* it is possible other features besides `StepSpecs` gets pulled in ... forget if there is a way to pick and choose which alpha features are pulled in.

@Roming22 @adambkaplan FYI

enarha:
Alpha features are generally not supported. Workaround has been agreed upon. The feature will be available as beta in later version of OSP (probably 1.13).

PR #724, opened 2023-07-31
Refactor update-dependencies 2/2

Roming22:
* Use the existing GitHubApp to manage changes to the repository
* Try to the make the task generic enough so it can be submitted to redhat-appstudio/build-definitions


Roming22:
I ran a test by having the pipeline run on a pull_request, and ended up with #746.
So I believe it works (at least well enough that we can fix bugs in a future PR).

sayan-biswas (approved):

PR #739, opened 2023-08-21
Fix monitoring stack for development

### LINKED ISSUE #738: Monitoring stack for Development setup is broken
The monitoring stack for development is broken because Kustomize is not able to accumulate resources as the following referenced resource doesn't exist in the original repository anymore.

https://github.com/openshift-pipelines/pipeline-service/blob/b1c87b7dc26c5b7f40105f5709c6353601686f04/developer/openshift/gitops/argocd/pipeline-service-o11y/appstudio-prometheus/kustomization.yaml#L6

sayan-biswas:
Fixes #738 

- Fix broken repository references in the kustomize file
- Fix permission for argoCD

### DIFF
--- a/developer/openshift/gitops/argocd/pipeline-service-o11y/appstudio-grafana/allow-argocd-to-manage.yaml
+++ b/developer/openshift/gitops/argocd/pipeline-service-o11y/appstudio-grafana/allow-argocd-to-manage.yaml
@@ -7,7 +7,7 @@ metadata:
     argocd.argoproj.io/sync-wave: "0"
 rules:
   - apiGroups:
-      - integreatly.org
+      - grafana.integreatly.org
     resources:
       - grafanas
       - grafanadashboards
--- a/developer/openshift/gitops/argocd/pipeline-service-o11y/appstudio-prometheus/allow-argocd-to-manage.yaml
+++ b/developer/openshift/gitops/argocd/pipeline-service-o11y/appstudio-prometheus/allow-argocd-to-manage.yaml
@@ -49,7 +49,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: openshift-gitops-manage-secrets
-  namespace: o11y
+  namespace: dummy-service
   annotations:
     argocd.argoproj.io/sync-wave: "0"
 rules:
@@ -69,7 +69,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: openshift-gitops-manage-secrets
-  namespace: o11y
+  namespace: dummy-service
   annotations:
     argocd.argoproj.io/sync-wave: "0"
 roleRef:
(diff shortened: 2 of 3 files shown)


enarha:
Do we need the dummy-service part of the infra-deployments folder we reference? I see there is a `monitoringstack` folder under `components/monitoring/prometheus/development`. Maybe it'll be enough to reference that instead?

sayan-biswas:
> Do we need the dummy-service part of the infra-deployments folder we reference? I see there is a `monitoringstack` folder under `components/monitoring/prometheus/development`. Maybe it'll be enough to reference that instead?

Honestly, we don't need that, but it's harmless and its a part of the upstream infra-deployment repository. If we don't want that to be deployed then we'll have to write additional patches to delete the resources from manifets.

enarha:
> > Do we need the dummy-service part of the infra-deployments folder we reference? I see there is a `monitoringstack` folder under `components/monitoring/prometheus/development`. Maybe it'll be enough to reference that instead?
> 
> Honestly, we don't need that, but it's harmless and its a part of the upstream infra-deployment repository. If we don't want that to be deployed then we'll have to write additional patches to delete the resources from manifests.
> 
> We'll need this I believe. https://github.com/redhat-appstudio/infra-deployments/blob/5ce93c6ed32471f6e1ba609ee9e0d1eea1cfbad7/components/monitoring/prometheus/development/kustomization.yaml#L4

It's not exactly harmless as it managed to break us :-). The fact that we reference something, but also manage some dependency of it locally is fragile as we've seen. Anyway, lets merge the fix and I'll ask the monitoring team to make the the monitoring stack "self contained" so we can only rely on import.

/lgtm

sayan-biswas:
> > > Do we need the dummy-service part of the infra-deployments folder we reference? I see there is a `monitoringstack` folder under `components/monitoring/prometheus/development`. Maybe it'll be enough to reference that instead?
> > 
> > 
> > Honestly, we don't need that, but it's harmless and its a part of the upstream infra-deployment repository. If we don't want that to be deployed then we'll have to write additional patches to delete the resources from manifests.
> > We'll need this I believe. https://github.com/redhat-appstudio/infra-deployments/blob/5ce93c6ed32471f6e1ba609ee9e0d1eea1cfbad7/components/monitoring/prometheus/development/kustomization.yaml#L4
> 
> It's not exactly harmless as it managed to break us :-). The fact that we reference something, but also manage some dependency of it locally is fragile as we've seen. Anyway, lets merge the fix and I'll ask the monitoring team to make the the monitoring stack "self contained" so we can only rely on import.
> 
> /lgtm

Agreed!!
We need to ask them  to put the `configmap` to enable user workload monitoring in the  `monitoringstack` folder.

gabemontero:
Yeah when we start getting close to pipeline-service operating outside the context of RHTAP, the entire argocd/pipeline-service-o11y path will need to be reworked or made optional somehow @Roming22 @enarha @sayan-biswas 

But per prior exchanges here I agree with @sayan-biswas @enarha let's merge this and move on to the next steps they noted.

PR #740, opened 2023-08-21
Fix HTTP2 error with openshift routes

sayan-biswas:
This is a workaround for the bug [PLNSRVCE-1210](https://issues.redhat.com/browse/PLNSRVCE-1210)

This error is appearing in the API service logs because when we are creating a route in an Openshift cluster with HTTP2 enabled, HAPROXY opens  TCP connection to that route for health check, and maybe closes the connection before the handshake is complete. The `net` golang package logs these requests.
But I have observed that these issue doesn't happen if we separate the ports in different service (not sure why this behaviour)

### DIFF
--- a/operator/test/test.sh
+++ b/operator/test/test.sh
@@ -288,6 +288,16 @@ test_security() {
 }
 
 test_results() {
+  # Check logs for OCP bug https://issues.redhat.com/browse/OCPBUGS-5916
+  printf "\n  - Check HTTP2 health probe errors: "
+  pattern="http2: server: error reading preface from client"
+  if kubectl logs deployment/tekton-results-api -c "api" -n "$NAMESPACE" 2>/dev/null | grep -ciq "$pattern"; then
+    echo "Failed"
+    exit 1
+  else
+    echo "OK"
+  fi
+
   test_pipelines
   echo -n "  - Results in database:"
 
--- a/developer/openshift/dev_setup.sh
+++ b/developer/openshift/dev_setup.sh
@@ -153,6 +153,14 @@ check_cluster_role() {
   fi
 }
 
+cluster_setup() {
+  echo "[cluster-setup]"
+
+  # By default HTTP2 is not enabled in test openshift clusters
+  echo "- Enabling HTTP2 for ingress" | indent 2
+  oc annotate ingresses.config/cluster ingress.operator.openshift.io/default-enable-http2=true | indent 6
+}
+
 install_openshift_gitops() {
   APP="openshift-gitops"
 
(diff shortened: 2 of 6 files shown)


gabemontero (commented):
@sayan-biswas were the repeating http2 logs showing up in the api server instance running during our CI?  In other words, is enough of the HAPROXY setup done in our CI to reproduce this?

If the answer is "yes" to any of that, it would be nice to update our tests here to dump the api server log and confirm the repeating error is no longer occurring.

Roming22 (approved):

ramessesii2 (commented):
lgtm. I'll wait for Sayan's reply on Gabe's ask.

sayan-biswas:
> @sayan-biswas were the repeating http2 logs showing up in the api server instance running during our CI? In other words, is enough of the HAPROXY setup done in our CI to reproduce this?
> 
> If the answer is "yes" to any of that, it would be nice to update our tests here to dump the api server log and confirm the repeating error is no longer occurring.

No, it doesn't. Probably that's why we have not noticed. This was reported by Adam as it was prominent in staging and prod setup. However to get this reproducible in our dev setup we need to enable the HTTP2, which is not enabled by default.

`oc annotate ingresses.config/cluster ingress.operator.openshift.io/default-enable-http2=true`

I'll add this in the dev setup script. Also I'll add some test if possible.

operator/test/test.sh
```diff
@@ -288,6 +288,13 @@ test_security() {
 }
 
 test_results() {
+  # Check logs for OCP bug https://issues.redhat.com/browse/OCPBUGS-5916
+  pattern="http2: server: error reading preface from client +.+ read: connection reset by peer"
+  if kubectl logs deployment/tekton-results-api -c "api" -n "$NAMESPACE" 2>&1 | grep -ciqE "$pattern"; then
```
gabemontero, line 293: I assume this `grep` is right, but I have no idea :-) 

You got a quick sample to confirm @sayan-biswas ?

Or, was there a reason checking for `http2: server: error reading preface from client` was not sufficient ?

gabemontero (commented):
Just a minor question @sayan-biswas otherwise looks great thanks!

@Roming22 FYI

operator/test/test.sh
sayan-biswas: Ya, its right, tested it. I used as a regular expression.
But I'll update and just put `http2: server: error reading preface from client` instead, to catch broader errors. I guess the rest of the messages are unwrapped errors from other methods down the stack.
I added few messages also, so I am pushing the PR again.

sayan-biswas: done

gabemontero (approved):
/lgtm

PR #743, opened 2023-08-24
Use pac-gitauth secret for push

Roming22:
This PR simplifies the push to GitHub by using the token properly. 

Because of issues pushing to the repository, I initially thought that I would need to send the changes through the GitHub API. This would have required the detailed information I was logging in `.commits.json`.

Since I've now managed to perform a `git push` to the repository, the task can be greatly simplified.

### DIFF
--- a/.tekton/tasks/update-repository.yaml
+++ b/.tekton/tasks/update-repository.yaml
@@ -98,6 +98,8 @@ spec:
           value: $(params.TARGET_GH_NAME)
         - name: TARGET_GH_OWNER
           value: $(params.TARGET_GH_OWNER)
+        - name: WORKSPACE_BASIC_AUTH_DIRECTORY_PATH
+          value: $(workspaces.basic-auth.path)
       args: ["$(params.SCRIPT_ARGS[*])"]
       script: |
         #!/bin/bash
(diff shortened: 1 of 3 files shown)


.tekton/tasks/update-repository.yaml
```diff
@@ -116,77 +118,22 @@ spec:
...
 
         # Create branch
         git branch --copy --force "$COMMIT_BRANCH"
         git checkout "$COMMIT_BRANCH"
 
         # Run script
-        UPSTREAM_COMMIT=$(git rev-parse HEAD)
         "${SCRIPT_PATH}" "${SCRIPT_ARGS[@]}"
 
-        # Log changes
```
gabemontero, line 128: If I read this correctly @Roming22 we were not even pushing at all before but building a list of the changes.

Is that correct?  

I think adding some additional details around this in the PR description about this, and *why* we want to move from building the log to a push, would be helpful.

Thanks.

Roming22: That's correct. I had issues pushing to the repo I thought that I would need to send the changes through the GitHub API, for which I needed the detailed information I was logging in `.commits.json`.

Since I've since then managed to perform a `git push` to the repository, the task can be greatly simplified.

gabemontero (approved):
/lgtm

are these post merge pac's that run @Roming22 ?

Feel free to merge and then post verification confirmation as needed.

Roming22:
@gabemontero Correct. I'm still in the process of developing the pipeline though. I still need to create/update the PR.
I just wanted to give bite-sized PRs instead of a big PR that covers a lot of ground.

PR #750, opened 2023-08-30
Remove refs to PaC namespace

enarha:
With PaC being installed by the OSP operator, we no longer need the pipelines-as-code namespace. Here we cleanup references to that namespace from the reset and test scripts.


gabemontero:
all green - theory confirmed @enarha :-)

gabemontero (approved):
/lgtm

PR #752, opened 2023-08-30
Fix podman image tag used in CI

enarha:
We were using the v4.5.1 tag and it's been replaced by v4.6.1. Those both are floating tags and not static ones, so it does not make much difference if we start using a bit more generic one v4.6 while still preserving the versions of the extra rpms installed. Same happened with the skopeo image, replaced with the new one.

### DIFF
--- a/developer/images/devenv/Dockerfile
+++ b/developer/images/devenv/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/podman/stable:v4.5.1
+FROM quay.io/podman/stable:v4.6
 RUN set -x \
     && mkdir ~/.kube \
     && mkdir -p /tmp/image-build \
--- a/operator/test/manifests/test/tekton-chains/simple-copy-pipeline.yaml
+++ b/operator/test/manifests/test/tekton-chains/simple-copy-pipeline.yaml
@@ -32,7 +32,7 @@ spec:
             description: Reference of the image skopeo will push.
           - name: SKOPEO_IMAGE
             description: The location of the skopeo image.
-            default: quay.io/skopeo/stable:v1.9.0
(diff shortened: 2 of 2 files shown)


gabemontero:
unrelated deploy cluster failure: 

```
++ kubectl get co console -o 'jsonpath={.status.conditions[?(@.type=="Available")].status}'
Unable to connect to the server: dial tcp: lookup api.ci-362de3957c.zffy.p3.openshiftapps.com on 172.30.0.10:53: no such host
+ available=
```

/retest

gabemontero:
OK this time we got past the cluster deploy, and pipeline service set up failed with 
```
- Deploy applications:
    - No Bitwarden secrets
```

This is presumably a AWS/Rosa flake

On the upgrade test, still see chain error noted in the slack thread
```
[chains]
  - Run pipeline: error: timed out waiting for the condition on pipelineruns/simple-copy-run-75v6g
 ```
but I was able to confirm via the pipelinerun/taskrun yaml on the failures, before they got pruned, that our CI code is still using `main` branch for all the ci images.  i.e. as we dsicussed in the slack thread, if we are convinced the CI image is the root cause, then force merging this and then retesting other PRs once the image update is available would be the path.

I'll retest to see about consistency during @enarha 's off hours.  Go ahead and force merge @enarha when you get in on Thursday if you still feel fixing the image is the required next step.

enarha:
The images used in the Dockerfile and the manifests are gone, so that's something we need to update. I'm not 100% sure that's going to fix the chain issue, I think it does and it worked locally, so I'm going to merge this PR and re-try another one. If we need to follow up with more changes here, we'll do in a separate PR.

PR #754, opened 2023-08-31
allow ingress annotation overwrite

ramessesii2:
Re-running the dev_setup.sh script results into an error:
    ```
    error: --overwrite is false but found the following declared
    annotation(s): 'ingress.operator.openshift.io/default-enable-http2'
    already has a value (true)
    ```
This PR aims to fix this error.


sayan-biswas (approved):

gabemontero (approved):
/lgtm

PR #758, opened 2023-09-06
fix issue that pipeline-service setup task take the correct revision

xinredhat:


ramessesii2 (approved):

gabemontero:
CI is green @Roming22 @xinredhat 

should we convert this from draft to reviewable PR and get this merged?

Roming22:
I don't think this PR should be merged. Have a look at #761, and you will see that we still would not have the expected behavior.

#760 might be a better fix.

.tekton/pipeline/upgrade-tests.yaml
```diff
@@ -90,7 +90,7 @@ spec:
         - name: repo_url
           value: $(params.repo_url)
         - name: revision
-          value: $(params.revision)
+          value: $(params.target_branch)
```
Roming22, line 93: I don't think that is the right fix. If changes are done to the script in the `ci` folder, they won't be tested unless the version checked out is the `revision`.

I believe the proper fix is to checkout the right branch before every test.

Roming22 (changes requested):

.tekton/pipeline/upgrade-tests.yaml
xinredhat: the each git-clone task does checkout the right branch before every test, so the `revision` should align with the git-clone task, for example https://github.com/openshift-pipelines/pipeline-service/blob/main/.tekton/pipeline/upgrade-tests.yaml#L66

.tekton/pipeline/acceptance-tests.yaml
```diff
@@ -52,17 +63,6 @@ spec:
           workspace: shared-workspace
         - name: source
           workspace: source
-    - name: clone-pipeline-service-git
```
Roming22, line 55: Not sure why this change is necessary.

.tekton/tasks/ci-runner-container-setup.yaml
```diff
@@ -22,15 +22,15 @@ spec:
...
-        requests:
-          memory: 500Mi
-          cpu: 300m
-      workingDir: "$(workspaces.source.path)"
-      env:
-        - name: KUBECONFIG
-          value: "$(workspaces.kubeconfig-dir.path)/kubeconfig"
-      command:
-        - $(workspaces.source.path)/ci/images/ci-runner/hack/bin/copy-plnsvc-code.sh
+    # - name: copy-plnsvc-code
```
Roming22, line 25: Delete.

ci/images/ci-runner/hack/bin/copy-plnsvc-code.sh
```diff
@@ -5,7 +5,9 @@ set -o pipefail
 set -x
 
 echo "Copy source code of pipeline service to the ci-runner container"
+kubectl exec -n default -it ci-runner -- bash -c "rm -rf /source/*"
```
Roming22, line 8: I'm expecting this to fail on the very first run, thus ending the PLR in failure. Check for the folder existence before deletion.

Same for the next line.

Roming22 (changes requested):
That's better but still likely to fail. 

I haven't wrapped my head around it yet, but I am wondering if the content of `ci/images/ci-runner/hack/sidecar` needs to always point to the PR content on the ci-runner container.

.tekton/pipeline/acceptance-tests.yaml
xinredhat: Yes, I just moved the block to the front. Because other tasks run after it. this movement is  helpful to understand the whole pipeline logic.

.tekton/tasks/ci-runner-container-setup.yaml
xinredhat: done.

xinredhat:
> but I am wondering if the content of `ci/images/ci-runner/hack/sidecar` needs to always point to the PR content on the ci-runner container.

I don't quite understand why you are considering it. the content of `ci/images/ci-runner/hack/sidecar` is part of the repo, it depends on which branch the git-clone task check out.

ci/images/ci-runner/hack/bin/copy-plnsvc-code.sh
xinredhat: I am confusing why you are expecting this to fail on the very first run, because directory `/source` already is there , it always succeed no matter if there are files or not under  `/source/`

Roming22:
@xinredhat  Can we close this PR since #760 has solved the issue?

PR #759, opened 2023-09-06
Fix http2 errors in tekton-results API server

sayan-biswas:
This is a workaround for the bug [PLNSRVCE-1210](https://issues.redhat.com/browse/PLNSRVCE-1210)

This error is appearing in the API service logs because when we are creating a route in an Openshift cluster with HTTP2 enabled, HAPROXY opens TCP connection to that route for health check, and maybe closes the connection before the handshake is complete. The net golang package logs these requests.

This workaround increases the time interval between the health probes, which prevents this error from flooding tekton-results API server logs.


gabemontero (commented):
Are you able to reproduce the issue when you infra-deployments bootstrap a cluster you start up, or can this only be seen in staging/prod @sayan-biswas ?  Or if we look at the logs for results when running the e2e's from infra-deployments, we'll see the error?

If you can repro "locally", this could be another candidate for verifying this via the infra-deployment local boostrap or PR path.

sayan-biswas:
@gabemontero Reproducible in staging/prod and locally also. I'll check infra-deployment cluster and confirm.

PR #760, opened 2023-09-06
Fix CI update-test issue

Roming22:


.tekton/tasks/pipeline-service-tests.yaml
```diff
@@ -6,10 +6,25 @@ metadata:
 spec:
   params:
     - name: target_branch
+    - name: branch
   workspaces:
     - name: kubeconfig-dir
     - name: source
   steps:
+    - name: checkout-plnsvc-code
```
xinredhat, line 14: This step is not needed, because the cloned codes already is copy to docker container at this [step](https://github.com/openshift-pipelines/pipeline-service/blob/main/.tekton/tasks/ci-runner-container-setup.yaml#L36)

ci/images/ci-runner/hack/bin/copy-plnsvc-code.sh
```diff
@@ -4,6 +4,8 @@ set -o nounset
 set -o pipefail
 set -x
 
+echo "[DEBUG:PLR] This message must ALWAYS be seen in the PLR logs"
+
 echo "Copy source code of pipeline service to the ci-runner container"
 kubectl cp "./" "default/ci-runner:/source"
```
xinredhat, line 10: this script is not needed, same as https://github.com/openshift-pipelines/pipeline-service/pull/760/files#r1318056837

.tekton/tasks/pipeline-service-tests.yaml
Roming22: This is the bit that will make sure that the content on the ci-runner matches the version we want to test.

ci/images/ci-runner/hack/bin/copy-plnsvc-code.sh
Roming22: We can discuss the need for the script in the first comment. Resolving this one.

.tekton/tasks/pipeline-service-tests.yaml
xinredhat: ok, I updated my PR https://github.com/openshift-pipelines/pipeline-service/pull/758, i think it should solve the issue.

Roming22:
The latest version should fix the CI properly:
* The version of the repository checkout on the PipelineRun pod is always the PR version
* The version of the repository checkout on the remote ci-runner node is the one that is being deployed/tested
* The CI scripts used on the remote ci-runner are always the PR version

This should ensure that:
* Any changes to the CI scripts or pipeline definitions are always taken into account no matter at which step of the pipeline we are.
* Any call to our dev/test scripts matches the current version under test

@xinredhat @ramessesii2 Please review.

Here's the summary of the changes:
    Fix upgrade-tests pipeline
    
    The wrong set of scripts were used when deploying/testing the
    baseline branch. Scripts from the baseline branch are now used
    when testing the baseline branch, and scripts from the PR are
    used when testing the PR branch.
    
    * Fix issue with a sidecar script that was not called from the sidecar.
    * Change target of source code copy to /workspace/source on the sidecar
    * Copy sidecar scripts to /workspace/sidecar on the sidecar
    * Checkout the revision being tested before setup and test
    * Rename ci-runner-container-setup.yaml to setup-ci-runner-container.yaml
    so filename and taskname match

ci/images/ci-runner/hack/sidecar/bin/plnsvc_setup.sh
```diff
@@ -11,6 +11,10 @@ REPO_REVISION=$2
 # to avoid the issue of "fatal: detected dubious ownership in repository at '/workspace/source'"
 git config --global --add safe.directory "$PWD"
 
+# Checkout the branch we want to setup
+git fetch origin "$REPO_REVISION"
+git checkout "$REPO_REVISION"
```
xinredhat, line 16: Honestly, I prefer using git-clone task as a common task to handle it

ci/images/ci-runner/hack/bin/run-plnsvc-test.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+set -o errexit
+set -o nounset
+set -o pipefail
+set -x
+
+echo "Run pipeline-service tests..."
+if kubectl -n default exec pod/ci-runner -- \
+    sh -c "/workspace/sidecar/bin/plnsvc_test.sh"; then
```
xinredhat, line 9: can we move the content of `/workspace/sidecar/bin/plnsvc_test.sh` to here? then we could delete ``/workspace/sidecar/bin/plnsvc_test.sh``

xinredhat:
others look good to me

ramessesii2:
Thanks for the PR Romain.
/lgtm

ci/images/ci-runner/hack/sidecar/bin/plnsvc_setup.sh
Roming22: My understanding is that using git-clone is inefficient.
* We'll need to clear the destination then copy the whole content over.
* I believe the directory might be deleted before the clone, so we're going to effectively clone the repo 3 times instead of one.

ci/images/ci-runner/hack/bin/run-plnsvc-test.sh
Roming22: The sidecar is the place where the version of the code that is tested is located.
If we run the tests from the TaskRun pod, then the pod would need to start switching between branches, which causes issues when the CI scripts are modified. E.g. if we change the `pipeline-service-setup` Task to a new step that references a new script, then the task would fail after we've checked out the target version that does not have the script.

PR #770, opened 2023-09-15
Add errexit,nounset,pipefail to chains-secrets-config job

Roming22:
This will ensure that any failure to setup the secret is properly reported in the job status.

### DIFF
--- a/operator/gitops/argocd/pipeline-service/openshift-pipelines/chains-secrets-config.yaml
+++ b/operator/gitops/argocd/pipeline-service/openshift-pipelines/chains-secrets-config.yaml
@@ -61,37 +61,43 @@ spec:
             - /bin/bash
             - -c
             - |
+              set -o errexit
+              set -o nounset
+              set -o pipefail
+
+              namespace="openshift-pipelines"
+              secret="signing-secrets"
+
               cd /tmp
-              # Try to handle that nicely. The object is expected to always exist so check the data.
-              SIG_KEY_DATA=$(kubectl get secret signing-secrets -n openshift-pipelines -o jsonpath='{.data}')
-              if [[ -n $SIG_KEY_DATA ]]; then
+
+              if [ "$(kubectl get secret "$secret" -n "$namespace" -o jsonpath='{.data}' --ignore-not-found --allow-missing-template-keys)" != "" ]; then
                 echo "Signing secret exists and is non-empty."
               else
                 # Delete secret/signing-secrets if already exists since by default cosign creates immutable secrets
-                kubectl delete secrets -n openshift-pipelines signing-secrets --ignore-not-found=true
+                kubectl delete secrets "$secret" -n "$namespace" --ignore-not-found=true
 
                 # To make this run conveniently without user input let's create a random password
-                RANDOM_PASS=$( head -c 12 /dev/urandom | base64 )
+                RANDOM_PASS=$( openssl rand -base64 30 )
 
                 # Generate the key pair secret directly in the cluster.
                 # The secret should be created as immutable.
-                echo "Generating k8s secret/signing-secrets with key-pair"
-                env COSIGN_PASSWORD=$RANDOM_PASS cosign generate-key-pair k8s://openshift-pipelines/signing-secrets
+                echo "Generating k8s secret/$secret in $namespace with key-pair"
+                env COSIGN_PASSWORD=$RANDOM_PASS cosign generate-key-pair "k8s://$namespace/$secret"
               fi
 
               # If the secret is not marked as immutable, make it so.
-              if [ "$(kubectl get secret signing-secrets -n openshift-pipelines -o jsonpath='{.immutable}')" != "true" ]; then
+              if [ "$(kubectl get secret "$secret" -n "$namespace" -o jsonpath='{.immutable}')" != "true" ]; then
                 echo "Making secret immutable"
-                kubectl get secret signing-secrets -n openshift-pipelines -o yaml \
-                | yq eval '.immutable = true' - \
+                kubectl patch secret "$secret" -n "$namespace" --dry-run=client -o yaml \
+                  --patch='{"immutable": true}' \
                 | kubectl apply -f -
               fi
 
               echo "Generating/updating the secret with the public key"
               kubectl create secret generic public-key \
-                --namespace openshift-pipelines \
+                --namespace "$namespace" \
                 --from-literal=cosign.pub="$(
-                  cosign public-key --key k8s://openshift-pipelines/signing-secrets
+                  cosign public-key --key "k8s://$namespace/$secret"
                 )" \
                 --dry-run=client \
                 -o yaml | kubectl apply -f -


zregvart (approved):
I approve this change 😸

ramessesii2:
/lgtm
Will merge once the tests pass. The last CI failures were unrelated since they complained the old - `results watcher not found`.

ramessesii2:
pipeline-service-test failed:
```
- tekton-results-api: Exists, Ready
        - tekton-results-watcher: ..............................tekton-results-watcher not found (timeout)
```

ramessesii2:
There seems to be a sort of CI outage?
```
There was an issue validating the commit: "could not validate payload, check your webhook secret?: payload signature check failed"
```

xinredhat:
> There seems to be a sort of CI outage?
issue https://issues.redhat.com/browse/RHTAPBUGS-809

xinredhat:
> pipeline-service-test failed:
> 
> ```
> - tekton-results-api: Exists, Ready
>         - tekton-results-watcher: ..............................tekton-results-watcher not found (timeout)
> command terminated with exit code 1
> ```

I still faced the same issue when I manually installed pipeline-service

ramessesii2:
> I also faced the same issue when I manually installed pipeline-service

Thanks @xinredhat for trying out locally. 
Can you share more info on this one? Is just `results-watcher` out-of-sync? I'm asking this because there's usually one or more service failing before Tekton-Results which ultimately leads to Results-watcher failing to sync.

xinredhat:
> I'm asking this because there's usually one or more service failing before Tekton-Results which ultimately leads to Results-watcher failing to sync.

no, actually Results-watcher was not created. I am investigating it.

xinredhat:
I found the following error when I was debugging it locally
logs of pod `tekton-chains-signing-secret-jswc8`:
 ````
I0920 10:06:34.643307       9 request.go:665] Waited for 1.170854761s due to client-side throttling, not priority and fairness, request: GET:https://172.30.0.1:443/apis/coordination.k8s.io/v1?timeout=32s
3
I0920 10:06:44.842652       9 request.go:665] Waited for 11.369199171s due to client-side throttling, not priority and fairness, request: GET:https://172.30.0.1:443/apis/pipelines.openshift.io/v1alpha1?timeout=32s
2
Error: updating secret signing-secrets in ns openshift-pipelines: secrets "signing-secrets" is forbidden: cannot set an ownerRef on a resource you can't delete: , <nil>
1
main.go:46: error during command execution: updating secret signing-secrets in ns openshift-pipelines: secrets "signing-secrets" is forbidden: cannot set an ownerRef on a resource you can't delete: , <nil>

```

File: operator/gitops/argocd/pipeline-service/openshift-pipelines/chains-secrets-config.yaml [hunk 1]
```diff
@@ -60,6 +60,10 @@ spec:
             - /bin/bash
             - -c
             - |
+              set -o errexit
+              set -o nounset
+              set -o pipefail
```
ramessesii2, line 65: I don't think it's a good idea to have these. 
For e.g. - we've statements in this script which are bound to fail in certain cases like -
The line SIG_KEY_DATA=$(kubectl get secret signing-secrets -n openshift-pipelines -o jsonpath='{.data}') might return a non-zero exit code if the secret doesn't exist. With set -o errexit, this will cause the entire script to fail.
This is why I think the CI is failing. 
Moreover, since we've `restartPolicy: OnFailure` on the job, it will lead to infinite loops of pod restarts.

ramessesii2 (commented):
I think the crux of the issue is `public-key` is empty because something goes amiss with `signing-secrets` creation and so `public-key` never gets the right  - this is what I'm getting trying it locally; but yes, since we didn't have `set -x errexit` in the job script it didn't fail but rather created empty `public-key`.

operator/gitops/argocd/pipeline-service/openshift-pipelines/chains-secrets-config.yaml
zregvart: Setting these makes the script more robust, without them only the status of the last command is reported as the exit status. For commands that can return non zero status proper error handling should be implemented.

ramessesii2: Yes, agreed.

In that case, we should consider handling the exit codes, for .eg. -
`SIG_KEY_DATA=$(kubectl get secret signing-secrets -n openshift-pipelines -o jsonpath='{.data}') || true`

Or the other option suggested above: Since the restartPolicy is `OnFailure`, we handle this logic in the script itself  and exit with a status code of 0 to prevent unnecessary restarts which is how things are done currently.

Btw I've another PR raised which should fix `public-key` being empty issue - https://github.com/openshift-pipelines/pipeline-service/pull/776

Roming22: @ramessesii2
1. I've refactored the script so that `SIG_KEY_DATA` is not an issue anymore.
2. Do you think we need to change the restartPolicy? `OnFailure` makes sense to me as we would try to setup the secret until we can (e.g. in the event some RBAC that are needed aren't applied on the first try).

ramessesii2: Thanks Romain for the refactoring. 
`SIG_KEY_DATA` value was just an example. But we're good since the [default number of container restarts in case of back-off failure due to logical errors has a limit of 6](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy). I didn't notice this before due to exponential back-off delay. 
So, yes I agree with continuing with an `OnFailure` restartPolicy.

ramessesii2 (approved):

operator/gitops/argocd/pipeline-service/openshift-pipelines/chains-secrets-config.yaml
```diff
@@ -61,37 +61,43 @@ spec:
...
+
+              namespace="openshift-pipelines"
+              secret="signing-secrets"
+
               cd /tmp
-              # Try to handle that nicely. The object is expected to always exist so check the data.
-              SIG_KEY_DATA=$(kubectl get secret signing-secrets -n openshift-pipelines -o jsonpath='{.data}')
-              if [[ -n $SIG_KEY_DATA ]]; then
+
+              if [ "$(kubectl get secret "$secret" -n "$namespace" -o jsonpath='{.data}')" != "" ]; then
```
zregvart, line 73: This will not exit with non zero and output an empty string when the secret is missing or the `data` key is empty

```suggestion
              if [ "$(kubectl get secret "$secret" -n "$namespace" -o jsonpath='{.data}' --ignore-not-found --allow-missing-template-keys)" != "" ]; then
```

operator/gitops/argocd/pipeline-service/openshift-pipelines/chains-secrets-config.yaml
```diff
@@ -61,37 +61,43 @@ spec:
...
                 echo "Signing secret exists and is non-empty."
               else
                 # Delete secret/signing-secrets if already exists since by default cosign creates immutable secrets
-                kubectl delete secrets -n openshift-pipelines signing-secrets --ignore-not-found=true
+                kubectl delete secrets "$secret" -n "$namespace" --ignore-not-found=true
 
                 # To make this run conveniently without user input let's create a random password
                 RANDOM_PASS=$( head -c 12 /dev/urandom | base64 )
```
zregvart, line 80: A bit stronger passphrase and uses single command

```suggestion
                RANDOM_PASS=$( openssl rand -base64 30 )
```

operator/gitops/argocd/pipeline-service/openshift-pipelines/chains-secrets-config.yaml
```diff
@@ -61,37 +61,43 @@ spec:
...
               fi
 
               # If the secret is not marked as immutable, make it so.
-              if [ "$(kubectl get secret signing-secrets -n openshift-pipelines -o jsonpath='{.immutable}')" != "true" ]; then
+              if [ "$(kubectl get secret "$secret" -n "$namespace" -o jsonpath='{.immutable}')" != "true" ]; then
                 echo "Making secret immutable"
-                kubectl get secret signing-secrets -n openshift-pipelines -o yaml \
+                kubectl get secret "$secret" -n "$namespace" -o yaml \
                 | yq eval '.immutable = true' - \
                 | kubectl apply -f -
```
zregvart, line 93: Using kubectl patch will remove the dependency on `yq`
```suggestion
                kubectl patch secret "$secret" -n "$namespace" --dry-run=client --patch='{"immutable": true}' -o yaml \
                | kubectl apply -f -
```

zregvart (commented):
Some suggestions...

operator/gitops/argocd/pipeline-service/openshift-pipelines/chains-secrets-config.yaml
Roming22: Because it is in an `if` statement, the failure was not triggering the script to exit. This is confirmed by the tests being successful.

I will still add the options because it does not harm either and is more explicit.

operator/gitops/argocd/pipeline-service/openshift-pipelines/chains-secrets-config.yaml
Roming22: Done

operator/gitops/argocd/pipeline-service/openshift-pipelines/chains-secrets-config.yaml
Roming22: Done

PR #774, opened 2023-09-19
Add Sprayproxy Grafana panels

enarha:
Sprayproxy is running on host clusters and actual data will only be visualized there. On the member clusters no data will be visualized.


enarha:
preview

Roming22 (approved):

PR #791, opened 2023-10-02
Add CI  to test pipeline-service on OCP 4.13

xinredhat:


xinredhat:
pipeline-service-test-newer-ocp  failed due to the follow reason
```
  - message: 'constraints not satisfiable: no operators found in channel stable of
      package openshift-gitops-operator in the catalog referenced by subscription
      openshift-gitops-operator, subscription openshift-gitops-operator exists'
```

gabemontero (changes requested):
on suggestion @xinredhat  .... curious @openshift-pipelines/pipelines-service wrt running as a service ramifications down the line wrt specific OCP versions

.tekton/test-pipeline-service-deployment-newer-ocp.yaml
```diff
@@ -0,0 +1,40 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: test-pipeline-service-deployment-newer-ocp
```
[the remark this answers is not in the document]
Roming22: I agree. It will also avoid a "version dance" if/when we change supported version (i.e. 4.13 won't need to be replace in the original pipeline and 4.14 won't need to be added to the "newer" pipeline, the only thing that happens is a file is added/deleted).

.tekton/pipeline-service-upgrade-test.yaml
```diff
@@ -10,20 +10,15 @@ metadata:
...
+    pipelinesascode.tekton.dev/task-5: "[.tekton/tasks/test-pipeline-service.yaml]"
+    pipelinesascode.tekton.dev/task-6: "[git-clone]"
     pipelinesascode.tekton.dev/max-keep-runs: "5"
 spec:
   pipelineRef:
     name: test-pipeline-service-upgrade
   params:
-    - name: aws_region
-      value: "us-east-1"
-    - name: ocp_version
```
Roming22, line 25: Why don't we keep the version as an explicit param?

xinredhat: I would like to define them in one place, such as task [`deploy-cluster`](https://github.com/openshift-pipelines/pipeline-service/blob/2f35a207fed5a79a1116c35390eb7b83951bc0b5/.tekton/tasks/deploy-cluster.yaml#L8-L13), other pipelines can reference this task and keep the same version and AWS region

.tekton/test-pipeline-service-deployment-newer-ocp.yaml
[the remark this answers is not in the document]
xinredhat: fixed.

.tekton/pipeline-service-upgrade-test.yaml
Roming22, line 1: Can you please rename it `.tekton/test-pipeline-service-upgrade-ocp-412.yaml` so it matches the other filenames?

Roming22:
I've added a minor comment. I'm OK with the PR being merged once the comment is resolved.

.tekton/pipeline-service-upgrade-test.yaml
xinredhat: fixed

PR #793, opened 2023-10-02
Update operator docs

Roming22:
Documentation deprecated when we moved out of KCP.


operator/docs/operations.md
```diff
@@ -33,14 +33,13 @@ Pipeline Service also employs vulnerability scanning to detect and address poten
...
+We use Tekton Pipelines to run our tests. All our tasks, pipelines, and pipelineruns live in the [.tekton folder](../../.tekton) in our repository. This approach allows us to define the test pipelines in a declarative way, making it easy to maintain and modify. The pipelines are executed through Pipeline Service (a.k.a dogfooding). This approach provides an opportunity to identify any potential issues with the service itself and address them before they affect our users.
 
 ### Upgrade Strategy for Operators
 
-Pipeline Service uses Operators, specifically OpenShift Pipelines and OpenShift GitOps, to manage the deployment and configuration of its components and this will continue to be the primary way to handle upgrades. However, some components of Pipeline Service are standalone at the moment and are installed/upgraded using upstream manifests.
+Pipeline Service uses Operators, specifically OpenShift Pipelines and OpenShift GitOps, to manage the deployment and configuration of its components and this will continue to be the primary way to handle upgrades.
```
gabemontero, line 40: should we mention our use of the grafana operator ?

gabemontero (changes requested):

operator/docs/operations.md
gabemontero: I should clarify ... at the moment, our install of grafana is "developer only" as RHTAP installs grafana, but in theory, when we are dealing with non-RHTAP as a service offerings, we'll conceivably want to install grafana, no?

Also, minio is certainly more "developer only", so I'm not suggesting adding that, but it does bring up the question of how we deal with AWS storage when we run as a service.  That bit I admittedly is a follow up form this PR item though.

Not as certain with grafana

Roming22: I've added a blurbed. If that's good enough, you can resolve the conversation.

gabemontero: sorry where did you add the blurb?  I don't see any updates to the line I attached the comment to.

Was it that line in the Monitoring section?

Roming22: Yes, I've added the monitoring section.

Roming22:
Merging, as all the changes are doc related, and the CI is currently down.

PR #794, opened 2023-10-03
DO-NOT-MERGE test OSP nightly deployment in CI

enarha:
This is to test the OSP nightly deployment in the Pipeline Service CI.


enarha:
I expect this to fail due inability to apply the ImageContentSourcePolicy manifest, yet I need a confirmation before digging deeper. It fails when manually deploying on ROSA cluster.

PR #803, opened 2023-10-06
Update CI HCP configuration

Roming22:
This is the configuration recommended by HCP.

### DIFF
--- a/ci/images/ci-runner/hack/bin/deploy-cluster.sh
+++ b/ci/images/ci-runner/hack/bin/deploy-cluster.sh
@@ -8,7 +8,7 @@ SCRIPT_DIR="$(
   pwd
 )"
 
-AWS_OIDC_CONFIG_ID="26mo8fa481kctu0oag5ubd7l0kctk9ut"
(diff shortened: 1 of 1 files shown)


xinredhat:
@Roming22 Based on the following error,  you also need to create unique  `operator-roles` for each unique oidc-config. 
```
INFO: Using arn:aws:iam::400884778152:role/ManagedOpenShift-Worker-Role for the Worker role
INFO: Using arn:aws:iam::400884778152:role/ManagedOpenShift-Support-Role for the Support role
ERR: Operator role 'arn:aws:iam::400884778152:role/plnsvc-ci-kube-system-kube-controller-manager' does not have trusted relationship to 'rh-oidc.s3.us-east-1.amazonaws.com/26olrg3eggjh8uioks9hfhgd6mo7oj2m' issuer URL
```

To make it simple, we can take Guilherme's suggestion because in our case it's using a  managed OIDC config
```
Considering it is a CI for testing the application I'd say sharing a single OIDC config is valid, we also do so within our CI.
On our case we generate a managed OIDC config at the start of the CI, use it for all HCP clusters and remove it at the end. 
```

PR #821, opened 2023-10-12
fix issues of pipeline-service test

xinredhat:
This PR is to fix issue of  [PLNSRVCE-1492](https://issues.redhat.com/browse/PLNSRVCE-1492)

### DIFF
--- a/developer/openshift/dev_setup.sh
+++ b/developer/openshift/dev_setup.sh
@@ -180,6 +180,7 @@ main() {
     "$PROJECT_DIR/developer/openshift/apps/$app.sh"
     echo
   done
+  check_CRDs
 }
 
 if [ "${BASH_SOURCE[0]}" == "$0" ]; then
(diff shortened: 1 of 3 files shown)


xinredhat:
this PR is ready for review. I have already tested that if the pipeline-service tests  fail, the `test-pipeline-service` task will also fail.

developer/openshift/dev_setup.sh
```diff
@@ -168,6 +168,23 @@ cluster_setup() {
   | indent 2
 }
 
+# Sometimes CRDs need a few seconds to be ready
```
Roming22, line 171: I feel this should be moved to `developer/openshift/utils.sh`

Roming22 (changes requested):

developer/openshift/dev_setup.sh
xinredhat: fixed

PR #823, opened 2023-10-13
Fix kube-linter issues

hugares:
* run containers as non root
* run containers with read only root file system

The 'mc' init container is using the /tmp folder so for that container, mount an empty dir volume to /tmp to be able to set root file system readonly while being able to write to /tmp.

PLNSRVCE-1476


hugares:
@sayan-biswas @Roming22

sayan-biswas (approved):

PR #825, opened 2023-10-16
Fix a minor error of devenv image

xinredhat:
It reported error `Error: Unable to find a match: python3-devel-3.11.5` when building devenv image. It needs to install `python3-devel-3.11.6` as replacement.


xinredhat:
Ready for review

Roming22 (approved):

PR #827, opened 2023-10-16
Set cpu/memory requests/limits for chains-secret-generation container

hugares:
The job takes around 150Mi so give it a bit a breathing room by setting the limit to 250Mi.

PLNSRVCE-1476


hugares:
@Roming22 Can you tell me if my change is causing the failure of ocp-412 and ocp-413 checks? I do not have access to the check detail/logs

Roming22:
@hugares Sorry, Idid not see your message. The issue with the CI was because of an outage of HCP.

Roming22 (approved):

PR #831, opened 2023-10-18
Fix TektonConfig

Roming22:
Restore a setting that was mistakenly removed with commit b2626be4fc649b5184b8bf3e661bbc46319c884b
which break the deployment of the application

### DIFF
--- a/operator/gitops/argocd/pipeline-service/openshift-pipelines/tekton-config.yaml
+++ b/operator/gitops/argocd/pipeline-service/openshift-pipelines/tekton-config.yaml
@@ -10,6 +10,8 @@ metadata:
     argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
     argocd.argoproj.io/sync-wave: "0"
 spec:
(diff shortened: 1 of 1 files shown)


gabemontero:
Interestingly, there is no godoc for that profile field @Roming22 @enarha :  https://github.com/tektoncd/operator/blob/main/pkg/apis/operator/v1alpha1/tektonconfig_types.go#L77

the default does appear to be basic - https://github.com/tektoncd/operator/blob/main/pkg/apis/operator/v1alpha1/tektonconfig_defaults.go#L28

searching for refs in the operator code, I see hits to v1alpha1 addons and the tekton dashboard

my "guess" is this could be intentional, as we need neither of those, but I agree @Roming22 should officially declare that as the original commit did not touch upon that as well

enarha:
So we are actually changing `all` to `basic` by removing that field. The documentation touches on that https://tekton.dev/docs/operator/tektonconfig/#profile

Roming22:
@enarha Thanks for catching that! Fixed.

PR #835, opened 2023-10-23
fix missing millisecond to second conversion in grafana dashboard execute overhead query

gabemontero:
Noticed this while I was working on the gitlab app-sre SLO/SLI docs

PR #838, opened 2023-10-25
optimize alert query; revert incorrect millisecond to second conversion fix

gabemontero:
During unsuccessful attempts to get our alert queries running on the app-sre grafana dashboard, I did come up with a simpler query which reaches the same basic effect of only using the last 30 minutes of data available.  This change will also work better with the use of range seen in other successful app-sre dashboard queries.
Also, I realized an earlier change by me  to divide our milliscecond metric's _count by 1000 was incorrect.  Only the _sum has actual millisecond based observations that need to be converted (as was originally done).

The PR where I incorrectly changed what @ramessesii2 had for converting milliseconds to seconds was https://github.com/openshift-pipelines/pipeline-service/pull/835

Here is a screen shot of how the new queries look in RHTAP red-hat prod:

optimized-queries

### DIFF
--- a/operator/gitops/argocd/grafana/dashboards/pipeline-service-dashboard.json
+++ b/operator/gitops/argocd/grafana/dashboards/pipeline-service-dashboard.json
@@ -146,7 +146,7 @@
         {
           "editorMode": "code",
           "exemplar": true,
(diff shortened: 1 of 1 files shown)


Roming22 (approved):

ramessesii2 (approved):

ramessesii2:
```
task deploy-pipeline-service-baseline has the status "Failed":

Normal  ResourceUpdated   9m46s  argocd-application-controller  Updated health status:  -> Missing
      Normal  ResourceUpdated   85s    argocd-application-controller  Updated health status: Missing -> Degraded
```
/retest test-pipeline-service-upgrade-ocp-412

PR #840, opened 2023-10-26
Switch to new resources for Rosa HCP cluster

xinredhat:


xinredhat:
the errors happened at task `deploy-cluster` are expected, because they are using the new resources, for example, oidc-config id, operator-roles, etc.

Roming22:
ROSA token updated on vault tests are expected to pass.

Roming22 (approved):

PR #842, opened 2023-10-26
Update checkov configuration

Roming22:
The --skip-fixes flag has been deprecated in v3.x.


Roming22:
@ramessesii2 Since you created the configuration, please validate that there is no impact to removing the flag,

ramessesii2 (approved):

ramessesii2:
I don't think removing the flag will affect our current configuration since `--skip-fixes` only [had effect](https://github.com/bridgecrewio/checkov/commit/3c970fea6452c1971832f4c1cfbb306f8d99308d#diff-bc34d70938c53319b4c3b13f584d4b0450f9909e61ff8b5e007bb557abb564a7L47) when using the Bridgecrew API key.

PR #845, opened 2023-10-30
Add option to delete images after building them

Roming22:
As the images are not used on the cluster, deleting them helps with resource management.


xinredhat (approved):
/lgtm

PR #851, opened 2023-11-02
Fix kube-linter issues for pac-secret-reaper CronJob

hugares:
* Set CPU/memory requests/limits
* Set root file system read only

The job can use up to around 50m/50Mi so set the limits higher to give it enough breathing room.

PLNSRVCE-1476


Roming22 (approved):

Roming22:
We are having CI issues unrelated to your change. Once they are solved, and the CI is green, we can merge the PR.

PR #857, opened 2023-11-08
Fix images definitions

Roming22:


Roming22:
Merging to trigger the creation of the new images and start solving the CI issues.

PR #862, opened 2023-11-14
WIP:upgrade rosa cli to 1.2.30 to fix billing account issue

xinredhat:
There is a hot fix that ROSA CLI should not require the billing account id for HCP in both 1.2.29 and 1.2.30, we can take 1.2.30 for fixing the issue of billing account


Roming22 (approved):

Roming22:
I merged it because it would have happened with #839 anyway.

PR #868, opened 2023-11-27
[DNM] enable immediate pruning of PipelineRuns in Tekton Results

ramessesii2:


Roming22:
As discussed, we'll wait for https://github.com/tektoncd/results/pull/672 to be merged and deployed, and deactivate the default pruner before merging this PR.

operator/gitops/argocd/pipeline-service/tekton-results/watcher-config.yaml
```diff
@@ -19,7 +19,7 @@ spec:
               "-auth_mode",
               "token",
               "-completed_run_grace_period",
-              "10m",
+              "-1", # anything less than 0 means completed PipelineRuns will be pruned immediately.
```
sayan-biswas, line 22: This might have undesirable consequences because the streaming of logs happens in a separate go routine which doesn't negotiate with the controller workflow and hence the deletion.

https://github.com/tektoncd/results/blob/097030b9f52bf01149f7555d86b26f7a77327c60/pkg/watcher/reconciler/dynamic/dynamic.go#L323-L339

There is an open bug on this.
https://github.com/tektoncd/results/issues/514

But it can be tested in staging. My guess there'll be a race condition.

ramessesii2:
Closed in favor of #905

PR #880, opened 2023-12-08
updated the name of reserved sprayproxy cluster

xinredhat:


xinredhat:
since this PR doesn't break the CI, which just update the excluded the name of sprayproxy cluster, I will merger it directly

PR #884, opened 2023-12-13
 fix the issuse of getting the hcp full version

xinredhat:


Roming22 (approved):
LGTM

@xinredhat I see the tests are OK, but the PR has the WIP flag. If the flag should not be there you can remove it and merge the PR.

PR #889, opened 2023-12-14
Disable 'check_owner' for Tekton Result watcher

Roming22:
This will enable the clean up of all PipelineRuns, including those having ownerReferences set to RHTAP resources.

Once the flag has been confirmed to work successfully, the default Pipelines pruner can be removed.

rh-pre-commit.version: 2.0.3


ramessesii2 (approved):
/lgtm

PR #891, opened 2023-12-20
Disable Results API configuration DB SSL

enarha:
This does not change the existing state!
Till now, the Results API had DB_SSLMODE set to required, but that was not supported by the code. There was no way to provide certificate, so the connection to the DB was never encrypted. A recent change in Tekton Results added that functionality and it makes the API pod fail to start when certificate is not provided. We intend to enable that, but first we need to fix the issue so the pod actually starts. Adding ceriticate and enabling the TLS will be added as separate change.


Roming22 (approved):

xinredhat (approved):
/lgtm

PR #895, opened 2024-01-04
Disabling check_owner because of a major bug

Roming22:
A bug in Tekton Results needs to be fixed before we can activate the feature.

rh-pre-commit.version: 2.0.3


gabemontero (approved):
/lgtm

PR #906, opened 2024-01-18
add config for enabling tekton results dbssl

avinal:
- add dbssl for dev_setup
- add config for enabling dbssl genrerally


developer/openshift/apps/pipeline-service.sh
```diff
@@ -56,6 +57,23 @@ EOF
   TEKTON_RESULTS_S3_PASSWORD="$(yq ".s3_password" "$tekton_results_credentials")"
 }
 
+generate_tekton_results_db_ssl_cert(){
+  TEKTON_RESULTS_DB_SSL="$WORK_DIR/certificates/tekton-results/ca.pem"
```
gabemontero, line 61: it has been forever, but I see  that `TEKTON_RESULTS_DB_SSL` is set/defined in this method, but also used in `patch_tekton_results_manifests`

does this work at all if you don't export `TEKTON_RESULTS_DB_SSL` ?

or does it happen to work if and only if `generate_tekton_results_db_ssl_cert` is called before `patch_tekton_results_manifests` ?

I *think* it is better to perhaps define `TEKTON_RESULTS_DB_SSL` via an `export TEKTON_RESULTS_DB_SSL` either in `main()` or in the setup function.

Or are my bash best practices recollections way off?  WDYT @Roming22 ?

avinal: I was confused about this as well. But it seems variables are by default global in scope in shell script. Using `export` will put it outside the script scope too, which is not really needed in our case. 

we have been using this pattern already above in the code: https://github.com/openshift-pipelines/pipeline-service/blob/main/developer/openshift/apps/pipeline-service.sh#L53-L56

Here is a reference: https://tldp.org/LDP/abs/html/othertypesv.html
https://tldp.org/LDP/abs/html/localvar.html

gabemontero: interesting ... thanks for the follow up and refs @avinal 

all that said, `shellcheck` and `yamlint` are both complaining in the static code analysis, so some follow up will be needed regardless

otherwise, I'll let @Roming22 chime in on the bash elements once you clean up those static checks, and we'll go from there

avinal:
I have fixed the static checks and will push once other tests are green.

gabemontero:
static check still failing @avinal 

the upgrade test was an hcp/rosa flake

avinal:
Now it passes :)

PR #909, opened 2024-01-24
Add CatalogSource including index with OSP nightly build

enarha:
Here we only add the CatalogSource, but does not make use of it (does not reference it in the kustomization.yaml). An automation job will be including that reference until we switch to nightly builds.


enarha:
I'm not sure about the CI failing. My change does not introduce any actual changes to take effect on the cluster. Comparing `kustomize build operator/gitops/argocd/pipeline-service/` between main and my feature branch does not produce any diff.

PR #910, opened 2024-01-24
Create a PR to update OSP nightly deployment

enarha:
Currently this includes:
    1. Check for nightly build and update the CatalogSource
    2. Update the kustomization.yaml to add the CatalogSource
    3. Update the Subscription to use the new CatalogSource and channel

Once we start using the nightly builds, we only need to update the CatalogSource to use the new index image (drop 2 and 3).


gabemontero:
I was about to post something about the checkov failure @enarha but before I could hit send I noticed you pushed a change and I now see the static-code-analysis is passing :-)

Roming22 (approved):

PR #918, opened 2024-01-30
Create namespaces before external secrets

enarha:
ExternalSecrets are creates at sync-wave "-1" and they are required for services like Tekton Results and OSP. But for the secrets to be created, the namespace should exist in advance. Here we create the namespace before the secret. It does not affect the deployment of the OSP operator, it ignores the fact the namespace it tries to create already exist.

### DIFF
--- a/operator/gitops/argocd/pipeline-service/openshift-pipelines/namespace.yaml
+++ b/operator/gitops/argocd/pipeline-service/openshift-pipelines/namespace.yaml
@@ -1,10 +1,10 @@
+# Solves https://issues.redhat.com/browse/PLNSRVCE-1620
 ---
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: tekton-results
+  name: openshift-pipelines
   labels:
     argocd.argoproj.io/managed-by: openshift-gitops
   annotations:
-    # Solves https://issues.redhat.com/browse/PLNSRVCE-1620
     argocd.argoproj.io/sync-wave: "-1"
--- a/operator/gitops/argocd/pipeline-service/openshift-pipelines/kustomization.yaml
+++ b/operator/gitops/argocd/pipeline-service/openshift-pipelines/kustomization.yaml
@@ -14,3 +14,4 @@ resources:
   - chains-observability-service.yaml
   - chains-public-key-viewer.yaml
   - chains-secrets-config.yaml
+  - namespace.yaml


enarha:
It looks like sync-wave "-1" should be enough (no need for "-2"), because "By kind (e.g. namespaces first and then other Kubernetes resources, followed by custom resources)".

gabemontero (approved):
/lgtm

I at least @enarha am not able to tail the test logs live from the `tekton-ci` ns on the prd cluster from the ocp console.

Are you able to see the logs?

Or were you able to use dev-setup.sh or manual oc apply -k or kustomize invocations and confirm the order ?

thanks

enarha:
@gabemontero I'm able to see the logs from the CI, but I think they are not very helpful in this case because we do not use those external secrets in any way in pipeline-service.
```
[ pipeline-service] -> kustomize build operator/gitops/argocd/pipeline-service/ | oc apply -f -
Warning: resource namespaces/openshift-pipelines is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by oc apply. oc apply should only be used on resources created declaratively by either oc create --save-config or oc apply. The missing annotation will be patched automatically.
namespace/openshift-pipelines configured
namespace/tekton-results created
serviceaccount/chains-secrets-admin created
```
shows that the namespaces are created indeed first. Specifically the openshift-pipelines should not have been there without this change.

gabemontero:
> @gabemontero I'm able to see the logs from the CI, but I think they are not very helpful in this case because we do not use those external secrets in any way in pipeline-service.
> 
> ```
> [ pipeline-service] -> kustomize build operator/gitops/argocd/pipeline-service/ | oc apply -f -
> Warning: resource namespaces/openshift-pipelines is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by oc apply. oc apply should only be used on resources created declaratively by either oc create --save-config or oc apply. The missing annotation will be patched automatically.
> namespace/openshift-pipelines configured
> namespace/tekton-results created
> serviceaccount/chains-secrets-admin created
> ```
> 
> shows that the namespaces are created indeed first. Specifically the openshift-pipelines should not have been there without this change.

yeah the creation order was all I was worried about it it was not clear earlier @enarha ... thanks for confirming 

I'll defer you if you want to get approval from @Roming22 before merging

operator/gitops/argocd/pipeline-service/openshift-pipelines/namespace.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: v1
```
Roming22, line 2: Can you add a comment to record why that needs to happen?
I'm worried we might ask ourselves in a few months and not remember why.

enarha: Added a comment pointing to the Jira issue as I found we did in other cases.

operator/gitops/argocd/pipeline-service/tekton-results/namespace.yaml
```diff
@@ -6,4 +6,4 @@ metadata:
   labels:
     argocd.argoproj.io/managed-by: openshift-gitops
   annotations:
-    argocd.argoproj.io/sync-wave: "0"
+    argocd.argoproj.io/sync-wave: "-1"
```
gabemontero, line 9: should we add a comment here too since a -1 sync-wave is "atypical" IIUC @enarha @Roming22 ?

enarha: Done.

gabemontero:
assuming the upgrade failure is a flake, but let's try again 

/retest

gabemontero:
clean / merging

PR #919, opened 2024-01-31
PLNSRVCE-1585: exporter needs patch permissions

gabemontero:
Part of the exporter's new approach to filter throttled pipelineruns because of the pipeline controller's aggressiver requeue strategy with throttling is that it labels the pipelinerun for later filtering.  However, the original introduction of the changes did not add patch permissions to the exporter's RBAC.  This change addresses that.

rh-pre-commit.version: 2.1.0

### DIFF
--- a/operator/gitops/argocd/pipeline-service/metrics-exporter/clusterrole.yaml
+++ b/operator/gitops/argocd/pipeline-service/metrics-exporter/clusterrole.yaml
@@ -19,7 +19,7 @@ rules:
 
   - apiGroups: ["tekton.dev"]
     resources: ["pipelineruns", "taskruns"]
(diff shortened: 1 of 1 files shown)


Roming22 (approved):

Roming22:
/lgtm

@xinredhat  Can you check if there's a reason why the upgrade test is flaking?

gabemontero:
> /lgtm
> 
> @xinredhat Can you check if there's a reason why the upgrade test is flaking?

the upgrade test flaked in https://github.com/openshift-pipelines/pipeline-service/pull/920 as well

gabemontero:
I'm merging this @Roming22 since an RBAC addition will not affect an upgrade

PR #930, opened 2024-02-07
Add cronjob trigger to OSP nightly workflow

enarha:
The intention always was to use a scheduler to execute the workflow daily. The manual trigger is kept for both easier testing and emergency updates.

### DIFF
--- a/.github/workflows/update-osp-nightly.yaml
+++ b/.github/workflows/update-osp-nightly.yaml
@@ -1,7 +1,10 @@
 ---
 name: update-osp-nightly
 run-name: Check for OSP nightly and create a PR
-on: workflow_dispatch
+on:
+  schedule:
+    - cron: '10 1 * * *'
+  workflow_dispatch:
 permissions: {}  # drop all permissions; the default triggers codecov failure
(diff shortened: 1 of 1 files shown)


gabemontero (approved):
/lgtm

am I correct in thinking that the only proper verification of this @enarha is to look for the github action firing at the appointed interval?

enarha:
> am I correct in thinking that the only proper verification of this @enarha is to look for the github action firing at the appointed interval?

Yes. If that helps, we have similar setup here https://github.com/openshift-pipelines/pipeline-service/blob/main/.github/workflows/periodic-scanner-quay.yaml#L3-L6 and I tested it myself.

gabemontero:
> > am I correct in thinking that the only proper verification of this @enarha is to look for the github action firing at the appointed interval?
> 
> Yes. If that helps, we have similar setup here https://github.com/openshift-pipelines/pipeline-service/blob/main/.github/workflows/periodic-scanner-quay.yaml#L3-L6 and I tested it myself.

cool ... I'll defer to you or @Roming22 on when to hit the merge button

PR #935, opened 2024-02-08
Migrate Tekton Chains to an ExternalSecret

Roming22:
In order to facilitate backup/restore of a cluster, the signing secret is now stored externally and deployed on the cluster using an ExternalSecret.

The original behavior is maintained for local development.

rh-pre-commit.version: 2.1.0

### DIFF
--- a/developer/openshift/apps/pipeline-service.sh
+++ b/developer/openshift/apps/pipeline-service.sh
@@ -17,10 +17,10 @@ setup_work_dir() {
   setup_tekton_results | indent
 }
 
-configure_argocd_apps(){
+configure_argocd_apps() {
   echo -n "- Updating source repository to '${GIT_URL}/tree/$GIT_REF': "
   # Patch the url/branch to target the expected repository/branch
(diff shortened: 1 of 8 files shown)


gabemontero:
is this PR still needed with your infra-deployments PR @Roming22 ?

Roming22:
@gabemontero Yes. The goal of this PR is to disable the automated creation of the secret.
The reason is that if the secret is managed by Vault, we do not want a secret to be created if there is no Vault secret, or if the secret is deleted by mistake and the Job runs before the secret is synced.

gabemontero (approved):
I think this lgtm @Roming22 

as to the details of the change, IIUC you have moved the old job that dynamically creates the chains secret to the developer-only flow out of this repo, and thus, on the konflux i.e. infra-deployments side, the job that dynamically creates the secret no longer gets started, correct?

assuming so, feel free to merge at your discretion; otherwise, please clarify

thanks

Roming22:
@gabemontero Correct. The procedure to deploy a new "long-live" cluster is to create the secret in Vault. No default secret will be created, except for "local" installs.

PR #936, opened 2024-02-08
Misc fix

Roming22:
* Do not run test.sh in debug mode by default.
* Run test.sh in debug mode in the CI
* Auto-formatting of modified shell scripts (you can turn off whitespace diff in the files view to make the review easier)
* Fix an issue that would make any change to the tekton-chains-signing-secret require a manual action for the change to be deployed. It is possible that to deploy this fix, the manual action of deleting the existing job might be required one last time.

### DIFF
--- a/operator/gitops/argocd/pipeline-service/openshift-pipelines/chains-secrets-config.yaml
+++ b/operator/gitops/argocd/pipeline-service/openshift-pipelines/chains-secrets-config.yaml
@@ -50,6 +50,7 @@ metadata:
   namespace: openshift-pipelines
   annotations:
     argocd.argoproj.io/sync-wave: "1"
+    argocd.argoproj.io/hook-delete-policy: BeforeHookCreation  # Delete the job so it can be recreated and updated during the sync.
 spec:
   template:
     spec:
--- a/operator/test/test.sh
+++ b/operator/test/test.sh
@@ -4,7 +4,6 @@
 set -o errexit
 set -o nounset
 set -o pipefail
-set -x
 
 usage() {
   echo "
@@ -102,7 +101,7 @@ check_pod_security() {
       echo "   - Skip: $pod is exempt"
       continue
     fi
-    scc="$(kubectl get "$pod" -o jsonpath='{.metadata.annotations.openshift\.io/scc}' -n "$1"  --ignore-not-found)"
+    scc="$(kubectl get "$pod" -o jsonpath='{.metadata.annotations.openshift\.io/scc}' -n "$1" --ignore-not-found)"
     if [ -z "$scc" ]; then
       echo "   - Skip: $pod could not be inspected (most like the pod terminated quickly)"
       continue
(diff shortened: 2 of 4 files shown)


operator/gitops/argocd/pipeline-service/openshift-pipelines/chains-secrets-config.yaml
```diff
@@ -50,6 +50,7 @@ metadata:
   namespace: openshift-pipelines
   annotations:
     argocd.argoproj.io/sync-wave: "1"
+    argocd.argoproj.io/hook-delete-policy: BeforeHookCreation
```
gabemontero, line 53: a comment explaining this argo feature so someone later on does not have to look it up would be nice

ci/images/ci-runner/hack/bin/run-plnsvc-test.sh
```diff
@@ -6,7 +6,7 @@ set -x
 
 echo "Run pipeline-service tests..."
 if kubectl -n default exec pod/ci-runner -- \
-    sh -c "/workspace/sidecar/bin/plnsvc_test.sh"; then
+    sh -c "/workspace/sidecar/bin/plnsvc_test.sh --debug"; then
```
gabemontero, line 9: since you did not touch plnsvc_test.sh, I'm assuming that it was already looking for `--debug` and employ `set -x` of some equivalent ?

gabemontero (commented):
a question and very minor suggestion (which I won't argue strongly for if you don't like)

ci/images/ci-runner/hack/bin/run-plnsvc-test.sh
Roming22: That's an error, thanks for catching it. I was supposed to update `ci/images/ci-runner/hack/sidecar/bin/plnsvc_test.sh`

operator/gitops/argocd/pipeline-service/openshift-pipelines/chains-secrets-config.yaml
Roming22: done

operator/test/test.sh
```diff
@@ -4,7 +4,6 @@
 set -o errexit
 set -o nounset
 set -o pipefail
-set -x
```
gabemontero, line 7: OK so I see where you turned off debug by default, but I don't see where the `--debug` arg is getting processed ... is that a default bash thing I'm missing ?

gabemontero (commented):
argocd comment helpful

See below - I'm still a little confused about how debug is turned on now, but maybe I am missing something

operator/test/test.sh
Roming22: This done in the parse_args function, around line 48/49.

gabemontero: ah cool leveraging an existing arg whose impl was too far from the deltas of this PR for me to see ... thanks for the ptr

gabemontero (approved):
/lgtm

PR #944, opened 2024-02-13
Remove SCRIPT_IMAGE param from update infra-deployments task

enarha:
The task we reference in the PipelineRun
quay.io/redhat-appstudio-tekton-catalog/task-slack-webhook-notification:0.1 is changed and the SCRIPT_IMAGE parameter removed. Instead a default image is used.
ref: https://github.com/redhat-appstudio/build-definitions/pull/792

The new image does not have `kustomize`, but it has `oc`, so we use `oc kustomize` instead.


gabemontero (approved):
/lgtm

Roming22 (approved):

PR #945, opened 2024-02-13
Fix kustomize command

enarha:
`oc kustomize` does not have `build` command.


Roming22 (approved):
CI won't test the change, so feel free to merge.

PR #946, opened 2024-02-13
gather more diagnostic data on failures during results test

gabemontero:
to deal with the recent permfail in the results during test-upgrade (and only this permutation of the test)

@openshift-pipelines/pipelines-service FYI / PTAL

rh-pre-commit.version: 2.2.0


gabemontero:
sure enough the upgrade test passed with my debug

Roming22 (approved):

gabemontero:
got the error but we need this debug to merge before it is used 

merging

PR #947, opened 2024-02-15
Remove the developer/devenv image from the CI

Roming22:
The image is brittle because the image tag is floating. Since the image is just used for development, it should not be an issue to remove it from the CI.

rh-pre-commit.version: 2.1.0


gabemontero (approved):
lgtm if you say so @Roming22 :-) 

never fully followed all the dev/ci image machinations here

Did you look to see if the docker image build failure is related to this change or not?

PR #948, opened 2024-02-18
Fix git RPM version no longer available

enarha:
The git-2.43.0 version is no longer available, making the Dockerfile build (part of test-docker-images-build job) fail.


avinal (approved):
/lgtm

PR #970, opened 2024-03-17
Update the update-osp-nightly GH action

enarha:
The peter-evans/create-pull-request GH action has newer version fixing the "Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20" error.


enarha:
The change does not have effect on the pipeline-service functionality and the upgrade test failure is not related.

enarha:
/hold
holding merge for now to avoid noise to the infra-deployments repo.

PR #980, opened 2024-03-22
Use newer Tekton Results version

enarha:
Use newer Tekton Results version including the "Support all data types in summary annotations" (fix for the json unmarshal bug). Drop the "employ some channel best practices to avoid hanging goroutines and memory leaks from unclosed buffered channels" commit.

The commit is from the downstream-0.9.1-03-no-21d96ba branch on the Results fork.


avinal (approved):
Looks good to me

enarha:
I was able to verify deploy, upgrade and downgrade including running the tests after each step. The downgrade test failure is a problem with our testing which we have to check and fix, but it's not related to the change.

PR #981, opened 2024-03-24
Add konflux-ci namespace to cleaning secrets script

enarha:
https://github.com/redhat-appstudio/infra-deployments/pull/3481 introduced change to the pipeline-service deploy.yaml files we generate automatically overriding those changes. Here we move the change to the pipeline-service repo.

### DIFF
--- a/operator/gitops/argocd/pipeline-service/openshift-pipelines/bugfix-pac-gitauth-secrets.yaml
+++ b/operator/gitops/argocd/pipeline-service/openshift-pipelines/bugfix-pac-gitauth-secrets.yaml
@@ -105,7 +105,7 @@ spec:
                   set -o errexit
                   set -o nounset
                   set -o pipefail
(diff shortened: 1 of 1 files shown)


enarha:
@ralphbean

gabemontero:
makes sense @enarha, though since @ralphbean changed our cronjob on us in prod arleady, do you have any idea if this means that *UNTIL* 

`disable their repository entry in the tekton-ci namespace and enable their repository in the konflux-ci namespace by editing the repository entries for each here in infra-deployments.` 

whatever this cron job is suppose to be doing for our pac secret is effectively broken?

Or do we need to reach out to Romain to answer that question?

enarha:
@gabemontero I really do not know to answer your question, but without this change each time we generate the deploy.yaml files (for both stage and prod) we override this values. Furthermore the changes we generate include that override which is confusing when you review the infra-deployments changes. I would rather merge that PR soon. It does not have any effect on infra-deployments other than preserving the configuration already there.

enarha:
After offline sync we decided to go ahead and merge.

PR #984, opened 2024-04-01
Cherry-pick add config for enabling tekton results dbssl

avinal:
- add dbssl for dev_setup
- add config for enabling dbssl generally

Signed-off-by: Avinal Kumar <avinal@redhat.com>

rh-pre-commit.version: 2.2.0
rh-pre-commit.check-secrets: ENABLED

PR #986, opened 2024-04-03
Revert "add config for enabling tekton results dbssl"

gabemontero:
This reverts commit 5e38cd44d8455e77def659ea116a684c5cfa4320.

From https://github.com/openshift-pipelines/pipeline-service/pull/984

@avinal @enarha FYI

Unfortunately we have had to revert this again.  We have new changes we need to start staging in.

The tl;dr is located at both https://github.com/redhat-appstudio/infra-deployments/pull/3109 and https://redhat-internal.slack.com/archives/C032EJ007C0/p1711979764900189

Exec level summary: per email exchange with  @scoheb the public AWS cert should be added to the system root CA of the result api server image and @avinal 's latest attempt while still not working, would not work for stage/prod anyway.

We have to remember to test the konflux e2e via a test PR in infra deployments that leverages whatever branch the next pipeline-server PR points to for bringing this back in.  Minimally that will be after @avinal returns from PTO


enarha (commented):
/lgtm

PR #989, opened 2024-04-15
add grafana views for tekton chains controller workque and latency

jkhelil:
a fix for https://issues.redhat.com/browse/SRVKP-4405

### DIFF
--- a/operator/gitops/argocd/pipeline-service/tekton-results/kustomization.yaml
+++ b/operator/gitops/argocd/pipeline-service/tekton-results/kustomization.yaml
@@ -3,7 +3,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 namespace: tekton-results
 resources:
-  - https://github.com/openshift-pipelines/tektoncd-results.git/config/overlays/base-only/?ref=bae7851ff584423503af324200f52cd28ca99116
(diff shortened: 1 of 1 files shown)


gabemontero:
@jkhelil when you get the chance can you launch your own OCP cluster, run dev_setup.sh -- use-local-branch , and capture a screen shot of the new panels?

I forgot to mention when we had our video conf that such screen shots were part of our methodology when updating these panels.

I may have one review comment I'll post separately.

thanks

jkhelil:
@gabemontero Here is a screenshot
image

gabemontero:
Certainly better than before @jkhelil but it looks like something got lost when I tried to describe it to you before.  I was hoping it would look like how the `Watcher Work Queue Depth` and `Watcher Reconcile Latency` works with Tekton Results.

See this screen shot:

ResultsDepthLantency


Can you copy those panel over to the chains section, and then modify the queries to grab those numbers for chains instead of results watcher?

thanks

jkhelil:
@gabemontero I am using the same queries as for watcher, i didnt touch th existant chains panels (work queue rate and cpu) the two first panels on my screenshot.
The panels i have added, are the two last ones. On my cluster, i tried to run chains, example to have some data, mm, but i think i need more example to see some data there.
but yeah just want to highlight that i am using the same queries in the two added panels

jkhelil:
@gabemontero ok got it. 
Fiixing it to look like this 
image

PR #994, opened 2024-05-07
Fix missing git rpm version

enarha:
Update to newer ubi-minimal base image version and the respective rpm versions.


enarha:
Failed on timeout. I checked before the timeout and it was all running and successful. This is a timing issue, but I'll just retry before extending the timeout for the pipeline.
/test test-docker-images-build

avinal (approved):
Version change only. LGTM

gabemontero:
didn't get to it in time, but lgtm

PR #1010, opened 2024-06-21
pin watcher to mem leak version, revert latest fix attempt

gabemontero:
undo only half of https://github.com/openshift-pipelines/pipeline-service/pull/1007 as the older watcher and newest api server work OK together

let's keep the api server tuning options

@enarha PTAL

rh-pre-commit.version: 2.3.0
rh-pre-commit.check-secrets: ENABLED

PR #1013, opened 2024-06-24
copy pipeline-svc install of yq to open-infra-deployment job

gabemontero:
Discovered with the post merge job in https://github.com/openshift-pipelines/pipeline-service/pull/1010 that `yq` is not available in the base image provided by build-definitions

if this does not work, per discussion with @enarha I'll try `awk`, then either pin both images, or maybe @enarha tries his python idea

rh-pre-commit.version: 2.3.0


gabemontero:
@enarha came up with a sed based solution

PR #1014, opened 2024-06-24
switch back off yq for sed for pinning watcher image, but get it working

gabemontero:
@enarha and I were able to test this locally

rh-pre-commit.version: 2.3.0


gabemontero:
we hit the intermittent validation webhook timeout on our open infra pr job that other users hit earlier today in konflux-users (I tagged you earlier today over in https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1719233399107819?thread_ts=1719230416.862599&cid=C04PZ7H0VA8 ):

```
 Pipeline tekton-ci/open-infra-deployment-pr-hn6mk can't be Run; it contains Tasks that don't exist: Couldn't retrieve Task "": error requesting remote resource: resolution took longer than global timeout of 1m0s 
```

but given our testing I think we'll be good next time @enarha .... just keep an eye on it if anything merges here while I am on PTO.

PR #1020, opened 2024-07-10
SRVKP-4523: add alert related deadlock / crashloop panels for core tekton controllers

gabemontero:
Also fixing issues with older panels as various var substitutions only work in App SRE, some metrics names were incorrect, and some 'prometheus' filters were no longer (if ever) valid


gabemontero:
cluster deploy failed

/retest

gabemontero:
OK Rosa working again

local testing good ... merging

PR #1024, opened 2024-07-12
[DO-NOT-MERGE-YET] turn on pprof in core controller to allow for goroutine dumps

gabemontero:
per the golang recommendations that pprof enablement is OK in production, we are enabling to allow for gathering goroutine dumps as part of diagnosing deadlocks.

rh-pre-commit.version: 2.3.0

### DIFF
--- a/operator/gitops/argocd/pipeline-service/openshift-pipelines/tekton-config.yaml
+++ b/operator/gitops/argocd/pipeline-service/openshift-pipelines/tekton-config.yaml
@@ -87,6 +87,14 @@ spec:
     options:
       disabled: false
       configMaps:
+        config-observability:
+          data:
+            metrics.count.enable-reason: 'false'
+            metrics.pipelinerun.duration-type: histogram
+            metrics.pipelinerun.level: pipeline
+            metrics.taskrun.duration-type: histogram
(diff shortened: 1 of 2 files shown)


operator/gitops/argocd/pipeline-service/openshift-pipelines/tekton-config.yaml
```diff
@@ -87,6 +87,14 @@ spec:
     options:
       disabled: false
       configMaps:
+        config-observability:
+          data:
+            metrics.count.enable-reason: 'false'
```
gabemontero, line 92: all the metrics props are defaults I have observed once we are deployed

I will run comparisons to confirm/deny whether tekton config application is an add/merge or a replace type of operation

enarha:
lgtm. Feel free to merge.

gabemontero:
> lgtm. Feel free to merge.

thanks @enarha 

unfortunately with local testing I cannot get goroutine dumps from the pipeline controller like I can with the results watcher, even though AFAIU I'm configuring the same way via the config-observability config map.

been debugging a bit, will a bit more, but may reach out on #team-ocp-pipelines for some help.

PR #1037, opened 2024-08-12
Add Deadlocked ResolutionRequests panel and update Pipelines controller restart panel

divyansh42:
Add Deadlocked ResolutionRequests panel same as App SRE dashboard and make changes in Pipelines controller restart panel as per App SRE dashboard

### DIFF
--- a/operator/gitops/argocd/grafana/dashboards/pipeline-service-dashboard.json
+++ b/operator/gitops/argocd/grafana/dashboards/pipeline-service-dashboard.json
@@ -227,7 +227,7 @@
       "type": "stat"
     },
     {
-      "description": "The number of times any of the pipelines controllers hav restarted",
+      "description": "The number of times any of the tekton controllers have restarted",
       "fieldConfig": {
         "defaults": {
           "color": {
@@ -281,7 +281,7 @@
       "targets": [
         {
           "editorMode": "code",
-          "expr": "sum(increase(kube_pod_container_status_restarts_total{namespace=\"openshift-pipelines\", pod=~\"tekton-.*\"}[2m]))",
+          "expr": "sum(increase(kube_pod_container_status_restarts_total{namespace=\"openshift-pipelines\", pod=~\"tekton-.*\"}[$__range]))",
           "legendFormat": "__auto",
           "range": true,
           "refId": "A"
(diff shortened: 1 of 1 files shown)


divyansh42:
image
@gabemontero This is how the present panel looks like

gabemontero:
looks good @divyansh42 

and I think none of the rows in the json after your insertion changed because everything fit on one row

So at the moment I believe this would be "mergeable"

I'll check out your PR locally and confirm it looks the same for me as well once I have a cluster up.

divyansh42:
> and I think none of the rows in the json after your insertion changed because everything fit on one row

@gabemontero 
Yes, that was the reason

gabemontero:
I see the same panel you showed below @divyansh42  and the panels below look OK when I expand the sections.

Per our chat in scrum let's move onto the resolution request controller restart panel.  I'll add some comments to your education Jira to reiterate what we discussed.

thanks

divyansh42:
@gabemontero this is the modified dashboard. Please take a look.
I have added the panel in the second row. Please let me know if that doesn't look good.
image

operator/gitops/argocd/grafana/dashboards/pipeline-service-dashboard.json
```diff
@@ -281,7 +281,7 @@
       "targets": [
         {
           "editorMode": "code",
-          "expr": "sum(increase(kube_pod_container_status_restarts_total{namespace=\"openshift-pipelines\", pod=~\"tekton-.*\"}[2m]))",
+          "expr": "sum(increase(kube_pod_container_status_restarts_total{namespace=\"openshift-pipelines\", pod=~\"tekton-.*\"}[$__range]))",
```
gabemontero, line 284: I didn't think `$__range` worked in our non app SRE envs, but I have you change up now, and it is working

Assuming you copied the query as is from the app SRE panel, this is a happy find :-)

gabemontero (approved):
looks good ... just one minor observation, no need to change anything

squash the commits, change the PR title to reflect what we are doing now, and then I'll merge and we'll move onto copying this change to infra-deployments on Wednesday during stand up / scrum.

also, on your question around you panels showing n/a's , aside from creating pipelinerun, taskruns, or resolutionrequests, I suspect at least some of what you have going on is you have to go to the top right hand corner and change the interval from last hour or what ever the default is to last 5 minutes or whatever the shortest interval is

operator/gitops/argocd/grafana/dashboards/pipeline-service-dashboard.json
divyansh42: Yes, I copied as it is.

gabemontero:
> I see one of the tests is failing. However, I am not able to get to the logs. It seems I don't have access.

yeah getting logs for us requires a backdoor of logging into the Konflux prod-rh01 cluster;  I have not bothered getting into that with you because this repo is getting deprecated, and soon you won't need to worry about these tests.

also, the "upgrade" test is known to be notoriously flakey.

The deployment-ocp-414 test is a good indication things are OK.

I'm going to merge this, and then let you bring up a cluster, and run dev_setup.sh without the `--use-current-branch` so that it will run against main branch.

If the panels still look as you expect, you'll go onto the new "update infra-deployments" process I've alluded to previously that will start discussing in detail in today's stand up / scrum

